PR #2, opened 2024-11-19
Add docs for layers

westnordost:
**I hereby contribute this under the terms of the BSD license.**

### Some things that I noticed that I didn't act upon in this PR

1. according to MapLibre style documentation, `sourceLayer` for all vector sources is not optional.

2. there are different kinds of expressions and different properties support different expressions. I didn't add all the info which expressions are supported by which properties into the documentation comments, in hope that this could somehow be represented by the type system

3. currently, it is each explained in the documentation comments what each the unit of the different parameters and thus it's valid ranges are (dp, degrees, ems, percentage, ...). One could consider to use [value types](https://kotlinlang.org/docs/inline-classes.html#members) for this. Jetpack Compose's [`Dp`](https://developer.android.com/reference/kotlin/androidx/compose/ui/unit/Dp) is a value class, too.  
This would remove the necessity to write in the documentation comment which unit and valid range is expected.

4. the symbol layer has some serious icon/text smurf-naming. Maybe it is possible to split that up?

5. especially in the symbol layer, we have many parameters that mutually exclude each other or are only applicable if another parameter is specified. This cries for [sealed interfaces](https://kotlinlang.org/docs/sealed-classes.html). E.g. the whole anchoring of icons/text.

6. some parameters can actually not be expressions according to the docs. For those, it usually also wouldn't make sense.

7. The current parameter order mirrors the order in the MapLibre layers documentation, but it is not very consistent amongst the different layers. E.g. the `opacity` property is at the beginning for some, at the end for other layers.


sargunv:
> according to MapLibre style documentation, sourceLayer for all vector sources is not optional.

I currently rely on runtime validation from the underlying SDK for that (though I think the SDK silently ignores it). But there's probably room for some source type validation in the type system; for example I think certain layers only work with certain types of sources? And it would be nice to integrate that into the type system (with interfaces, probably?)

> This would remove the necessity to write in the documentation comment which unit and valid range is expected.

I understand using value classes to express units and validate ranges for const properties, but not sure how to integrate that with the expression system (without the external validator discussed in Slack)

> the symbol layer has some serious icon/text smurf-naming. Maybe it is possible to split that up?
I guess, I am talking data classes, again. Copying is cheap, though...

Agreed. Earlier I had data classes for layout and paint properties (mirroring the JSON spec) but found that particular split to not really provide any value so removed it. Instead, I think we can handle:

> especially in the symbol layer, we have many parameters that mutually exclude each other or are only applicable if another parameter is specified. This cries for [sealed interfaces](https://kotlinlang.org/docs/sealed-classes.html). E.g. the whole anchoring of icons/text.

with some targeted application of sealed interfaces and data classes for related and mutually exclusive properties. Though I don't know if, as the style schema changes, if things we consider to be similarly grouped properties today actually diverge in the future. Though I guess it's not bad to just make a small, well documented, breaking change and move on if that day comes.

> some parameters can actually not be expressions according to the docs. For those, it usually also wouldn't make sense.

It's actually unclear to me from reading the docs which can be expressions and which can not. So I allow anywhere the underlying SDK allows an expression to be an expression. There's probably room for further validation here, though I'd be wary of tightening the types such that properties that may in the future allow expressions cause breaking changes in this library. (Introducing a "constant expression" type might be useful here, so we can in the future make "expressions" assignable to it but only allow constants today).

> The current parameter order mirrors the order in the MapLibre layers documentation, but it is not very consistent amongst the different layers. E.g. the opacity property is at the beginning for some, at the end for other layers.

This was convenient when writing the parameters as I could easily see which were left to-do, but that's not necessary now that they're all in there. A reorder would be reasonable, perhaps paired with application of data classes to groups of related parameters.

> I think it would be preferrable if those "enums" that are just objects with string constants would be real enums or in some cases sealed interfaces, but I didn't find a way how to do that without refactoring stuff I wouldn't want to touch right now

Do you have an example? I'm not sure which objects you're referring to (or is this

westnordost:
> I understand using value classes to express units and validate ranges for const properties, but not sure how to integrate that with the expression system

As far as I understand the current setup with `Expression`s, the only solution I can come up with is the following:

Every value class used would have to be passed through (including implementing `ExpressionScope.const(MyValueClass)` functions) until `util.kt/normalizeJsonLike`, where it can be resolved to its value.
```kotlin
is MyValueClass -> value.value // or whatever
```

> It's actually unclear to me from reading the docs which can be expressions and which can not. So I allow anywhere the underlying SDK allows an expression to be an expression. 

I've come to the same conclusion today. Some properties that do not mention supporting data-driven-styling or interpolate expressions **do** support expressions like 
```
"text-keep-upright":  ["step", ["zoom"], false, 20, true]
```
(for e.g. labels on roads - keeps text upright only starting on zoom level 20).
I don't know if this counts as a data expression. But it is **an** expression, so even when nothing about expressions is mentioned in the layer spec, it doesn't mean that no expressions can be used.

> A reorder would be reasonable

I can do this as a follow-up of this PR.

> targeted application of sealed interfaces and data classes for related and mutually exclusive properties

I can do this in another follow-up PR, but it would just be a draft / suggestion for you to look at, because I am not sure if it conflicts with the Expression stuff.

>> I think it would be preferrable if those "enums" that are just objects with string constants would be real enums or in some cases sealed interfaces [...]

> [...] Yeah I think it should be possible if I refactor Expression as a sealed interface instead of a class.

Also replying to the in-code comment: I haven't quite understood how and why the Expression stuff is structured the way it is, so I can not really comment on 
```diff
-public const val Bevel: String = "bevel"
+public const val Bevel: Expression<String> = const("bevel")
```
At the moment, I am worried that this may conflict with introducing sealed interfaces for some stuff. Also, I would very much like to land at those enums being real enums so that (just like with value classes) it is clear from the function signature, what is being expected. Currently, the documentation comment that explains which constants to use is necessary (I would like to get rid of that).

In any case,regarding refactor plans of `Expression`, I think it would be better to do this after this PR has been merged, to keep this PR's scope small.

westnordost:
>> targeted application of sealed interfaces and data classes for related and mutually exclusive properties

> I can do this in another follow-up PR, but it would just be a draft / suggestion for you to look at, because I am not sure if it conflicts with the Expression stuff.

I've played around with it a bit, but didn't find anything where this is well applicable. I thought the whole `textAnchor` stuff would be, but, not really, in my opinion. `translate` + ̀ translateAnchor` could be put into a data class, but that's not going to make any big difference.
For `color`+`pattern`(+`gradient`), maybe it would make sense, but having to write e.g. always `paint = Paint.Color(const(Color.Red))` also seems unnecessarily complex to me.

sargunv:
Fair enough. I think this is a scenario where it's not the responsibility of the compose wrapper to reorganize the maplibre APIs (except how it relates to usage with Compose on platforms we support)

PR #8, opened 2024-11-25
Change Android minSdk to 21

michalgwo:
It would be nice to support a wider range of Android devices by setting minSdk to 21. Treat it as a suggestion, for now, it compiles successfully with minSdk 21, but I don't know if there are plans to use any functionality available only on SDK 24 and above, so feel free to reject the PR.

### DIFF
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -1,7 +1,7 @@
 [versions]
 agp = "8.5.2"
 android-compileSdk = "34"
-android-minSdk = "24"
+android-minSdk = "21"
 android-targetSdk = "34"
 compose = "1.7.1"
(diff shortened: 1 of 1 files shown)


sargunv:
Happy to drop the min to 21 for now. 24 was arbitrary and I've only tested on recent versions, so this is a good reminder to test some older stuff

sargunv:
It seems like minSdk 21 breaks the instrumented tests because it hangs on a permission prompt. I'm not familiar with how to get around this but I imagine it's a common issue and should be documented somewhere?

For now I bumped it to 23 as that's the level that allows runtime permissions.

Filed https://github.com/sargunv/maplibre-compose/issues/9, would accept a PR that fixes the permission issues to bring it back down to 21.

PR #36, opened 2024-11-27
Expose screenLocationFromPosition to CameraState

michalgwo:
It's useful to have access to screenLocationFromPosition outside of the library. A common use case is when someone wants to query rendered features around a certain point, they need the screen location of that point, for example:

``` kotlin
FloatingActionButton(
    onClick = {
      val position = Position(latitude = 45.521, longitude = -122.675)
      val queryCenter = cameraState.screenLocationFromPosition(position)

      queryCenter?.let {
        val queryRect = Rect(queryCenter.x - 20, queryCenter.y - 20, queryCenter.x + 20, queryCenter.y + 20)
        val result = cameraState.queryRenderedFeatures(queryRect)
      }
  })
```

By the way, what are your thoughts on pull requests without creating the issue first? Sometimes when the change is so tiny, I find it easier to make a pull request than create the issue, and you can treat it as a starting point of a discussion rather than a definitive solution. But if you prefer me to create issues first, I will keep this in mind in the future.

### DIFF
--- a/lib/maplibre-compose/src/commonMain/kotlin/dev/sargunv/maplibrecompose/compose/CameraState.kt
+++ b/lib/maplibre-compose/src/commonMain/kotlin/dev/sargunv/maplibrecompose/compose/CameraState.kt
@@ -51,8 +51,8 @@ public class CameraState internal constructor(firstPosition: CameraPosition) {
     map.animateCameraPosition(finalPosition, duration)
   }
 
-  public fun screenLocationFromPosition(position: Position): Offset? {
-    return map?.screenLocationFromPosition(position)
+  public fun screenLocationFromPosition(position: Position): Offset {
+    return map?.screenLocationFromPosition(position) ?: error("Map requested before it was initialized")
   }
 
   public fun queryRenderedFeatures(offset: Offset): List<Feature> {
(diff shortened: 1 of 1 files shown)


lib/maplibre-compose/src/commonMain/kotlin/dev/sargunv/maplibrecompose/compose/CameraState.kt
```diff
@@ -50,6 +51,10 @@ public class CameraState internal constructor(firstPosition: CameraPosition) {
     map.animateCameraPosition(finalPosition, duration)
   }
 
+  public fun screenLocationFromPosition(position: Position): Offset? {
+    return map?.screenLocationFromPosition(position)
+  }
```
sargunv, line 56:
```suggestion
  public fun screenLocationFromPosition(position: Position): Offset {
    return map?.screenLocationFromPosition(position) ?: error("Map requested before it was initialized")
  }
```

generally if this is called before the map is attached, that should be considered an error

maybe there's a use case for the caller to fail gracefully there, but I'd expose that as a separate `isInitialized` or something

sargunv:
Thanks for this!

> By the way, what are your thoughts on pull requests without creating the issue first?

that's totally fine

sargunv:
also technically this was covered by an issue as part of https://github.com/sargunv/maplibre-compose/issues/12

PR #46, opened 2024-11-28
use compose Dp for pixel sizes

westnordost:
### I see the following advantages for this:

1. consistent with other Compose API
2. easier to see from signature which type is expected. It is often not clear from the name alone which unit is expected. E.g. `blur` in most layers is in `dp`, but in `CircleLayer`, it is actually some kind of fraction. Offsets are usually in `dp`, but any text offsets are in `em` (i.e. relative to the font size). `iconSize` is also not in `dp` but a fraction.
3. shorter documentation comments (Units don't need to be mentioned in the comments)

### Possibilities for expansion:

In the `SymbolLayer`, actually all offsets concerning text are en `em`: `textOffset`, `textRadialOffset`, `textVariableAnchorOffset` but also `texMaxWidth` and `textLineHeight`.
It would be possible to add another value class named `Em` and allow these to be created with an expression like `10.em` (like for `dp`), but this extension function actually already exists in Compose, it creates a `TextUnit`, which can be `sp` (scale-independent pixels) or `em`. So, to create a rival `val Number.em get() = Em(this)` would only confuse.
Can we use `TextUnit` instead? We could, but then we need to internally support it also when the value was specified in `sp` instead. Read: Convert it to em, for which we need the specified font size. Which we have, of course.
So, doable, but [MapLibre-native does (currently) not automatically scale fonts](https://github.com/maplibre/maplibre-native/issues/3057), i.e. the font size is specified in `dp`, not in `sp`. I think it would be confusing to be able to specify the text in `sp` but not have the font scaling working. This _could_ also be solved in this library, but I think this is rather something that should be solved upstream.

When it is solved, I'd propose to change the unit of `textSize` to `TextUnit` along with the fields mentioned above. Until then, in my opinion, it is best to leave them as `Number`s and just document that their unit are ems.

### DIFF
--- a/demo-app/src/commonMain/kotlin/dev/sargunv/maplibrecompose/demoapp/demos/AnimatedLayerDemo.kt
+++ b/demo-app/src/commonMain/kotlin/dev/sargunv/maplibrecompose/demoapp/demos/AnimatedLayerDemo.kt
@@ -9,6 +9,7 @@ import androidx.compose.runtime.Composable
 import androidx.compose.runtime.getValue
 import androidx.compose.ui.Modifier
 import androidx.compose.ui.graphics.Color
+import androidx.compose.ui.unit.dp
 import dev.sargunv.maplibrecompose.compose.MaplibreMap
 import dev.sargunv.maplibrecompose.compose.layer.AnchorBelow
 import dev.sargunv.maplibrecompose.compose.layer.LineLayer
--- a/demo-app/src/commonMain/kotlin/dev/sargunv/maplibrecompose/demoapp/demos/AnimatedLayerDemo.kt
+++ b/demo-app/src/commonMain/kotlin/dev/sargunv/maplibrecompose/demoapp/demos/AnimatedLayerDemo.kt
@@ -55,7 +55,7 @@ fun AnimatedLayerDemo() = Column {
         id = "amtrak-routes",
         source = routeSource,
         color = const(animatedColor),
-        width = const(4f),
+        width = const(4.dp),
       )
     }
   }
--- a/demo-app/src/commonMain/kotlin/dev/sargunv/maplibrecompose/demoapp/demos/CameraFollowDemo.kt
+++ b/demo-app/src/commonMain/kotlin/dev/sargunv/maplibrecompose/demoapp/demos/CameraFollowDemo.kt
@@ -22,8 +22,8 @@ import androidx.compose.runtime.mutableStateOf
 import androidx.compose.runtime.remember
 import androidx.compose.runtime.setValue
 import androidx.compose.ui.Modifier
-import androidx.compose.ui.geometry.Offset
 import androidx.compose.ui.graphics.Color
+import androidx.compose.ui.unit.DpOffset
 import androidx.compose.ui.unit.dp
 import dev.sargunv.maplibrecompose.compose.CameraState
 import dev.sargunv.maplibrecompose.compose.MaplibreMap
@@ -64,12 +64,12 @@ private fun animateTestPosition(start: Position, end: Position): State<Position>
     end,
     typeConverter = remember { PositionVectorConverter(origin = START_POINT) },
     animationSpec =
-      remember {
-        infiniteRepeatable(
-          animation = tween(durationMillis = 30_000),
-          repeatMode = RepeatMode.Reverse,
-        )
-      },
+    remember {
+      infiniteRepeatable(
+        animation = tween(durationMillis = 30_000),
+        repeatMode = RepeatMode.Reverse,
+      )
+    },
   )
 }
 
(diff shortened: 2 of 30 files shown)


westnordost:
A TODO is still open.

sargunv:
(disclaimer, I've barely skimmed the PR so far, so some of my questions maybe answered there already)

Is there some Maplibre documentation on how it does scaling? Android has Dp which is defined by a physical dimension (1/160th of an inch). iOS has Point, whose definition I'm not aware of but I believe it changes by device class. Does a unit of 1 passed to line width in MapLibre correspond to 1/160", on either platform? If not, I think Dp would be misleading. 

Actually that makes me wonder, is Dp in multiplatform compose even defined the same on other platforms as on Android, or is it based on Point on iOS 🤔 

If we switch to such units, we'll need to consider how they're used in math expressions, interpolation, etc too. Maybe like you made a common interface for all enum values in expressions, some common interface for all numeric values in expressions (though Kotlin can't do extension interfaces)? Maybe you handled this bit already

westnordost:
> Does a unit of 1 passed to line width in MapLibre correspond to 1/160", on either platform?

Good question. I don't think it is documented anywhere in MapLibre. But I can test it. On Android, that is:

In your example app, the list with the demos is separated each by a divider of a width of 1dp. 
When I change the line width of the line displayed in the AnimatedLayerDemo to 1, I can confirm that that line width is exactly the same as the width of the 1dp divider.

You could do the same test on iOS, if you have a device at hand.

> If we switch to such units, we'll need to consider how they're used in math expressions, interpolation, etc too. 

Isn't this irrelevant? In Java/Objective-C code, it's all Floats/Doubles again.

sargunv:
https://osmus.slack.com/archives/C02B2CBSNBU/p1732835252027049

We confirmed it looks like `Dp` on android (and it seems you have as well).

iOS uses `UIScreen.scale` which is 1x, 2x, or 3x: https://developer.apple.com/documentation/uikit/uiscreen/1617836-scale?language=objc

It doesn't define it further, but if it's a multiple of the original iPhone's DPI, that'd line up pretty closely with `Dp`. Original iPad had a lower DPI though, and I wouldn't put it past Apple to define their scale differently on different device categories. But if they do, maybe Compose Multiplatform also uses that same scale factor to scale Dp, instead of the 160dpi definition? If so I think we're still good to use Dp here.

Some TODOs (I can do these when I have some time):
* do your experiment but on ios and ipados (I have a spare iphone but not ipad, though on a simulator should be fine)
* repeat on an android tablet (I don't expect this to be an issue, but just to cover our bases)
* look into scaling on MapLibre JS (I expect it relies on some browser API to get a scale factor, which I expect should behave similarly to compose dp, but should confirm)

> Isn't this irrelevant? In Java/Objective-C code, it's all Floats/Doubles again.

It's relevant because `interpolate` has variants for `Color`, `Number`, `Point`, and `List<Number>`. With this PR, we'll need to replace the `Point` with both `DpOffset` and `Offset` I assume, and add `Dp` and `List<Dp>`. So that's +3 `interpolate` overloads. And we'll have to review all the other expression functions that operate on numbers (all the arithmetic ones at least, unsure if there's others to watch out for) and add overloads. And then someday for the text units too. This is the main reason I haven't gone down this route yet, well that and the dp uncertainty (I'm not against this if it results in a better API though)

sargunv:
A request, for the PR, it would be good to add `interpolate` to line-width in the the amtrak demo ("animated layers" I think) to show how to use `Dp` in complex expressions. We can copy some road width interpolation from the base style perhaps

sargunv:
on iphone, a 7dp radius `CircleLayer` and a 14dp side `Box` look like the same scale to me 👍 

CleanShot 2024-11-28 at 23 22 05@2x

ipad mini, ditto 👍 

CleanShot 2024-11-28 at 23 25 15@2x

and full blown ipad, lgtm 👍 

CleanShot 2024-11-28 at 23 27 40@2x

So it seems Compose adapts its `Dp` unit to iOS's definition of scale, breaking from the Android standard of 1/160".

this is pretty promising

sargunv:
And I've verified a 32dp radius CircleLayer on [Maputnik](https://maplibre.org/maputnik/) and a 64dp height TopAppBar on Compose Web ([People in Space demo](https://joreilly.github.io/PeopleInSpace/)) are the same size on desktop and mobile, so it seems a MapLibre GL JS "pixel" is the same as a CSS `px` is the same as a Compose Web `Dp`. So Dp is definitely a reasonable unit to use.

Happy to merge once the expression stuff is solved

sargunv:
re merge conflicts: I'm working on integrating some changes piecemeal that should hopefully reduce the scope of what needs to be updated here

see:
* https://github.com/sargunv/maplibre-compose/pull/63
* https://github.com/sargunv/maplibre-compose/pull/64
* https://github.com/sargunv/maplibre-compose/pull/65
  *  `PaddingValues.Absolute`, despite the clunky name, is more suitable to replace Insets than Rect

westnordost:
Isn't the stuff in `util.kt` missing now? I.e. getting from Dp, DpOffset and PaddingValues.Absolute back to Numbers?

sargunv:
I should probably comment relevant code better, but the gist is:
* Types that are the same on both platforms get mapped when the Expression is constructed, so `Expression.ofDp(1.dp)` creates an Expression with `value = 1f`. Then util.kt treats it like a number. 
* Types that are different on both platforms (color, offset, padding) put the original type in `value` and map to the platform specific value in util.kt

So with this version there's never actually a Dp or DpOffset object in Expression.value, it's passed as float or Offset internally. 

Long term, if/when we build directly on the native core, I'm hoping to cut out the platform specific quirks like that and map to the json-like type directly in commonMain code for all values.

westnordost:
Good solution!

westnordost:
Yeah, in general, explanatory comments in code is also a good - or maybe the best - place to document why a particular implementation was chosen. Too often it happens to myself that I look at some of my old(er) code and am confused why the heck I did it this way, resulting in search through the git history and linked issues until I find an answer. I could've spared myself that many times, in hindsight.

PR #76, opened 2024-12-02
feat: add camera move reason tracking

### LINKED ISSUE #12: Flesh out camera state
Context: the discussion here: https://github.com/sargunv/maplibre-compose/discussions/7

Since we can't hoist camera state without reimplementing much of the upstream SDK (see discussion above) we currently provide the camera position (including lat/lng, zoom, pitch, bearing, padding) as state. This is enough to read where the camera is, but the caller can't do much with just that. We should additionally provide:

* Some state for the consumer to know when the user is moving the camera with gestures, so the consumer can cancel their own animations (required for allowing the user to break out of a point follow animation)
* Some functions to convert map coordinates to/from screen space coordinates based on the camera projection (so the user can position overlays, for instance)

Example from Google Maps: https://googlemaps.github.io/android-maps-compose/maps-compose/com.google.maps.android.compose/-camera-position-state/index.html

They provide:

```kt
// we don't yet have these
var cameraMoveStartedReason: CameraMoveStartedReason
var isMoving: Boolean
val projection: Projection

 // we have functionality comparable to these
var position: CameraPosition
suspend fun animate(update: CameraUpdate, durationMs: Int = MAX_VALUE)
fun move(update: CameraUpdate)
```

sargunv:
Added camera movement tracking to improve map interaction. To showcase this, in the camera follow demo, the camera now stops following when users interact with the map through gestures like panning, zooming, rotating, or tilting. Users can resume following with a new "Follow" button.

closes #12

### DIFF
--- a/lib/maplibre-compose/src/commonMain/kotlin/dev/sargunv/maplibrecompose/compose/CameraState.kt
+++ b/lib/maplibre-compose/src/commonMain/kotlin/dev/sargunv/maplibrecompose/compose/CameraState.kt
@@ -5,6 +5,7 @@ import androidx.compose.runtime.mutableStateOf
 import androidx.compose.runtime.remember
 import androidx.compose.ui.unit.DpOffset
 import androidx.compose.ui.unit.DpRect
+import dev.sargunv.maplibrecompose.core.CameraMoveReason
 import dev.sargunv.maplibrecompose.core.CameraPosition
 import dev.sargunv.maplibrecompose.core.MaplibreMap
 import dev.sargunv.maplibrecompose.core.expression.Expression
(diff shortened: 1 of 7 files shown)


sargunv:
* **#76** <a href="https://app.graphite.dev/github/pr/sargunv/maplibre-compose/76?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/sargunv/maplibre-compose/76?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`

<h2></h2>

lib/maplibre-compose/src/commonMain/kotlin/dev/sargunv/maplibrecompose/compose/CameraState.kt
```diff
@@ -42,8 +44,8 @@ public class CameraState internal constructor(firstPosition: CameraPosition) {
       positionState.value = value
     }
 
-  public val isInitialized: Boolean
-    get() = map != null
```
sargunv, line 46: removing this as it wasn't state and so might be confusing

if someone needs it someday, reimplement with state

PR #79, opened 2024-12-02
document some settings

westnordost:
also, I noticed that on iOS, there doesn't seem to be an option to turn on/off the
- double tap to zoom in
- double tap, then hold and drag up or down to zoom
gestures separately from zooming via gesture at all.

So, to achieve the same behavior on all platforms, on Android, we need to set all these separate settings to true/false when the allow-zoom setting changes.

### DIFF
--- a/lib/maplibre-compose/src/commonMain/kotlin/dev/sargunv/maplibrecompose/core/CameraPosition.kt
+++ b/lib/maplibre-compose/src/commonMain/kotlin/dev/sargunv/maplibrecompose/core/CameraPosition.kt
@@ -5,6 +5,20 @@ import androidx.compose.runtime.Immutable
 import androidx.compose.ui.unit.dp
 import io.github.dellisd.spatialk.geojson.Position
 
+/**
+ * Defines how the camera is oriented towards the map.
+ *
+ * @param bearing Direction that the camera is pointing in, in degrees clockwise from north.
+ * @param target Target position to align with the center of the screen, i.e. where the camera
+ *   is pointing at.
+ * @param tilt The camera angle, in degrees, from the nadir (directly down). A value in the range of
+ *   `[0 .. 60]`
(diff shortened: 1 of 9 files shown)


westnordost:
I also noticed that minZoom is missing from GeoJsonOptions. I added it.

There are some TODOs:
- The documentation mentions that the cluster id can be used with the [cluster inspection methods](https://maplibre.org/maplibre-gl-js/docs/API/classes/GeoJSONSource/#getclusterexpansionzoom), but such methods are currently missing on `GeoJsonSource`
- Examples for `clusterProperties` are missing, because I think the current data structure does not correctly reproduce what is possible [as documented in the style spec](https://maplibre.org/maplibre-style-spec/sources/#clusterproperties)(?)

lib/maplibre-compose/src/commonMain/kotlin/dev/sargunv/maplibrecompose/core/CameraPosition.kt
```diff
@@ -1,11 +1,25 @@
...
+ *   pointing at.
+ * @param tilt The camera angle, in degrees, from the nadir (directly down). A value in the range of
+ *   `[0 .. 60]`
+ * @param zoom Zoom level at target. A value in the range of `[0 .. 25.5]`
+ * @param padding By default, the camera points at the center of the screen, i.e. it also rotates
+ *   and tilts around that point. By specifying a padding, this center point can be altered, for
+ *   example if you want to display a bottom sheet above the lower part of the map, focussing on a
+ *   POI in the upper part of the map.
+ */
+@Stable
```
sargunv, line 22: I'd leave a comment on why this is Stable, not Immutable (PaddingValues).

In the future we might need to change `PaddingValues` below to `PaddingValues.Absolute`, and if we do we'll want to change this back to Immutable

PR #88, opened 2024-12-03
Docs for camera state

westnordost:
also added a new method and consolidated several method overloads into one, after checking in the native source code that the behavior doesn't change.

### DIFF
--- a/lib/maplibre-compose/src/androidMain/kotlin/dev/sargunv/maplibrecompose/core/AndroidMap.kt
+++ b/lib/maplibre-compose/src/androidMain/kotlin/dev/sargunv/maplibrecompose/core/AndroidMap.kt
@@ -1,5 +1,7 @@
 package dev.sargunv.maplibrecompose.core
 
+import android.graphics.PointF
+import android.graphics.RectF
 import androidx.compose.foundation.layout.PaddingValues
 import androidx.compose.ui.unit.Density
 import androidx.compose.ui.unit.DpOffset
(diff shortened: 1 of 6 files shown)


westnordost:
Hmm, regarding the non-working iOS, you might need to help me. I implemented that blindly, because I don't have a Mac at hand.

lib/maplibre-compose/src/androidMain/kotlin/dev/sargunv/maplibrecompose/core/AndroidMap.kt
```diff
@@ -273,27 +277,25 @@ internal class AndroidMap(
...
-    map
-      .queryRenderedFeatures(
-        offset.toPointF(density),
-        predicate?.toMLNExpression(),
-        *layerIds.orEmpty().toTypedArray(),
-      )
+  ): List<Feature> {
+    // Kotlin hack to pass null to a java nullable varargs
+    val query: (PointF, MLNExpression?, Array<String>?) -> List<MLNFeature> =
+      map::queryRenderedFeatures
```
sargunv, line 283: Passing null instead of empty array in order to align with a future fix to: https://github.com/maplibre/maplibre-native/issues/2828

(it also just looks weird to an unfamiliar reader that an empty array would be the same as naming every layer)

But since Kotlin doesn't support Java's nullable varargs, we have to eliminate the varargs with the soft cast in `val query: ...`.

PR #98, opened 2024-12-07
some docs for Sources

westnordost:
also,
- added the tileSize parameter for RasterSource
- remember*Source return typed *

### DIFF
--- a/lib/maplibre-compose/src/commonMain/kotlin/dev/sargunv/maplibrecompose/compose/source/rememberGeoJsonSource.kt
+++ b/lib/maplibre-compose/src/commonMain/kotlin/dev/sargunv/maplibrecompose/compose/source/rememberGeoJsonSource.kt
@@ -8,7 +8,7 @@ import io.github.dellisd.spatialk.geojson.GeoJson
 
 /**
  * Remember a new [GeoJsonSource] with the given [id] and [options] from the GeoJson data at the
- * given [dataUrl].
+ * given [uri].
  *
  * @throws IllegalArgumentException if a layer with the given [id] already exists.
  */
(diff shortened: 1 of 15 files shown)


westnordost:
I just documented the status quo, you may want to check if you think the type of exceptions thrown is what you want.

lib/maplibre-compose/src/commonMain/kotlin/dev/sargunv/maplibrecompose/compose/source/rememberGeoJsonSource.kt
```diff
@@ -4,30 +4,40 @@ import androidx.compose.runtime.Composable
...
+ * @throws IllegalArgumentException if a layer with the given [id] already exists.
+ */
 @Composable
 @Suppress("NOTHING_TO_INLINE")
 public inline fun rememberGeoJsonSource(
   id: String,
   uri: String,
   options: GeoJsonOptions = GeoJsonOptions(),
-): Source =
+): GeoJsonSource =
```
sargunv, line 21: Okay to merge for now but I think we'll probably want to split these types into mutable and non mutable interfaces, and return the non mutable interface here so there's only one path for setting the uri (via the Composable parameter). 

Making mutations internal would also work, but I'd like to split the modules between core and compose so internal wouldn't work forever

westnordost: Why should Source(s) be mutable at all?

sargunv: The geojson data can be updated. We use that in the demo app right now on the camera follow demo, to move the location puck.

In compose that's done with a parameter to the Composable, but the underlying non-compose model just has a setter for it.

westnordost: Oh, right! I didn't look at the code, I wrongly assumed that (in compose) the whole GeoJsonSource would be replaced on any change to it.

PR #107, opened 2024-12-11
Expressions documentation and refinement

westnordost:
### Changes (and suggestions)

- Added documentation comments

- Added `// regions` (used to group code and collapse/expand these sections in IDEs such as VSCode or IntelliJ)

- Fixed parameter names of `slice`: `length` was wrong, it is `endIndex` ([see source code](https://github.com/maplibre/maplibre-style-spec/blob/main/src/expression/definitions/slice.ts#L22))

- Fixed order of `fallback` parameter for `match` and `case` - it was inconsistent (even though it is consistent in the spec)

- Documentation claims that "==" and "!=" only works for Booleans, Strings, Numbers, so I added overloads for these types instead of for * ([see source code](https://github.com/maplibre/maplibre-style-spec/blob/7604485fc0b66394b62a44fa9d4cd9816d7d8858/src/expression/definitions/comparison.ts#L29-L30))

- Documentation claims that `within` only works for Polygons, so replaced `Geometry` with `Polygon` and `MultiPolygon` in parameter ([see source code](https://github.com/maplibre/maplibre-style-spec/blob/main/src/expression/definitions/within.ts#L11))

- renamed some parameter names to be more "speaking"

- Made many functions that work on another expression into extension functions, analogous to many math functions:
  - `string(e, …)` → `e.asString(…)`
  - `number(e, …)` → `e.asNumber(…)`
  - `boolean(e, …)` → `e.asBoolean(…)`
  - `'object'(e, …)` → `e.asMap(…)` also getting rid of ugly apostrophes
  - `array(e, …)` → `e.asList(…)`
  - `numberFormat(e, ...)` → `e.numberFormat()`
  - `toString(e)` → `e.convertToString()` "convert" is necessary because there's already a `toString` :-/
  - `toNumber(e, …)` → `e.convertToNumber(…)`
  - `toBoolean(e)` → `e.convertToBoolean()`
  - `toColor(e, …)` → `e.convertToString(…)`
  - `i 'in' list` → `list.contains(i)` because those apostrophes around the "in" are quite ugly
  - `indexOf(i, e, …)` → `e.indexOf(i, …)`
  - `slice(list, …)` → `list.slice(…)`
  - `slice(str, …)` → `str.substring(…)` also renamed to be similar to kotlin naming ¯\_(ツ)_/¯
  - `length(e) → `e.length()`
  - `toRgba(e)` → `e.toRgbaComponents()`
  - `isSupportedScript(str)` → `str.isScriptSupported` also switched the two words because before it sounded like str is the script. No, str is a normal string that is *written* in a script
  - `upcase(str)` -> `str.uppercase()` used Kotlin-naming. Also, boy, does it smell like updog in here?
  - `downcase(str)` -> `str.lowercase()` see above
  - 

- Added
  - `string.indexOf(substring, …)` because it was missing
  - `map[i]` / `map.get(i)` it was part of `get(i, …)` before, which didn't make much sense
  - `map.has(i)` it was part of `has(i, …)` before, which didn't make much sense. **NOTE**: I did not rename to kotlin-idiomatic function name `containsKey`. Inconsistent?

- Removed / Replaced
  - `at(i, e)` → there's already `e[i]` / `get(i)`
  - `not(e)` → there's already `!`
  - `sum(…)` → there's `+`
  - `product(…)` → there's `*`
  - `concat(…)` → use `+` operator for strings
  - `rgb(…)` and `rgba(…)` → `rgbColor(…)` because I think it makes sense to write what is being instantiated. (Or alternatively, just `color`, but maybe there will be e.g. a `hsvColor` one day)

### DIFF
--- a/demo-app/src/commonMain/kotlin/dev/sargunv/maplibrecompose/demoapp/demos/ClusteredPointsDemo.kt
+++ b/demo-app/src/commonMain/kotlin/dev/sargunv/maplibrecompose/demoapp/demos/ClusteredPointsDemo.kt
@@ -63,7 +63,7 @@ object ClusteredPointsDemo : Demo {
           rememberGeoJsonSource(
             "bikes",
             gbfsData,
-            GeoJsonOptions(cluster = true, clusterMaxZoom = 16, clusterRadius = 32),
+            GeoJsonOptions(cluster = true, clusterRadius = 32, clusterMaxZoom = 16),
           )
 
         CircleLayer(
--- a/lib/maplibre-compose/src/commonMain/kotlin/dev/sargunv/maplibrecompose/compose/layer/CircleLayer.kt
+++ b/lib/maplibre-compose/src/commonMain/kotlin/dev/sargunv/maplibrecompose/compose/layer/CircleLayer.kt
@@ -29,7 +29,9 @@ import dev.sargunv.maplibrecompose.core.source.Source
  *   this, the layer will be hidden. A value in the range of `[0..24]`.
  * @param filter An expression specifying conditions on source features. Only features that match
  *   the filter are displayed. Zoom expressions in filters are only evaluated at integer zoom
- *   levels. The `feature-state` expression is not supported in filter expressions.
+ *   levels. The
+ *   [featureState][dev.sargunv.maplibrecompose.core.expression.ExpressionScope.featureState]
+ *   expression is not supported in filter expressions.
  * @param visible Whether the layer should be displayed.
  * @param sortKey Sorts features within this layer in ascending order based on this value. Features
  *   with a higher sort key will appear above features with a lower sort key.
--- a/lib/maplibre-compose/src/commonMain/kotlin/dev/sargunv/maplibrecompose/compose/layer/FillExtrusionLayer.kt
+++ b/lib/maplibre-compose/src/commonMain/kotlin/dev/sargunv/maplibrecompose/compose/layer/FillExtrusionLayer.kt
@@ -27,7 +27,9 @@ import dev.sargunv.maplibrecompose.core.source.Source
  *   this, the layer will be hidden. A value in the range of `[0..24]`.
  * @param filter An expression specifying conditions on source features. Only features that match
  *   the filter are displayed. Zoom expressions in filters are only evaluated at integer zoom
- *   levels. The `feature-state` expression is not supported in filter expressions.
+ *   levels. The
+ *   [featureState][dev.sargunv.maplibrecompose.core.expression.ExpressionScope.featureState]
+ *   expression is not supported in filter expressions.
  * @param visible Whether the layer should be displayed.
  * @param translate The geometry's offset relative to the [translateAnchor]. Negative numbers
  *   indicate left and up (on the flat plane), respectively.
--- a/lib/maplibre-compose/src/commonMain/kotlin/dev/sargunv/maplibrecompose/compose/layer/FillLayer.kt
+++ b/lib/maplibre-compose/src/commonMain/kotlin/dev/sargunv/maplibrecompose/compose/layer/FillLayer.kt
@@ -26,7 +26,9 @@ import dev.sargunv.maplibrecompose.core.source.Source
  *   this, the layer will be hidden. A value in the range of `[0..24]`.
  * @param filter An expression specifying conditions on source features. Only features that match
  *   the filter are displayed. Zoom expressions in filters are only evaluated at integer zoom
- *   levels. The `feature-state` expression is not supported in filter expressions.
+ *   levels. The
+ *   [featureState][dev.sargunv.maplibrecompose.core.expression.ExpressionScope.featureState]
+ *   expression is not supported in filter expressions.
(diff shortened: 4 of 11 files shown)


westnordost:
### Open questions

1. The `let`/`var` API is already hideous on MapLibre-GL-JS and even worse in Kotlin due to the apostrophes. Suggestion:
  Rename `let(name, value, expression)` → `withVariable(name, value, expression)`
  Rename `'var'(name)` → `useVariable(name)`

2. Couldn't `TInterpolationType` be a sealed interface? Would make it clearer what to specify for the first parameter of the `interpolate` functions.
```
public sealed interface Interpolation {
  public data object Linear : Interpolation
  public data class Exponential(base: Expression<Number>) : Interpolation
  public data class CubicBezier(
    x1: Expression<Number>,
    y1: Expression<Number>,
    x2: Expression<Number>,
    y2: Expression<Number>,
  ) : Interpolation
}
```

3. maybe it would improve clarity to rename functions that refer to the feature to more speaking names like `featureId`, `featureProperties`, `featureHas`, `featureGeometryType` etc... (or `feature.id`, `feature.get`, `feature.has`, ... with a local `object feature { ... }`)?

4. The allowed output types for `interpolate` are actually only number, array<number> and color. So for us, this would mean Number, List<Number>, Color, Dp, PaddingValues.Absolute, Offset, OffsetDp (the three last ones being internally List<Number>s). I hesitate to add overloads for all of those. But currently, it is just "anything" which is wrong.

5. `geometryType` returns a string that is either "Point", "MultiPoint", "LineString", "MultiLineString", "Polygon" or "MultiPolygon". This would likely be used for comparison. Couldn't this be an enum? (Not sure how to implement)

6. for `properties()` - shouldn't it return `Expression<Map<String, Any>>` instead of `Expression<Map<String, T>>`? After all, not all property values are the same

7. alternative suggestion for `case` syntax:
Currently:
```
case(
  all(has("color1"), has("color2") then
    interpolate(linear(), zoom(), 1 to get("color1"), 20 to get("color2")),
  has("color") then
    get("color"),
  const(Color.Red)
)
```
Suggestion:
```
switch(
  Case(
    label = has("color1") and has("color2"),
    output = interpolate(linear(), zoom(), 1 to get("color1"), 20 to get("color2"))
  ),
  Case(
    label = has("color"),
    output = get("color")
  )
  fallback = const(Color.Red)
)
```
It's less compact, but IMO it is more explicit, less magic. Also, "switch-case" terminology is what people know. Also, the `then` from the current solution is shown together with the `then` for the MatchBranches in IDEs

8. analogous to the above, suggestion for `match` syntax:
Currently:
```
match(
  get("building_type"),
  "residential" then const(Color.Cyan),
  listOf("commercial", "industrial") then const(Color.Yellow),
  const(Color.Red),
)
```
Suggestion:
```
match(
  get("building_type"),
  Arm(
    label = "residential"
    output = const(Color.Cyan),
  ),
  Arm(
    label = listOf("commercial", "industrial"),
    output = const(Color.Yellow)
  ),
  fallback = const(Color.Red),
)
```

### TO DOs
- add many overloads where the parameter(s) are not Expressions but literals
- add documentation to `const` functions

sargunv:
> let/var

yup sounds good

> Couldn't TInterpolationType be a sealed interface?

sounds reasonable. it could extend `Expression` so it integrates nicely with the mappers

> feature data functions

I like the `feature.foo` syntax if the `feature` object is accessible in an ExpressionScope block. If not, `featureFoo` is fine

> interpolate overloads

Actually originally this had overloads for each supported type, but I generified it to `<T>` when we added all the Compose types like Dp. Not opposed to overloads I guess? But also, I've been considering a different approach to Expression typing, simulating union types by defining a family of non-generic Expression interfaces. So probably leave it be for now, and we can add overloads if it doesn't work out

> geometryType enum

it can work the same as any other layer property enum type, right?

> for properties()

Yes, though it'll likely lead to more unchecked `as` casting.

actually, all these extra types makes me think we need some solution for something like `.asString()` to also narrow a type to specialized types like `asPoint` or `asLineCap`. Right now the user would have to do an unchecked `as` cast to narrow a type in such a way. Unsure about the best way to implement this

> alternative suggestion for case/match syntax:

I like the general direction, though the naming in your example is unclear to me (like I've never seen "Arm" used this way). `switch/Case` I think aligns with `match`. Unsure about case (if it's not reserved, I'd use `when/Condition`)

sargunv:
Re: the above ^ I think in any case where we're calling constructors for some special values (case, interpolation type, etc) we should retain some corresponding function in the `ExpressionScope` so the consumer can use the entire expression dsl without additional imports

westnordost:
I took the "Arm" naming from the Rust documentation. https://doc.rust-lang.org/reference/expressions/match-expr.html

sargunv:
I ended up using the name `switch()` for both case and match, and then `case(label, output)` and `condition(test, output)` for the "arms" of each function. See https://github.com/sargunv/maplibre-compose/pull/117

sargunv:
everything from the above comments is implemented now, except a sealed interface for interpolation type.

PR #112, opened 2024-12-12
refactor: rename let/var to withVariable/useVariable

sargunv:
Renamed variable binding functions to be more descriptive:
- `let` -> `withVariable`
- `var` -> `useVariable`

These new names better reflect their purpose in the expression DSL and avoid potential naming conflicts with Kotlin keywords.

related: https://github.com/sargunv/maplibre-compose/pull/107

PR #113, opened 2024-12-12
feat: overhaul expression typing to simulate union types with sealed interfaces

sargunv:
This PR overhauls expression typing to simulate union types with sealed interfaces. By replacing the old literal type parameters to `Expression<T>` with sealed interfaces we control like `StringValue`, `BooleanValue`, `DpValue`, we more directly control the assignability of types and therefore don't need to write a whole bunch of overloads to represent the expression function typings. Interfaces can extend multiple other interfaces, allowing us to get something close to union types.

One potential downside is the type parameter of an expression won't be directly clickable in the docs to the const type of the expression. But imo this was misleading in the past anyway (as the type parameter on the expression didn't really say anything meaningful about the value inside) and the new typing enumerates precisely the full set of expression types that exist. And we can add kdoc comments on expression interfaces linking to the expression function (usually `const`) intended to instantiate a literal of that expression type.


sargunv:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/sargunv/maplibre-compose/113?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>
> <a href="https://graphite.dev/docs/merge-pull-requests">Learn more</a>

* **#113** <a href="https://app.graphite.dev/github/pr/sargunv/maplibre-compose/113?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/sargunv/maplibre-compose/113?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#112** <a href="https://app.graphite.dev/github/pr/sargunv/maplibre-compose/112?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

lib/maplibre-compose/src/commonMain/kotlin/dev/sargunv/maplibrecompose/core/expression/Expression.kt
```diff
@@ -1,81 +1,106 @@
...
-      this >= 0 && this < constSmallInts.size && this.toInt().toFloat() == this
-
-    internal fun ofFloat(float: Float): Expression<Number> {
-      return when {
-        float.isSmallInt() -> constSmallInts[float.toInt()]
-        (float * 20f).isSmallInt() -> constSmallFloats[(float * 20f).toInt()]
-        else -> Expression(float)
-      }
-    }
+public sealed interface Expression {
```
sargunv, line 3: Most of the magic is here

lib/maplibre-compose/src/commonMain/kotlin/dev/sargunv/maplibrecompose/core/expression/ExpressionScope.kt
```diff
@@ -838,11 +719,11 @@ public interface ExpressionScope {
    * zoom 24, it is 256. Applied to for example line width, this has the visual effect that the line
    * stays the same width in meters on the map (rather than on the viewport).
    */
-  public fun <Output> interpolate(
-    type: Expression<TInterpolationType>,
-    input: Expression<Number>,
-    vararg stops: Pair<Number, Expression<Output>>,
-  ): Expression<Output> = interpolateImpl("interpolate", type, input, *stops)
+  public fun <Output : Expression.Interpolateable> interpolate(
```
sargunv, line 722: example of a properly typed expression now _without_ adding a bunch of overloads

westnordost:
Will review tomorrow (UTC+1, i.e. in ~10 hours from now )

westnordost:
> as the type parameter on the expression didn't really say anything meaningful about the value inside

That's only true for the Token types, though.

lib/maplibre-compose/src/commonMain/kotlin/dev/sargunv/maplibrecompose/core/expression/Expression.kt
```diff
@@ -1,81 +1,102 @@
...
+
+/**
+ * Represents an expression that resolves to any numeric quantity. Corresponds to numbers in the
+ * JSON style spec.
+ */
+public sealed interface ScalarExpression :
+  Expression, MatchableExpression, InterpolateableExpression, ComparableExpression
+
+/** Represents an expression that resolves to an integer quantity. */
+public sealed interface IntScalarExpression : ScalarExpression
```
westnordost, line 20: What's the difference between IntScalarExpression and IntExpression?

(Also, shouldn't it be IntegerExpression because that's how the type is named in Kotlin?)

lib/maplibre-compose/src/commonMain/kotlin/dev/sargunv/maplibrecompose/core/expression/Expression.kt
```diff
@@ -1,81 +1,102 @@
...
+ * Represents an expression that resolves to a collator object for use in locale-dependent
+ * comparison operations. See [ExpressionScope.collator].
+ */
+public sealed interface CollatorExpression : Expression
+
+/** Represents an expression that resolves to a formatted string. See [ExpressionScope.format]. */
+public sealed interface FormattedExpression : Expression
+
+/** Represents an expression that resolves to a geometry object. */
+public sealed interface GeoJsonExpression : Expression
```
westnordost, line 81: GeoJson includes Features, too (with tags and all). Does this resolve to GeoJson, or just a GeoJson geometry?

lib/maplibre-compose/src/commonMain/kotlin/dev/sargunv/maplibrecompose/core/expression/Expression.kt
```diff
@@ -1,81 +1,102 @@
...
+
+/**
+ * Represents an expression that resolves to a 2D floating point offset in device-independent
+ * pixels.
+ */
+public sealed interface DpOffsetExpression : VectorExpression
+
+/**
+ * Represents an expression that resolves to an absolute (RTL unaware) padding applied along the
+ * edges inside a box.
```
westnordost, line 67:
```suggestion
 * Represents an expression that resolves to an absolute (layout direction unaware) padding applied
 *  along the edges inside a box.
```

lib/maplibre-compose/src/commonMain/kotlin/dev/sargunv/maplibrecompose/core/expression/Expression.kt
```diff
@@ -1,81 +1,102 @@
...
-    // return Expression<*> because without ["literal" ... ] MapLibre may not treat it as a list
-    internal fun ofList(list: List<Expression<*>>): Expression<*> =
-      Expression<Any?>(list.map { it.value })
-
-    // return Expression<*> because without ["literal" ... ] MapLibre may not treat it as a list
-    internal fun ofMap(map: Map<String, Expression<*>>): Expression<*> =
-      Expression<Any?>(map.entries.associate { (key, value) -> key to value.value })
-  }
+/** Wraps a JSON-like value that represents an expression, typically used for styling map layers. */
+public sealed interface Expression {
```
westnordost, line 4: I think it would add a lot of value if the doc comments for the different linked directly to the functions in the `ExpressionScope`. You already do that for some, but it could be more. (I could also do that after this is merged.)

E.g.

```kotlin
/**
 * Represents an expression that resolves to a value that can be ordered with other values of its
 * type. I.e. it can be compared with [gt][ExpressionScope.gt], [gte][ExpressionScope.gte], 
 * [lt][ExpressionScope.lt], [lte][ExpressionScope.lte].
 */
public sealed interface ComparableExpression : Expression
```

lib/maplibre-compose/src/commonMain/kotlin/dev/sargunv/maplibrecompose/core/expression/ExpressionImpl.kt
```diff
@@ -0,0 +1,74 @@
...
+    private val ofWhite = ExpressionImpl(Color.White)
+    private val ofEmptyList = ExpressionImpl(emptyList<Any?>())
+    private val ofZeroOffset = ExpressionImpl(Offset.Zero)
+    private val ofZeroPadding = ExpressionImpl(ZeroPadding)
+
+    private val ofEmptyMap = ExpressionImpl(emptyMap<String, Any?>())
+
+    private fun isSmallInt(f: Float) = f >= 0 && f < 512 && f.toInt().toFloat() == f
+
+    fun ofFloat(float: Float) =
```
westnordost, line 45: I am surprised that there isn't a 

```suggestion

    fun ofInt(int: Int) = 
      if (int >= 0 && int < constSmallInts.size) constSmallInts[int]
      else ExpressionImpl(int)
      
    fun ofFloat(float: Float) =
```

lib/maplibre-compose/src/androidMain/kotlin/dev/sargunv/maplibrecompose/core/layer/HillshadeLayer.kt
```diff
@@ -13,27 +15,27 @@ internal actual class HillshadeLayer actual constructor(id: String, actual val s
   Layer() {
   override val impl = MLNHillshadeLayer(id, source.id)
 
-  actual fun setHillshadeIlluminationDirection(direction: Expression<Number>) {
+  actual fun setHillshadeIlluminationDirection(direction: IntExpression) {
```
westnordost, line 18:
```suggestion
  actual fun setHillshadeIlluminationDirection(direction: FloatExpression) {
```

see https://github.com/maplibre/maplibre-native/blob/72d2c70130fcd4e320cf623f9bec811108b6e201/include/mbgl/style/layers/hillshade_layer.hpp#L48

lib/maplibre-compose/src/androidMain/kotlin/dev/sargunv/maplibrecompose/core/layer/RasterLayer.kt
```diff
@@ -12,35 +14,35 @@ internal actual class RasterLayer actual constructor(id: String, actual val sour
   Layer() {
   override val impl = MLNRasterLayer(id, source.id)
 
-  actual fun setRasterOpacity(opacity: Expression<Number>) {
+  actual fun setRasterOpacity(opacity: FloatExpression) {
     impl.setProperties(PropertyFactory.rasterOpacity(opacity.toMLNExpression()))
   }
 
-  actual fun setRasterHueRotate(hueRotate: Expression<Number>) {
+  actual fun setRasterHueRotate(hueRotate: IntExpression) {
```
westnordost, line 21:
```suggestion
  actual fun setRasterHueRotate(hueRotate: FloatExpression) {
```

see https://github.com/maplibre/maplibre-native/blob/72d2c70130fcd4e320cf623f9bec811108b6e201/src/mbgl/style/layers/raster_layer_properties.hpp#L35

(even though it may uncommon, the hue is essentially an angle)

lib/maplibre-compose/src/androidMain/kotlin/dev/sargunv/maplibrecompose/core/layer/SymbolLayer.kt
```diff
@@ -33,245 +38,243 @@ internal actual class SymbolLayer actual constructor(id: String, source: Source)
...
     impl.setProperties(PropertyFactory.iconTextFitPadding(textFitPadding.toMLNExpression()))
   }
 
-  actual fun setIconImage(image: Expression<TResolvedImage>) {
+  actual fun setIconImage(image: ImageExpression) {
     impl.setProperties(PropertyFactory.iconImage(image.toMLNExpression()))
   }
 
-  actual fun setIconRotate(rotate: Expression<Number>) {
+  actual fun setIconRotate(rotate: IntExpression) {
```
westnordost, line 102:
```suggestion
  actual fun setIconRotate(rotate: FloatExpression) {
```

float in native code

https://github.com/maplibre/maplibre-native/blob/72d2c70130fcd4e320cf623f9bec811108b6e201/include/mbgl/style/layers/symbol_layer.hpp#L65

lib/maplibre-compose/src/androidMain/kotlin/dev/sargunv/maplibrecompose/core/layer/SymbolLayer.kt
```diff
@@ -33,245 +38,243 @@ internal actual class SymbolLayer actual constructor(id: String, source: Source)
...
     impl.setProperties(PropertyFactory.textRotationAlignment(rotationAlignment.toMLNExpression()))
   }
 
-  actual fun setTextField(field: Expression<TFormatted>) {
+  actual fun setTextField(field: FormattedExpression) {
     impl.setProperties(PropertyFactory.textField(field.toMLNExpression()))
   }
 
-  actual fun setTextFont(font: Expression<List<String>>) {
+  actual fun setTextFont(font: ListExpression<StringExpression>) {
```
westnordost, line 166: By the way, this *looks* odd. It reads like this is going to evaluate to `List<StringExpression>`

lib/maplibre-compose/src/androidMain/kotlin/dev/sargunv/maplibrecompose/core/layer/SymbolLayer.kt
```diff
@@ -33,245 +38,243 @@ internal actual class SymbolLayer actual constructor(id: String, source: Source)
...
     )
   }
 
-  actual fun setTextAnchor(anchor: Expression<SymbolAnchor>) {
+  actual fun setTextAnchor(anchor: EnumExpression<SymbolAnchor>) {
     impl.setProperties(PropertyFactory.textAnchor(anchor.toMLNExpression()))
   }
 
-  actual fun setTextMaxAngle(maxAngle: Expression<Number>) {
+  actual fun setTextMaxAngle(maxAngle: IntExpression) {
```
westnordost, line 208: also a float

lib/maplibre-compose/src/androidMain/kotlin/dev/sargunv/maplibrecompose/core/layer/SymbolLayer.kt
```diff
@@ -33,245 +38,243 @@ internal actual class SymbolLayer actual constructor(id: String, source: Source)
...
     impl.setProperties(PropertyFactory.textMaxAngle(maxAngle.toMLNExpression()))
   }
 
-  actual fun setTextWritingMode(writingMode: Expression<List<TextWritingMode>>) {
+  actual fun setTextWritingMode(writingMode: ListExpression<EnumExpression<TextWritingMode>>) {
     impl.setProperties(PropertyFactory.textWritingMode(writingMode.toMLNExpression()))
   }
 
-  actual fun setTextRotate(rotate: Expression<Number>) {
+  actual fun setTextRotate(rotate: IntExpression) {
```
westnordost, line 216: also a float

lib/maplibre-compose/src/commonMain/kotlin/dev/sargunv/maplibrecompose/core/expression/ExpressionScope.kt
```diff
@@ -513,116 +480,76 @@ public interface ExpressionScope {
...
-  @JvmName("eqNumber")
-  public infix fun Expression<Number>.eq(other: Expression<Number>): Expression<Boolean> =
-    callFn("==", this, other)
+  public fun <T : Expression> coalesce(vararg values: T): T = callFn("coalesce", *values)
 
   /** Returns whether this expression is equal to [other]. */
-  @JvmName("eqDp")
-  public infix fun Expression<Dp>.eq(other: Expression<Dp>): Expression<Boolean> =
-    callFn("==", this, other)
+  public infix fun Expression.eq(other: Expression): BooleanExpression = callFn("==", this, other)
```
westnordost, line 520: The actual limitation for comparisons is that the left hand side and right hand side **must** be of the same type according to the style definition ([source](https://github.com/maplibre/maplibre-style-spec/blob/main/src/expression/definitions/comparison.ts#L96C1-L96C9)). That's why I had a different overload for the different types, to ensure that.

That holds true for the other comparisons, too, by the way. `4 > "hello"` is considered invalid, too.

Also, why only Boolean, String, Number? Because equality is defined as strict equality (see [source](https://github.com/maplibre/maplibre-style-spec/blob/main/src/expression/definitions/comparison.ts#L29)), so two lists, or maps, or colors can never be equal, even if they have the same content.

lib/maplibre-compose/src/commonMain/kotlin/dev/sargunv/maplibrecompose/core/expression/Expression.kt
sargunv: Scalar is the super type of dimensionless quantities (number) and those with dimensions (dp). So a DpExpression is a FloatScalarExpression but not a FloatExpression

lib/maplibre-compose/src/commonMain/kotlin/dev/sargunv/maplibrecompose/core/expression/ExpressionScope.kt
```diff
@@ -1118,30 +920,29 @@ public interface ExpressionScope {
...
   /**
    * Gets the feature's geometry type as a string: "Point", "MultiPoint", "LineString",
    * "MultiLineString", "Polygon" or "MultiPolygon".
    */
-  public fun geometryType(): Expression<String> = callFn("geometry-type")
+  public fun geometryType(): StringExpression = callFn("geometry-type")
 
   /** Gets the feature's id, if it has one. */
-  public fun <T> id(): Expression<T> = callFn("id")
+  public fun <T> id(): T = callFn("id")
```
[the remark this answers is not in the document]
sargunv: I think I forgot `T : Expression` here

lib/maplibre-compose/src/androidMain/kotlin/dev/sargunv/maplibrecompose/core/layer/RasterLayer.kt
```diff
@@ -12,35 +14,35 @@ internal actual class RasterLayer actual constructor(id: String, actual val sour
...
     impl.setProperties(PropertyFactory.rasterContrast(contrast.toMLNExpression()))
   }
 
-  actual fun setRasterResampling(resampling: Expression<RasterResampling>) {
+  actual fun setRasterResampling(resampling: EnumExpression<RasterResampling>) {
     impl.setProperties(PropertyFactory.rasterResampling(resampling.toMLNExpression()))
   }
 
-  actual fun setRasterFadeDuration(fadeDuration: Expression<Number>) {
+  actual fun setRasterFadeDuration(fadeDuration: IntExpression) {
```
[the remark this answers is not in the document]
sargunv: I was actually thinking of adding a DurationExpression just for this property, so the user can pass like const(1.seconds)

lib/maplibre-compose/src/androidMain/kotlin/dev/sargunv/maplibrecompose/core/layer/SymbolLayer.kt
sargunv: If the native code allows the list to be produced by some expression (like getting it from feature properties) then we can't do `List<StringExpression>`

westnordost: Yes, I just stated that it *looks* odd, nevermind

lib/maplibre-compose/src/commonMain/kotlin/dev/sargunv/maplibrecompose/core/expression/ExpressionScope.kt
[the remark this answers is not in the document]
westnordost: (be sure to also look at other places, i.e. search for "): T", I spotted this in a few functions

lib/maplibre-compose/src/commonMain/kotlin/dev/sargunv/maplibrecompose/core/expression/Expression.kt
westnordost: maybe I don't understand what is meant with "dimension". You mean, a unit?

(I remember the words of my physics teacher in school: "What is the result of this equation?" - student: "It's 20" - teacher: "20 what? eggs?")

lib/maplibre-compose/src/commonMain/kotlin/dev/sargunv/maplibrecompose/core/expression/ExpressionScope.kt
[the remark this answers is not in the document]
sargunv: yup I did a find/replace for `Expression<T>` to `T` but hand edited `<T>` to `<T : Expression>`, so missed some on the latter

lib/maplibre-compose/src/commonMain/kotlin/dev/sargunv/maplibrecompose/core/expression/ExpressionImpl.kt
sargunv: `const` in the DSL has a variant for int so it didn't seem necessary, though I agree it's weird, can add it

lib/maplibre-compose/src/commonMain/kotlin/dev/sargunv/maplibrecompose/core/expression/Expression.kt
sargunv: yup exactly. scalars are anything that's represented by a single numeric value.

A scalar with a unit shouldn't be comparable with scalars of other units. Arithmetic like addition and division should match units on both sides, while arithmetic like multiplication and division may have a unit on at most one side. (I still need to fix the functions to support this).

would be nice to somehow write tests for this; in the TypeScript world I often write "tests" that certain code should fail to compile by marking it with `// @ts-expect-error`. Rust has a `compile_fail` attribute in doc tests. Kotlin, to my knowledge, has nothing for this.

westnordost:
```
@Test foo() {
  assertFails { throw RuntimeException() }
}
```

westnordost: oh nevermind, you meant fails to compile

sargunv: yup, it's really handy any time you're writing code with complex type checking. comes up often in TS and Rust, not so often in Kotlin but would be handy here

maybe https://github.com/tschuchortdev/kotlin-compile-testing can help

lib/maplibre-compose/src/commonMain/kotlin/dev/sargunv/maplibrecompose/core/expression/ExpressionScope.kt
```diff
@@ -865,200 +813,172 @@ public interface ExpressionScope {
...
 
   /** Returns the largest integer that is less than or equal to [value]. */
-  @JvmName("floorDp")
-  public fun floor(value: Expression<Dp>): Expression<Dp> = callFn("floor", value)
+  public fun floor(value: Expression<FloatValue>): Expression<IntValue> = callFn("floor", value)
 
   /** Returns the shortest distance in meters between the evaluated feature and [geometry]. */
-  public fun distance(geometry: Expression<Geometry>): Expression<Number> =
+  public fun distance(geometry: Expression<GeoJsonValue>): Expression<FloatValue> =
     callFn("distance", geometry)
```
westnordost, line 982: With the new typing, shouldn't there be a `public fun const(geoJson: GeoJson)`, then?

sargunv: We didn't have it with the old typing either, but yeah that's a missing piece rn

sargunv:
I think for bare types we'd have to overload per type yeah

I _might_ have a way around it, unsure if it'll work

sargunv:
The idea: 

```kt
  public operator fun <U, V : ScalarValue<U>> Expression<V>.plus(
    other: U,
  ): Expression<V> = callFn("+", this, constScalar(other)).cast()

  public operator fun <U, V : ScalarValue<U>> U.plus(
    other: Expression<V>,
  ): Expression<V> = callFn("+", constScalar(this), other).cast()
```

PR #120, opened 2024-12-14
refactor: make expression type casting more explicit

sargunv:
Improved type safety for expressions by removing implicit auto-casting on get and variables. Variables now have a typed definition, and get returns an unknown expression that must explicitly be cast or validated to narrow the type.

### DIFF
--- a/lib/maplibre-compose/src/commonMain/kotlin/dev/sargunv/maplibrecompose/core/expression/ExpressionScope.kt
+++ b/lib/maplibre-compose/src/commonMain/kotlin/dev/sargunv/maplibrecompose/core/expression/ExpressionScope.kt
@@ -7,6 +7,7 @@ import androidx.compose.ui.unit.Dp
 import androidx.compose.ui.unit.DpOffset
 import dev.sargunv.maplibrecompose.core.util.JsOnlyApi
 import kotlin.enums.enumEntries
+import kotlin.jvm.JvmInline
 import kotlin.jvm.JvmName
 import kotlin.time.Duration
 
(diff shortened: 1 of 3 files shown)


sargunv:
* **#120** <a href="https://app.graphite.dev/github/pr/sargunv/maplibre-compose/120?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/sargunv/maplibre-compose/120?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

lib/maplibre-compose/src/commonMain/kotlin/dev/sargunv/maplibrecompose/core/expression/ExpressionScope.kt
```diff
@@ -82,18 +83,23 @@ public object ExpressionScope {
...
   /**
-   * Binds expressions to named variables, which can then be referenced in the result expression
-   * using [useVariable].
+   * Binds expressions to variables declared by [declare], which can then be referenced with [use]
+   * in the [expression].
    */
-  public fun <T : ExpressionValue> withVariable(
-    name: String,
-    value: Expression<*>,
+  public fun <T : ExpressionValue> Variable<T>.bind(
```
westnordost, line 96: This is cool, on the one hand, on the other hand, the expressiveness with "withVariable" is lost: It is a really odd construct to have the rest of the expression be part of the declaration/binding of a variable, "withVariable" explained what's happening a bit. `variable.bind` doesn't. Also, separating the declaration and binding does give room for an additional misunderstanding/mistake  that couldn't happen if both were one: That a user declares a variable and then tries to use it, without "binding" it.

sargunv: The same misunderstanding could happen if the variable is a string 

I might have an idea to deal with this, will hack on it in a bit

sargunv: Something like:

```kt
withVariable(const(1)) { var ->
  var.value()
}
```

Only question there is how we get the string name. It's never shown to the user, so maybe auto generated? Or an optional parameter to withVariable

westnordost: hmm, I don't know. The scope should contain and *only* contain an expression, right? What happens if there is more?

sargunv: it should return an expression. I imagine it'd be a call-in-place inlined block (kind of like `let`) so adding more code there is the same as adding more code right next to the expression

sargunv: it's now part of https://github.com/sargunv/maplibre-compose/pull/133/files

```kt
val example: Expression<StringValue> = withVariable("foo", const(1)) {
  foo -> foo.use().convertToString()
}
```

PR #128, opened 2024-12-16
Sort functions in fitting regions

westnordost:
It bugged me that the string functions like length(), indexOf(), substring() etc. are not in the "String" region, so I sorted it a bit:

- remove "Lookup" section
- add sections "List", "Map", sort the functions into these
- move "String" section up, right next to "List" and "Map" and put the related functions in there

No other changes, just changed the order of functions in that file.


sargunv (approved):

sargunv:
what a strange CI error

PR #129, opened 2024-12-16
document MapLibreMap

westnordost:
This would be the last item of my documentation crusade.

This is draft because there are two open questions:

1. @sargunv sure it should be `MaplibreMap` and not `MapLibreMap`? MapLibre is usually referred to like that

2. the `onFpsChanged` listener/callback is documented in Android to actually be called *on every frame*. I am currently waiting for confirmation that the documentation is indeed correct. If it is correct, this should probably be renamed.

### DIFF
--- a/demo-app/src/commonMain/kotlin/dev/sargunv/maplibrecompose/demoapp/demos/FrameRateDemo.kt
+++ b/demo-app/src/commonMain/kotlin/dev/sargunv/maplibrecompose/demoapp/demos/FrameRateDemo.kt
@@ -36,7 +36,7 @@ object FrameRateDemo : Demo {
           modifier = Modifier.weight(1f),
           styleUri = DEFAULT_STYLE,
           maximumFps = maximumFps,
-          onFpsChanged = fpsState::recordFps,
+          onFrame = fpsState::recordFps,
         )
 
         Column(modifier = Modifier.padding(16.dp)) {
--- a/demo-app/src/commonMain/kotlin/dev/sargunv/maplibrecompose/demoapp/util.kt
+++ b/demo-app/src/commonMain/kotlin/dev/sargunv/maplibrecompose/demoapp/util.kt
@@ -64,8 +64,8 @@ internal class FrameRateState(private val spinner: String = "◐◓◑◒") {
   private var rollingAverage by mutableStateOf(0.0)
   private var spinnerIndex by mutableStateOf(0)
 
-  fun recordFps(fps: Double) {
-    rollingAverage = (rollingAverage * 0.9) + (fps * 0.1)
+  fun recordFps(framesPerSecond: Double) {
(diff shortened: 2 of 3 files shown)


sargunv:
> sure it should be MaplibreMap and not MapLibreMap? MapLibre is usually referred to like that

my preference is "MapLibre map" in prose and `MaplibreMap` in pascal case. In camel/pascal case I prefer to ignore proper capitalization and only capitalize where there would be a space in prose. Like "JSON parser" -> `JsonParser`. Ditto for snake/kebab case: `maplibre_map`, `json-parser`, etc.

sargunv:
> the onFpsChanged listener/callback is documented in Android to actually be called on every frame.

It makes sense, as FPS does change every frame (except when the map goes idle). The MapLibre SDK on iOS names it something similar if I recall, that's where I got the name.

westnordost:
How about `onFrame` or `onFrameRendered`?

sargunv:
Works for me

westnordost:
Why are all the Expression functions in this `ExpressionScope` object, by the way? From the function signature of `MaplibreMap`, it looks like one is supposed to write expressions (not Maplibre map content in general, i.e. sources, layers, anchors).

```
content: @Composable ExpressionScope.() -> Unit = {},
```

I was wondering if it is possible to limit which composables can be defined for a Composable scope, (something like `@Composable(group = "MapLibre")` but something like that doesn't seem to exist.

sargunv:
> Why are all the Expression functions in this ExpressionScope object, by the way? From the function signature of MaplibreMap, it looks like one is supposed to write expressions (not Maplibre map content in general, i.e. sources, layers, anchors).

I thought it'd be convenient to have them all included by default within the scope instead of needing to import many functions (good for IDE typeahead too). Not super attached to that though. Could rename the scope to `object Expressions` and not provide it as a receiver in the mapContent

> I was wondering if it is possible to limit which composables can be defined for a Composable scope, (something like @Composable(group = "MapLibre") but something like that doesn't seem to exist.

Kotlin has something like this for regular DSLs (`@DslMarker`?), maybe that'll work

sargunv:
> Kotlin has something like this for regular DSLs (@DslMarker?), maybe that'll work

that said, other `@Composable` APIs like Redwood and Molecule don't bother with this (not to mention Compose UI itself)

westnordost:
> not to mention Compose UI itself

Are there several different composable types that are not compatible with each other in Compose UI.

Anyway, was just an idea.

sargunv:
> Are there several different composable types that are not compatible with each other in Compose UI?

oh no, just meant that it's a composition incompatible with others like Redwood, Molecule, and ours

that said, I just dig a bit deeper and found this in Compose UI

CleanShot 2024-12-16 at 15 46 12@2x

PR #130, opened 2024-12-16
added some local style examples

westnordost:
(from versatiles.org)


westnordost:
(Shall I remove all linebreaks from the styles?)

PR #131, opened 2024-12-17
feat: add material3 attribution button and dialog

sargunv:
Key changes:
- Created new `AttributionButton` and `AttributionDialog` components
- Added `StyleState` to track and expose style-related information
- Implemented attribution link extraction on Android and iOS

The dialog doesn't extract non-link text from the attribution. This is mirroring the Android SDK's functionality. The iOS SDK's attribution dialog displays non-link text as a button that doesn't work, so I figured it's probably not important.

CleanShot 2024-12-16 at 19 21 00@2x

### DIFF
--- a/lib/maplibre-compose/src/androidMain/kotlin/dev/sargunv/maplibrecompose/core/source/Source.kt
+++ b/lib/maplibre-compose/src/androidMain/kotlin/dev/sargunv/maplibrecompose/core/source/Source.kt
@@ -1,11 +1,27 @@
 package dev.sargunv.maplibrecompose.core.source
 
+import android.text.Html
+import android.text.style.URLSpan
 import org.maplibre.android.style.sources.Source as MLNSource
 
 public actual sealed class Source {
   internal abstract val impl: MLNSource
   internal actual val id: String
     get() = impl.id
+
+  public actual val attributionLinks: List<AttributionLink>
+    get() {
+      // TODO minSdk 24 to get rid of deprecation warning
+      @Suppress("DEPRECATION") val spanned = Html.fromHtml(impl.attribution)
+
+      val spans = spanned.getSpans(0, spanned.length, URLSpan::class.java)
+      return spans.map {
+        AttributionLink(
+          title = spanned.slice(spanned.getSpanStart(it)..<spanned.getSpanEnd(it)).toString(),
+          url = it.url,
+        )
+      }
+    }
 
   override fun toString(): String = "${this::class.simpleName}(id=\"$id\")"
 }
--- a/lib/maplibre-compose/src/commonMain/kotlin/dev/sargunv/maplibrecompose/compose/controls/AttributionDialog.kt
+++ b/lib/maplibre-compose/src/commonMain/kotlin/dev/sargunv/maplibrecompose/compose/controls/AttributionDialog.kt
@@ -1,0 +1,66 @@
+package dev.sargunv.maplibrecompose.compose.controls
+
+import androidx.compose.foundation.layout.Column
+import androidx.compose.foundation.layout.padding
+import androidx.compose.foundation.layout.widthIn
+import androidx.compose.foundation.layout.wrapContentHeight
+import androidx.compose.foundation.layout.wrapContentWidth
+import androidx.compose.material3.AlertDialogDefaults
+import androidx.compose.material3.BasicAlertDialog
+import androidx.compose.material3.ExperimentalMaterial3Api
+import androidx.compose.material3.MaterialTheme
+import androidx.compose.material3.Surface
+import androidx.compose.material3.Text
+import androidx.compose.material3.TextButton
+import androidx.compose.runtime.Composable
+import androidx.compose.runtime.remember
+import androidx.compose.ui.Alignment
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.platform.LocalUriHandler
+import androidx.compose.ui.semantics.heading
+import androidx.compose.ui.semantics.semantics
+import androidx.compose.ui.unit.dp
+import dev.sargunv.maplibrecompose.compose.StyleState
+
+@Composable
+@OptIn(ExperimentalMaterial3Api::class)
+public fun AttributionDialog(
+  styleState: StyleState,
+  onDismiss: () -> Unit,
+  titleText: String = "Data sources",
+  buttonText: String = "OK",
+) {
+  val attributions = remember(styleState) { styleState.queryAttributionLinks() }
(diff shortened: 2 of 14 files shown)


sargunv:
* **#131** <a href="https://app.graphite.dev/github/pr/sargunv/maplibre-compose/131?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/sargunv/maplibre-compose/131?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

lib/maplibre-compose/src/androidMain/kotlin/dev/sargunv/maplibrecompose/core/source/Source.kt
```diff
@@ -1,11 +1,27 @@
...
 
 public actual sealed class Source {
   internal abstract val impl: MLNSource
   internal actual val id: String
     get() = impl.id
 
+  public actual val attributionLinks: List<AttributionLink>
+    get() {
+      // TODO minSdk 24 to get rid of deprecation warning
+      @Suppress("DEPRECATION") val spanned = Html.fromHtml(impl.attribution)
```
westnordost, line 15: Actually, Compose does support creating annotated strings directly since v1.7.0, see https://developer.android.com/reference/kotlin/androidx/compose/ui/text/package-summary#(androidx.compose.ui.text.AnnotatedString.Companion).fromHtml(kotlin.String,androidx.compose.ui.text.TextLinkStyles,androidx.compose.ui.text.LinkInteractionListener)

This is **currently** an Android-only API. There's no reason for it to eventually be multiplatform, though, I'd say. Jetbrains people just need to implement an own simple multiplatform HTML parser (, which is actually not that much work).

For my project, I had to do this (because I was using Compose 1.6 but also because at some other place I take apart an HTML and split it into sections):

- **Parser**: https://github.com/streetcomplete/StreetComplete/tree/master/app/src/main/java/de/westnordost/streetcomplete/util/html
- **Html nodes to Annotated string**: https://github.com/streetcomplete/StreetComplete/blob/bf52d5ff17d3c500257e72bf1ee89ed5e17c508a/app/src/main/java/de/westnordost/streetcomplete/ui/util/Html.kt

Without tests and given a helper class for iterating through a string, that's just about 2x 170 lines of code.

I'd absolutely be happy to change the license to contribute that upstream to Jetbrains, but I don't really know where I can do that.

Anyway, I do realize this is a bit off topic, because the MapLibre SDK for iOS already provides the parsed attribution links, so parsing HTML is only necessary for Android (at the moment, i.e. disregarding supporting  JS / Desktop in the future). Although, of course, this is also relevant for when/if eventually migrating to use a TBD C Api directly instead of Android/iOS SDKs.

lib/maplibre-compose/src/commonMain/kotlin/dev/sargunv/maplibrecompose/compose/controls/AttributionDialog.kt
```diff
@@ -0,0 +1,66 @@
...
+import androidx.compose.ui.unit.dp
+import dev.sargunv.maplibrecompose.compose.StyleState
+
+@Composable
+@OptIn(ExperimentalMaterial3Api::class)
+public fun AttributionDialog(
+  styleState: StyleState,
+  onDismiss: () -> Unit,
+  titleText: String = "Data sources",
+  buttonText: String = "OK",
```
westnordost, line 31: I'd say, this library needs translations, now. (Maybe the Android library already has some, which can be copied into this project?)

When the text is properly translated, I don't even think it makes sense to expose these parameters anymore.

lib/maplibre-compose/src/commonMain/kotlin/dev/sargunv/maplibrecompose/compose/controls/AttributionDialog.kt
```diff
@@ -0,0 +1,66 @@
...
+) {
+  val attributions = remember(styleState) { styleState.queryAttributionLinks() }
+  val uriHandler = LocalUriHandler.current
+
+  BasicAlertDialog(
```
westnordost, line 36: This alert dialog definition looks unnecessarily complex (with the creation of a surface with hardcoded width, hardcoded padding, text styles etc.). I only used material2 AlertDialogs so far, but looking at 

https://developer.android.com/develop/ui/compose/components/dialog#alert

it looks like you could use the `AlertDialog` composable.

lib/maplibre-compose/src/androidMain/kotlin/dev/sargunv/maplibrecompose/core/source/Source.kt
sargunv: Yup, I tried exactly that and then found it's Android only. 

Ideally I'd like to provide the raw attribution string here in core, and let the compose layer parse the HTML to an annotated string. Probably something to leave for when migrating to wrapping the native SDK directly. 

For now, I matched the iOS functionality that parses out attribution (minus text outside links, because the Android SDK seems to ignore it anyway and the iOS SDK handles it poorly)

lib/maplibre-compose/src/commonMain/kotlin/dev/sargunv/maplibrecompose/compose/controls/AttributionDialog.kt
sargunv: Would it be easy for the consuming app to provide translations for languages the library itself doesn't support? If so then yeah, agreed. Having them as parameters is just a quick hack to provide that flexibility now

westnordost: I don't know the compose resource system, but in Android resource system, it wouldn't be possible.

lib/maplibre-compose/src/commonMain/kotlin/dev/sargunv/maplibrecompose/compose/controls/AttributionDialog.kt
sargunv: Ah, I saw `AlertDialog` was deprecated but there's an `expect fun` overload that's not deprecated. Will check it out

sargunv: https://github.com/sargunv/maplibre-compose/pull/138

PR #134, opened 2024-12-17
feat: support text scaling (sp instead of dp)

sargunv:
Add new `ScalarValue<TextUnit>` and `const` functions for the unit. There's a `@Composable` one to provide font scaling automatically. If the provided text unit is in ems, it's considered relative to 16sp, the default font size of the map.

### DIFF
--- a/lib/maplibre-compose/src/commonMain/kotlin/dev/sargunv/maplibrecompose/core/expression/ExpressionsDsl.kt
+++ b/lib/maplibre-compose/src/commonMain/kotlin/dev/sargunv/maplibrecompose/core/expression/ExpressionsDsl.kt
@@ -1178,7 +1178,7 @@ public object ExpressionsDsl {
   // region Feature data
 
   /** Object to access feature-related data, see [feature] */
-  public object FeatureScope {
+  public object Feature {
     /**
      * Returns the value corresponding to the given [key] in the current feature's properties or
      * `null` if it is not present.
@@ -1239,7 +1239,7 @@ public object ExpressionsDsl {
   }
 
   /** Accesses to feature-related data */
-  public val feature: FeatureScope = FeatureScope
+  public val feature: Feature = Feature
 
   // endregion
 
--- a/lib/maplibre-compose/src/commonMain/kotlin/dev/sargunv/maplibrecompose/core/expression/ExpressionsDsl.kt
+++ b/lib/maplibre-compose/src/commonMain/kotlin/dev/sargunv/maplibrecompose/core/expression/ExpressionsDsl.kt
@@ -1,10 +1,15 @@
 package dev.sargunv.maplibrecompose.core.expression
 
 import androidx.compose.foundation.layout.PaddingValues
+import androidx.compose.runtime.Composable
 import androidx.compose.ui.geometry.Offset
 import androidx.compose.ui.graphics.Color
+import androidx.compose.ui.platform.LocalDensity
+import androidx.compose.ui.unit.Density
 import androidx.compose.ui.unit.Dp
 import androidx.compose.ui.unit.DpOffset
+import androidx.compose.ui.unit.TextUnit
+import androidx.compose.ui.unit.TextUnitType
 import dev.sargunv.maplibrecompose.core.util.JsOnlyApi
 import kotlin.enums.enumEntries
 import kotlin.jvm.JvmInline
@@ -28,6 +33,27 @@ public object ExpressionsDsl {
 
   /** Creates a literal expression for a [Dp] value. */
   public fun const(dp: Dp): Expression<DpValue> = Expression.ofFloat(dp.value).cast()
+
+  /**
+   * Creates a literal expression for a specified [TextUnit] value. If [textUnit] is in `em`, it is
+   * considered relative to `16.sp`.
+   */
+  public fun const(textUnit: TextUnit, fontScale: Float): Expression<TextUnitValue> =
+    when (textUnit.type) {
+      TextUnitType.Sp -> const(textUnit.value * fontScale).cast()
+      TextUnitType.Em -> const(textUnit.value * fontScale * 16f).cast()
+      else -> error("Unsupported TextUnit type: ${textUnit.type}")
+    }
+
+  /**
+   * Creates a literal expression for a specified [TextUnit] value. If [textUnit] is in `em`, it is
+   * considered relative to `16.sp` (the default symbol layer text size).
+   */
+  @Composable
+  public fun const(
+    textUnit: TextUnit,
+    density: Density = LocalDensity.current,
+  ): Expression<TextUnitValue> = const(textUnit, density.fontScale)
 
   /**
    * Creates a literal expression for a [Duration] value.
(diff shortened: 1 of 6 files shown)


sargunv:
* **#134** <a href="https://app.graphite.dev/github/pr/sargunv/maplibre-compose/134?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/sargunv/maplibre-compose/134?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

lib/maplibre-compose/src/commonMain/kotlin/dev/sargunv/maplibrecompose/core/expression/ExpressionsDsl.kt
```diff
@@ -29,6 +34,27 @@ public object ExpressionsDsl {
   /** Creates a literal expression for a [Dp] value. */
   public fun const(dp: Dp): Expression<DpValue> = Expression.ofFloat(dp.value).cast()
 
+  /**
+   * Creates a literal expression for a specified [TextUnit] value. If [textUnit] is in `em`, it is
+   * considered relative to `16.sp`.
+   */
+  public fun const(textUnit: TextUnit, fontScale: Float): Expression<TextUnitValue> =
```
westnordost, line 41: Why should it be possible for the user to pass the `Density`, yet alone the `fontScale` himself at all?

westnordost (commented):
Hmm, there are a few problems with this:

1. When https://github.com/maplibre/maplibre-native/issues/3057 is implemented, this very likely needs to be redone because if it is implemented, the sp value could just be passed on. (So at least the issue should be linked somewhere)

2. There are more properties which are defined in EMs: `textLetterSpacing`, `textMaxWidth`, `textLineHeight`, `textOffset`, `textRadialOffset` and parts of `textVariableAnchorOffset`. I'd say it would be consistent that if the `TextUnit` type is used for `textSize`, it should be used for these other properties, too. As a `TextUnit` can always also be specified in `sp`, the value in EMs would need to be calculated if it is given in `sp` in relation to the font size. I.e.: `textLineHeight in ems = textLineHeight in sp / textSize in sp`. This would need to be recalculated for all these properties each time `textSize` is changed.

2. **As far as I know** (Compose documentation isn't really helping), `1.em` text size means "100% text size in relation to the current text style". So, if the `LocalTextStyle` has a font size of `10.sp`, then `1.em` is also `10.sp`.

I.e. `textSize` in ems is relative to the `LocalTextStyle`, the other properties with em as a unit are relative to the `textSize`.

lib/maplibre-compose/src/commonMain/kotlin/dev/sargunv/maplibrecompose/core/expression/ExpressionsDsl.kt
sargunv: In the non composable function, because LocalDensity isn't available

in the composable version, prob not necessary yeah.

sargunv:
(1) Fully expecting to redo this if/when scaling is supported in native. Will link the issue

(2) I don't think supporting TextUnit is possible for those properties. a `ScalarValue<TextUnit>` is internally represented as SP. If it was initially provided as EM, we would divide it by the textSize expression, but if it was initially provided as SP, we wouldn't. That information is lost at the time of the EM -> SP conversion.

I think this is better thought of as an "SP unit" rather than a "TextUnit unit". To reduce confusion, we _could_ update the documentation to say "in multiples of the text size" instead of "in ems".

(3) I'm also unsure, but in any case the documentation for `TextStyle` and `LocalTextStyle` refers specifically to a `Text` component. I think it's reasonable for a map component to have its own internal text style, and the value passed in here is merely configuration for that style. Or in other words, imagine a component like this:

```
@Composable
fun fun App() {
  FakeMap(fontSize = 1.em)
}

@Composable
fun FakeMap(fontSize: TextUnit) {
  ProvideTextStyle(someTextStyle) {
    Text("example", fontSize = fontSize)
  }
}
```

Even in that component, the font size passed in is relative to the text style internal to `FakeMap`, not the local text style outside that component. Our map (or map layer) is essentially that.

https://github.com/sargunv/maplibre-compose/pull/136 makes some tweaks

sargunv:
There's one more issue, that symbol layers defined in the base JSON style don't get scaled. Technically fixable, but I think it'd be weird to redefine the meaning of the json style spec, so I think it's better to leave that API as DP, while making our composable API support SP.

sargunv:
Hmm, I may have a workable idea to make all the sp/em properties support TextUnit fully. What if `TextUnitValue` wasn't a `ScalarValue<TextUnit>`, but rather a direct descendant of `ExpressionValue`? And then it was backed by a class, with a type discriminator and functions to map it to an dp/em float expression? SymbolLayer's composable could do that mapping before passing it to the SymbolLayer wrapper class (passing the textSize expression into the em converters)

```kt
public class TextUnitValue internal constructor(
  internal val expr: Expression<FloatValue>,
  internal val type: TextUnitType,
) :
  ExpressionValue,
  MatchableValue,
  InterpolateableValue<TextUnit>,
  ComparableValue<TextUnit>,
  EquatableValue {
  internal fun toDpExpression(
    textSizeExpression: Expression<FloatValue>,
    fontScale: Float,
  ): Expression<FloatValue> {
    return when (type) {
      TextUnitType.Sp -> expr * const(fontScale)
      TextUnitType.Em -> expr * textSizeExpression * const(fontScale)
      else -> error("Unsupported text unit type: $type")
    }
  }

  internal fun toEmExpression(textSizeExpression: Expression<FloatValue>): Expression<FloatValue> {
    return when (type) {
      TextUnitType.Sp -> expr / textSizeExpression
      TextUnitType.Em -> expr
      else -> error("Unsupported text unit type: $type")
    }
  }
}
```

westnordost:
> There's one more issue, that symbol layers defined in the base JSON style don't get scaled. Technically fixable, but I think it'd be weird to redefine the meaning of the json style spec, so I think it's better to leave that API as DP, while making our composable API support SP. Upstream will need to figure out how to handle this (breaking change to the style spec? new scale property?) too, and we can follow their lead rather than solving it ourselves.

on web, I think there is not even really the notion of sp at all, so its natural that the spec doesn't really mention this.

Instead, every browser has its own way to deal with this. E.g. IIRC if you scale your font size to 200% on Android, *all* web content is displayed at 200% (including MapLibre). So, the accessibility font size *is* respected, and in MapLibre web automatically, too, but does more than just to scale the text.

lib/maplibre-compose/src/commonMain/kotlin/dev/sargunv/maplibrecompose/core/expression/ExpressionsDsl.kt
westnordost: (Uf, wrong button and the typed text is gone)

I overread that one function is composable, the other not. But I'd say, this would only ever be called in composable functions, wouldn't it?

sargunv:
Some data points:

* Google Maps on Android scales text on the map independently of the rest of the map style.
* Google Maps on web (Chromium on Android) doesn't respond to the accessibility font scale setting at all
* MapLibre on web (same browser) doesn't respond to the accessibility font scale setting at all

I haven't tried desktop yet, not checked what Compose- does on web / desktop / iOS. 

https://github.com/user-attachments/assets/f9587661-39f8-413a-8081-a07d0859f921


https://github.com/user-attachments/assets/d05998e4-5a27-4ad1-b784-32bb37b8c9a8


https://github.com/user-attachments/assets/3d4e03bc-67c6-4ca2-aa6e-ae9016f7d65c

PR #140, opened 2024-12-18
feat: support TextUnit on most sp/em properties

sargunv:
textOffset isn't supported yet; gotta think more on how to handle that one as apparently expressions aren't allowed inside list literals

### DIFF
--- a/lib/maplibre-compose/src/commonMain/kotlin/dev/sargunv/maplibrecompose/compose/layer/SymbolLayer.kt
+++ b/lib/maplibre-compose/src/commonMain/kotlin/dev/sargunv/maplibrecompose/compose/layer/SymbolLayer.kt
@@ -1,6 +1,7 @@
 package dev.sargunv.maplibrecompose.compose.layer
 
 import androidx.compose.runtime.Composable
+import androidx.compose.runtime.remember
 import androidx.compose.ui.geometry.Offset
 import androidx.compose.ui.graphics.Color
 import androidx.compose.ui.platform.LocalDensity
@@ -18,6 +19,7 @@ import dev.sargunv.maplibrecompose.core.expression.EnumValue
 import dev.sargunv.maplibrecompose.core.expression.Expression
 import dev.sargunv.maplibrecompose.core.expression.ExpressionsDsl.cast
 import dev.sargunv.maplibrecompose.core.expression.ExpressionsDsl.const
+import dev.sargunv.maplibrecompose.core.expression.ExpressionsDsl.dp
 import dev.sargunv.maplibrecompose.core.expression.ExpressionsDsl.nil
 import dev.sargunv.maplibrecompose.core.expression.ExpressionsDsl.times
 import dev.sargunv.maplibrecompose.core.expression.FloatValue
@@ -29,7 +31,6 @@ import dev.sargunv.maplibrecompose.core.expression.ImageValue
 import dev.sargunv.maplibrecompose.core.expression.ListValue
 import dev.sargunv.maplibrecompose.core.expression.OffsetValue
 import dev.sargunv.maplibrecompose.core.expression.PaddingValue
-import dev.sargunv.maplibrecompose.core.expression.SpValue
 import dev.sargunv.maplibrecompose.core.expression.StringValue
 import dev.sargunv.maplibrecompose.core.expression.SymbolAnchor
 import dev.sargunv.maplibrecompose.core.expression.SymbolPlacement
@@ -38,6 +39,7 @@ import dev.sargunv.maplibrecompose.core.expression.TextJustify
 import dev.sargunv.maplibrecompose.core.expression.TextPitchAlignment
 import dev.sargunv.maplibrecompose.core.expression.TextRotationAlignment
 import dev.sargunv.maplibrecompose.core.expression.TextTransform
+import dev.sargunv.maplibrecompose.core.expression.TextUnitValue
 import dev.sargunv.maplibrecompose.core.expression.TextWritingMode
 import dev.sargunv.maplibrecompose.core.expression.TranslateAnchor
 import dev.sargunv.maplibrecompose.core.expression.ZeroPadding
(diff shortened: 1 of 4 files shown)


sargunv:
* **#140** <a href="https://app.graphite.dev/github/pr/sargunv/maplibre-compose/140?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/sargunv/maplibre-compose/140?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

lib/maplibre-compose/src/commonMain/kotlin/dev/sargunv/maplibrecompose/compose/layer/SymbolLayer.kt
```diff
@@ -504,17 +520,39 @@ public fun SymbolLayer(
       set(iconTranslate) { layer.setIconTranslate(it) }
       set(iconTranslateAnchor) { layer.setIconTranslateAnchor(it) }
 
+      set(textSizeSp to fontScale) { (textSizeSp, fontScale) ->
+        layer.setTextSize((textSizeSp * const(fontScale)).dp)
+      }
```
westnordost, line 525: I'd use `Pair(textSizeSp, fontScale)` for clarity (the `to` metaphor doesn't really work here), or, even better (IMO), use the `listOf` API consistently here when more than one parameter is involved.

(I do wonder why the `set` method doesn't simply have a vararg parameter)

westnordost (commented):
API-wise, this looks pretty cool-

lib/maplibre-compose/src/commonMain/kotlin/dev/sargunv/maplibrecompose/compose/layer/SymbolLayer.kt
sargunv: I'm gonna use remember instead, and pass a single value in here. 

Simplified reimplementation coming in a bit

lib/maplibre-compose/src/commonMain/kotlin/dev/sargunv/maplibrecompose/compose/layer/SymbolLayer.kt
```diff
@@ -428,27 +430,27 @@ public fun SymbolLayer(
...
+  textLineHeight: Expression<TextUnitValue> = const(1.2f.em),
   textJustify: Expression<EnumValue<TextJustify>> = const(TextJustify.Center),
   textWritingMode: Expression<ListValue<EnumValue<TextWritingMode>>> = nil(),
   textKeepUpright: Expression<BooleanValue> = const(true),
   textRotate: Expression<FloatValue> = const(0f),
 
   // text anchoring
   textAnchor: Expression<EnumValue<SymbolAnchor>> = const(SymbolAnchor.Center),
-  textOffset: Expression<OffsetValue> = const(Offset.Zero),
+  textOffset: Expression<OffsetValue> = const(Offset.Zero), // TODO text unit type
```
westnordost, line 451: open TODO. (Didn't I read something about offsetX, offsetY just then? Also, there is `TextOffsetValue` but it is not used here?)

sargunv: It turns out MLN doesn't support expressions inside list literals, so we can't split it into two expressions.

I'm updating the `literal` functions and then I'll try to find a new approach for dealing with the TextUnitOffset. Not sure if it'll be possible (can we divide an sp offset by an expression to get em?)

westnordost:
Oh, **important**:

I noticed that `textVariableAnchorOffset` also has offsets in EM. The type is basically 

`List<Pair<SymbolAnchor, TextUnitOffset>>`

lib/maplibre-compose/src/commonMain/kotlin/dev/sargunv/maplibrecompose/compose/layer/SymbolLayer.kt
westnordost: if all fails, an option would be to revert back to `Float` for "EM" fields ¯\_(ツ)_/¯

sargunv:
Yeah, I might consider that a `Map<SymbolAnchor< Expression<TextOffsetValue>>` and render the map as a list internally

sargunv:
https://github.com/sargunv/maplibre-compose/pull/142 adds support for TextUnit in textOffset

sargunv:
### Merge activity

* **Dec 18, 3:20 PM PST**: A user merged this pull request with [Graphite](https://app.graphite.dev/github/pr/sargunv/maplibre-compose/140).

lib/maplibre-compose/src/commonMain/kotlin/dev/sargunv/maplibrecompose/compose/layer/SymbolLayer.kt
sargunv: It worked, #142 has it

sargunv: We convert from TextUnit to sp or em by multiplying it with some scalar coming from a variable bound in the symbol layer composable. To convert a 2D text unit, we need to multiply both components (x and y) with the scalar. We can't do expressions inside the list of x and y, so we need some way to multiply an entire list of numbers by some scalar.

MapLibre expressions don't provide this directly, but linear interpolation with stops above/below our scalar is effectively the same thing as multiplication by a scalar. And interpolation works on lists of numbers.

So if you assume, say, the SP to/from EM conversion will always multiply by a value in range 0..1000, we can define those stops on a linear interpolation, input the scalar, and we've got runtime conversion between SP/EM for 2d text units

PR #155, opened 2024-12-20
feat: add state for meters per dp at camera target

sargunv:
going to use this for a composable scale bar

unsure if density calculation is correct; will confirm with a scale bar implementation before release

### DIFF
--- a/lib/maplibre-compose/src/commonMain/kotlin/dev/sargunv/maplibrecompose/compose/CameraState.kt
+++ b/lib/maplibre-compose/src/commonMain/kotlin/dev/sargunv/maplibrecompose/compose/CameraState.kt
@@ -40,6 +40,7 @@ public class CameraState internal constructor(firstPosition: CameraPosition) {
 
   internal val positionState = mutableStateOf(firstPosition)
   internal val moveReasonState = mutableStateOf(CameraMoveReason.NONE)
+  internal val metersPerDpAtTargetState = mutableStateOf(0.0)
 
   /** how the camera is oriented towards the map */
   // if the map is not yet initialized, we store the value to apply it later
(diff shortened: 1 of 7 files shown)


sargunv:
* **#155** <a href="https://app.graphite.dev/github/pr/sargunv/maplibre-compose/155?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/sargunv/maplibre-compose/155?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

lib/maplibre-compose/src/commonMain/kotlin/dev/sargunv/maplibrecompose/compose/CameraState.kt
```diff
@@ -54,6 +55,10 @@ public class CameraState internal constructor(firstPosition: CameraPosition) {
   public val moveReason: CameraMoveReason
     get() = moveReasonState.value
 
+  /** meters per dp at the target position */
+  public val metersPerDpAtTarget: Double
```
westnordost, line 59: Small suggestion: Rename to `metersPerDp`? The `CameraState` already represents, well, the current state of the camera, including `position`, so it should be clear from context what is meant.
(Also, "target position" is mentioned nowhere else)

sargunv: I wanted to clarify a bit because the meters per dp varies by longitude, pretty significantly if zoomed out. So this is the scale at the center of the view (if camera padding is zero). Or at the target of the view.

Happy to rename it though. we should clarify in the docs regardless of name

sargunv: Separately, I made the scale bar work like the MLN one, showing scale at the target, even though the bar is in the corner. 

But I think it's more intuitive for the bar to show scale exactly where it's at, as at least when I use a map I pan something I want to measure over to the scale bar to measure it. This is what the Google scale bar does, but I dropped that functionality for simplicity (and to test out this state).

sargunv: So there will likely be additional functions "meters per dp at longitude" and "meters per dp at screen location". Whatever naming we decide for this state should probably be consistent with those functions too

PR #170, opened 2024-12-23
add localized abbreviations for m/km

westnordost:
Researched by looking at all articles in all languages linked from https://en.wikipedia.org/wiki/Metre 😅

Only added strings.xml for languages that have an ISO 639-1 or ISO 639-2 code. (I.e. not ISO 639-3 because[ that isn't supported anyway](https://www.jetbrains.com/help/kotlin-multiplatform-dev/compose-multiplatform-resources-setup.html#language-and-regional-qualifiers))

"m" should actually be quite understood anywhere, but some languages have an abbreviation in their own script that is preferred over the international symbol ("m").

There are no translations for feet, because feet is only really used in the US, i.e. in an English-speaking country. Hence, in other languages, an abbreviation doesn't really exist.

### DIFF
--- a/lib/maplibre-compose-material3/src/commonMain/composeResources/values-ar/strings.xml
+++ b/lib/maplibre-compose-material3/src/commonMain/composeResources/values-ar/strings.xml
@@ -1,4 +1,4 @@
 <resources>
-  <string name="meters_abbreviation">متر</string>
-  <string name="kilometers_abbreviation">كم</string>
+  <string name="meters_symbol">متر</string>
+  <string name="kilometers_symbol">كم</string>
 </resources>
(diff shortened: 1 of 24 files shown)


sargunv:
thanks for this!

I went ahead and removed japanese (as it's the same m and km as english) and added pies for spanish

westnordost:
Well, I added Japanese for the comment. To denote "yes, it has been checked, but it is really m/km"

sargunv:
In that case I’d do the same for Korean, Hindi, Telegu, just about everything with a non Latin script that uses Latin style symbols for SI units. Is there a particular reason to comment for Japanese specifically?

westnordost:
It was specifically mentioned in the wiki article (that till X, some character was used bit it isn't anymore)

PR #171, opened 2024-12-24
A few fixes for the scalebar

westnordost:
1. fix RTL
2. fix line didn't scale with density
3. consistent look of text with scale bar (use white halo for text)


westnordost:
By the way, the scalebar doesn't behave like the Google Maps (or OpenStreetMap, etc.) scalebar in that it doesn't scale the scalebar but always shows the exact width of the scalebar, also intra-zoom.

I think a scalebar as seen in the google maps app is better, it's much less distractive. Would you agree?

sargunv:
> I think a scalebar as seen in the google maps app is better, it's much less distractive. Would you agree?

I have no real preference between the two kinds. I'd accept a PR to make it scale the bar like the one seen in Google Maps. It should be able to scale in the proper direction depending on where it's aligned (the MapLibre Native iOS scale bar does this, but the Android plugin one doesn't).

sargunv:
* **#171** <a href="https://app.graphite.dev/github/pr/sargunv/maplibre-compose/171?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/sargunv/maplibre-compose/171?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

Issue #181, opened 2024-12-24
Expression AST improvements


sargunv:
The last few releases have focused on improving the API surface for expressions, and so the underlying implementation has gotten complex. We should probably refactor this into something more maintainable.

## Current state

Expressions right now have multiple AST representations in this library:
* Level 0: The platform specific AST defined by the MapLibre SDK
  * on Android, that's a json-like tree of `JsonElement` from Gson
  * on iOS, that's a json-like tree of List, Map, String, primitives, UIColor, CGVector, and UIEdgeInsets
  * Web will be JSON, so probably comparable to Android
* Level 1: The "resolved" multiplatform AST mirroring the lowest common denominator of the platform specific ASTs
  * That's List, Map, primitives, String, Color, Offset, and PaddingValues.Absolute
* Level 2: The "unresolved" multiplatform AST. This is mostly the same as Level 2, but with two variants:
  * 2a: properties that accept TextUnit expressions include references to special variables that help convert between SP and EM in the expression. These variables are unbound in this AST level.
  * 2b: properties that accept Formatted or Image expressions allow ImageBitmap (and soon Painter) values in the tree.
* Level 3: The "api surface" AST
  * This adds additional types like enums, Duration, Dp, DpOffset, Vector, and distinguishes Int vs Float
  * This AST doesn't actually exist; the constructors immediately convert literals to the corresponding Level 2 literal. But the `ExpressionValue` type system is defined at this level.

Converting values from level 3 to 2 is usually trivial (`Dp.value` for example) and done at expression instantiation. Converting from 2 to 1 is only done by particular properties (those that make use of images or TextUnit) within the Layer composables. Converting from 1 to 0 is done in platform specific source sets, within the Layer wrapper setters.

Resolving 2a expressions (TextUnit) requires some context to bind the SP/EM conversion factor variables. For SP properties (`textSize`) the conversion factor from EM to SP is just the default font size (`const(16.sp)`). For EM properties (all others), the conversion factor from SP to EM is the inverse of the textSize expression.

Resolving 2b expressions (Image) requires `LocalStyleNode` context to register the image with the style if it hasn't yet been registered, and retrieve the generated ID.

Reading what I've written, it's clear to me that what we've written is essentially a simple compiler or transpiler. We have our high level language with rich types (level 3), a universal IR (level 1), and the output "binary" (level 0). 

## Problems

[...]

PR #191, opened 2024-12-26
refactor: strongly typed expression AST

sargunv:
This PR refactors the expression implementation to create a strongly typed AST:
* Now, an `Expression<T>` refers to a "level 3" AST as described in #181
* A `CompiledExpression<T>` refers to a "level 1" AST as described in that isse
* "Level 2" as described in that issue no longer exists
* `Expression` and `CompiledExpression` are sealed interfaces without use of `Any?`, so the conversion from level 1 to 0 has exhaustiveness checking

Additionally, I moved the expression stuff into its own module so it can be used as part of a style generator or something.

And I split up some giant files into smaller ones, and grouped some things into packages (which makes this a breaking change)


sargunv:
* **#191** <a href="https://app.graphite.dev/github/pr/sargunv/maplibre-compose/191?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/sargunv/maplibre-compose/191?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

lib/maplibre-compose-expressions/src/commonMain/kotlin/dev/sargunv/maplibrecompose/expression/ExpressionContext.kt
```diff
@@ -0,0 +1,34 @@
+package dev.sargunv.maplibrecompose.expression
+
+import androidx.compose.ui.graphics.ImageBitmap
+import androidx.compose.ui.graphics.painter.Painter
+import dev.sargunv.maplibrecompose.expression.ast.Expression
+import dev.sargunv.maplibrecompose.expression.value.FloatValue
+
+public interface ExpressionContext {
```
westnordost: I have no idea what this is for, maybe add an explanation since this is public?

lib/maplibre-compose-expressions/src/commonMain/kotlin/dev/sargunv/maplibrecompose/expression/dsl/inputs.kt
```diff
@@ -0,0 +1,20 @@
+package dev.sargunv.maplibrecompose.expression.dsl
+
+import dev.sargunv.maplibrecompose.expression.ast.Expression
+import dev.sargunv.maplibrecompose.expression.ast.FunctionCall
+import dev.sargunv.maplibrecompose.expression.value.FloatValue
+
+/**
+ * Gets the current zoom level. Note that in layer style properties, [zoom] may only appear as the
+ * input to a top-level [step] or [interpolate] (, [interpolateHcl], [interpolateLab], ...)
+ * expression.
```
westnordost: does the linking to these functions continue to work in the IDE, now that these functions are not in the same file anymore?

lib/maplibre-compose-expressions/src/commonMain/kotlin/dev/sargunv/maplibrecompose/expression/dsl/ramps.kt
```diff
@@ -0,0 +1,142 @@
...
+ *
+ * interpolates exponentially from 1 to 256 in zoom levels 16 to 24. Below zoom 16, it is 1, above
+ * zoom 24, it is 256. Applied to for example line width, this has the visual effect that the line
+ * stays the same width in meters on the map (rather than on the viewport).
+ */
+public fun <T, V : InterpolateableValue<T>> interpolate(
+  type: Expression<InterpolationValue>,
+  input: Expression<FloatValue>,
+  vararg stops: Pair<Number, Expression<V>>,
+): Expression<V> = interpolateImpl("interpolate", type, input, *stops)
```
westnordost: Now that the interpolation type continues to be not a sealed class, and probably that will stay that way now(?), the documentation comments should mention which functions produce an `Expression<InterpolationValue>`

lib/maplibre-compose-expressions/src/commonMain/kotlin/dev/sargunv/maplibrecompose/expression/dsl/types.kt
```diff
@@ -0,0 +1,372 @@
...
+ *     textScale = const(1.5f),
+ *   ),
+ *   span(feature.get("name").asString().substring(const(1)))
+ * )
+ * ```
+ *
+ * Capitalizes the first letter of the features' property "name" and formats it to be extra-large,
+ * the rest of the name is written normally.
+ */
+public fun format(vararg spans: FormatSpan): Expression<FormattedValue> =
```
westnordost: I'd expect the `format` stuff to be either in its own file or in the file with the other string-related functions rather than in the file for "casting".

Same with `image`.

I mean, they are not really related, right?

lib/maplibre-compose-expressions/src/commonMain/kotlin/dev/sargunv/maplibrecompose/expression/value/CirclePitchAlignment.kt
```diff
@@ -0,0 +1,13 @@
+package dev.sargunv.maplibrecompose.expression.value
+
+import dev.sargunv.maplibrecompose.expression.ast.StringLiteral
+
+/** Orientation of circles when the map is pitched. */
+public enum class CirclePitchAlignment(override val literal: StringLiteral) :
```
westnordost: You've put the expression stuff into an own dependency now. But `CirclePitchAlignment` and all the others are not really directly related to expressions but rather to layer properties, aren't they?

westnordost (commented):
I like that a lot of things that were before in the huge ExpressionScope.kt are now in each their own files. You even split up the functions in the ExpressionScope. Looks very clean.

Otherwise, I can't say much about the code, as I don't understand the reason for this refactor.

All the stuff in expression/ast that is public is not relevant for the average user of maplibre-compose. Since maplibre-compose/expression/ast is a transitive dependency, AFAIK it should be possible to make these not visible to users of maplibre-compose as long as they didn't explicitly add the dependency in their build.gradle, AFAIK? (In case you don't do that yet)

sargunv:
> AFAIK it should be possible to make these not visible to users of maplibre-compose as long as they didn't explicitly add the dependency in their build.gradle, AFAIK?

We can hide the whole module with `implementation` instead of `api` but that'd hide the DSL too. I don't think it's possible to hide just part of the module from IDE typeahead (other than hiding it from docs entirely)

lib/maplibre-compose-expressions/src/commonMain/kotlin/dev/sargunv/maplibrecompose/expression/value/CirclePitchAlignment.kt
sargunv: they're implementations of a `sealed interface EnumValue` so they need to be in the same module. And I see them as specializations of String, similar to how DpValue, TextUnitValue, etc are specializations of NumberValue.

lib/maplibre-compose-expressions/src/commonMain/kotlin/dev/sargunv/maplibrecompose/expression/dsl/types.kt
sargunv: makes sense. I mostly just used the existing `// region` declarations to split these files

lib/maplibre-compose-expressions/src/commonMain/kotlin/dev/sargunv/maplibrecompose/expression/dsl/inputs.kt
sargunv: Yup because they're in the same package

PR #196, opened 2024-12-26
feat: Add bare Kotlin type overloads for expression functions

### LINKED ISSUE #193: Add bare Kotlin type overloads for expression functions

sargunv:
closes #193

### DIFF
--- a/lib/maplibre-compose-expressions/src/commonMain/kotlin/dev/sargunv/maplibrecompose/expressions/dsl/collections.kt
+++ b/lib/maplibre-compose-expressions/src/commonMain/kotlin/dev/sargunv/maplibrecompose/expressions/dsl/collections.kt
@@ -16,6 +16,11 @@ public operator fun <T : ExpressionValue> Expression<ListValue<T>>.get(
   index: Expression<IntValue>
 ): Expression<T> = FunctionCall.of("at", index, this).cast()
 
+/** Returns the item at [index]. */
+@JvmName("getAt")
+public operator fun <T : ExpressionValue> Expression<ListValue<T>>.get(index: Int): Expression<T> =
+  get(const(index))
+
 /** Returns whether this list contains the [item]. */
 @JvmName("containsList")
 public fun <T : ExpressionValue> Expression<ListValue<T>>.contains(
@@ -40,6 +45,16 @@ public fun <T : ExpressionValue> Expression<ListValue<T>>.indexOf(
 }
 
 /**
+ * Returns the first index at which the [item] is located in this list, or `-1` if it cannot be
+ * found. Accepts an optional [startIndex] from where to begin the search.
+ */
+@JvmName("indexOfList")
+public fun <T : ExpressionValue> Expression<ListValue<T>>.indexOf(
+  item: Expression<T>,
+  startIndex: Int? = null,
+): Expression<IntValue> = indexOf(item, startIndex?.let { const(it) })
+
+/**
  * Returns the items in this list from the [startIndex] (inclusive) to the end of this list if
  * [endIndex] is not specified or `null`, otherwise to [endIndex] (exclusive).
  */
@@ -55,6 +70,15 @@ public fun <T : ExpressionValue> Expression<ListValue<T>>.slice(
   return FunctionCall.of("slice", *args.toTypedArray()).cast()
 }
 
+/**
+ * Returns the items in this list from the [startIndex] (inclusive) to the end of this list if
+ * [endIndex] is not specified or `null`, otherwise to [endIndex] (exclusive).
+ */
+public fun <T : ExpressionValue> Expression<ListValue<T>>.slice(
+  startIndex: Int,
+  endIndex: Int? = null,
+): Expression<ListValue<T>> = slice(const(startIndex), endIndex?.let { const(it) })
+
 /** Gets the length of a this list. */
 @JvmName("lengthOfList")
 public fun Expression<ListValue<*>>.length(): Expression<IntValue> =
@@ -65,6 +89,13 @@ public operator fun <T : ExpressionValue> Expression<MapValue<T>>.get(
   key: Expression<StringValue>
 ): Expression<T> = FunctionCall.of("get", key, this).cast()
 
+/** Returns the value corresponding the given [key] or `null` if it is not present in this map. */
+public operator fun <T : ExpressionValue> Expression<MapValue<T>>.get(key: String): Expression<T> =
+  get(const(key))
+
 /** Returns whether the given [key] is in this map. */
 public fun Expression<MapValue<*>>.has(key: Expression<StringValue>): Expression<BooleanValue> =
   FunctionCall.of("has", key, this).cast()
+
+/** Returns whether the given [key] is in this map. */
+public fun Expression<MapValue<*>>.has(key: String): Expression<BooleanValue> = has(const(key))
--- a/lib/maplibre-compose-expressions/src/commonMain/kotlin/dev/sargunv/maplibrecompose/expressions/dsl/decision.kt
+++ b/lib/maplibre-compose-expressions/src/commonMain/kotlin/dev/sargunv/maplibrecompose/expressions/dsl/decision.kt
@@ -18,20 +18,20 @@ import kotlin.jvm.JvmName
  * to `true`, or the [fallback] value otherwise.
  *
  * Example:
- * ```
+ * ```kt
  * switch(
  *   condition(
- *     test = feature.has(const("color1")) and feature.has(const("color2")),
+ *     test = feature.has("color1") and feature.has("color2"),
  *     output = interpolate(
  *       linear(),
  *       zoom(),
- *       1 to feature.get(const("color1")).convertToColor(),
- *       20 to feature.get(const("color2")).convertToColor()
+ *       1 to feature.get("color1").convertToColor(),
+ *       20 to feature.get("color2").convertToColor()
  *     ),
  *   ),
  *   condition(
- *     test = feature.has(const("color")),
- *     output = feature.get(const("color")).convertToColor(),
+ *     test = feature.has("color"),
+ *     output = feature.get("color").convertToColor(),
  *   ),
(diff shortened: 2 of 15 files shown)


sargunv:
* **#196** <a href="https://app.graphite.dev/github/pr/sargunv/maplibre-compose/196?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/sargunv/maplibre-compose/196?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

lib/maplibre-compose-expressions/src/commonMain/kotlin/dev/sargunv/maplibrecompose/expressions/dsl/math.kt
```diff
@@ -9,13 +9,13 @@ import dev.sargunv.maplibrecompose.expressions.value.NumberValue
...
-public fun ln2(): Expression<FloatValue> = FunctionCall.of("ln2").cast()
+public val LN_2: Expression<FloatValue> = FunctionCall.of("ln2").cast()
 
 /** Returns the mathematical constant π */
-public fun pi(): Expression<FloatValue> = FunctionCall.of("pi").cast()
+public val PI: Expression<FloatValue> = FunctionCall.of("pi").cast()
 
 /** Returns the mathematical constant e */
-public fun e(): Expression<FloatValue> = FunctionCall.of("e").cast()
+public val E: Expression<FloatValue> = FunctionCall.of("e").cast()
```
sargunv, line 18: tweaking these to match conventions in `kotlin.math`

lib/maplibre-compose-expressions/src/commonMain/kotlin/dev/sargunv/maplibrecompose/expressions/dsl/variables.kt
```diff
@@ -13,15 +13,12 @@ import kotlin.jvm.JvmInline
  *   x.use() + const(3)
  * }
  * ```
- *
- * Variable names starting with `__` are reserved for internal use by the library.
```
sargunv, line 17: this is no longer needed with the new AST

lib/maplibre-compose-expressions/src/commonMain/kotlin/dev/sargunv/maplibrecompose/expressions/dsl/math.kt
westnordost, line 1: all overloads with bare types are missing for the math functions

lib/maplibre-compose-expressions/src/commonMain/kotlin/dev/sargunv/maplibrecompose/expressions/dsl/ramps.kt
westnordost, line 1: The stops for the interpolate expressions and step expressions don't have bare Kotlin overloads.

lib/maplibre-compose-expressions/src/commonMain/kotlin/dev/sargunv/maplibrecompose/expressions/dsl/feature.kt
```diff
@@ -43,10 +52,27 @@ public object Feature {
    * or any primitive data type. Note that [state] can only be used with layer properties that
    * support data-driven styling.
    */
-  // TODO: document which layer properties support feature state expressions on which platforms
```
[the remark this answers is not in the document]
sargunv: I replaced the TODO with an issue instead: #195

lib/maplibre-compose-expressions/src/commonMain/kotlin/dev/sargunv/maplibrecompose/expressions/dsl/decision.kt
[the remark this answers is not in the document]
sargunv: It's just stylistic preference. I prefer to think of the expressions DSL as regular functions, and short-lived objects like `Case` and `Condition` are implementation details to get the typing and syntax right.

Additionally, Dokka shows documentation for functions on a different page from types, and this keeps documentation for the construct and also the branches on the same page.

lib/maplibre-compose-expressions/src/commonMain/kotlin/dev/sargunv/maplibrecompose/expressions/dsl/ramps.kt
sargunv: With current typing (the `Pair`), either all stops will have to be Kotlin values, or all stops Expressions. Thoughts on doing something like `case` and `condition` (maybe `stop`?) so we can use different overloads on each stop?

lib/maplibre-compose-expressions/src/commonMain/kotlin/dev/sargunv/maplibrecompose/expressions/dsl/math.kt
sargunv: For math operators, it felt odd to me to have different types on two sides of the operator. Maybe it's okay though

sargunv: also for anything related to numbers with units, we'd write a lot of overloads for bare types (no unit, Dp, TextUnit, Duration, maybe Meters in the future). Seems not worth it?

lib/maplibre-compose-expressions/src/commonMain/kotlin/dev/sargunv/maplibrecompose/expressions/dsl/collections.kt
```diff
@@ -16,6 +16,11 @@ public operator fun <T : ExpressionValue> Expression<ListValue<T>>.get(
...
 ): Expression<T> = FunctionCall.of("at", index, this).cast()
 
+/** Returns the item at [index]. */
+@JvmName("getAt")
+public operator fun <T : ExpressionValue> Expression<ListValue<T>>.get(index: Int): Expression<T> =
+  get(const(index))
+
 /** Returns whether this list contains the [item]. */
 @JvmName("containsList")
 public fun <T : ExpressionValue> Expression<ListValue<T>>.contains(
```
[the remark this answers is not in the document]
sargunv: what would be the most useful overloads here? I could see overloads on the left for List<String>, List<Boolean>, List<Number> and then on the right for String, Boolean, Number, so six in total?

lib/maplibre-compose-expressions/src/commonMain/kotlin/dev/sargunv/maplibrecompose/expressions/dsl/decision.kt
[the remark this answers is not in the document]
sargunv: and similar to math, overloads for generic stops means writing a whole lot of overloads for many different types

lib/maplibre-compose-expressions/src/commonMain/kotlin/dev/sargunv/maplibrecompose/expressions/dsl/collections.kt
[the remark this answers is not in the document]
westnordost: most expressions that use `contains` check if the value of some feature property is in a set of values, so e.g. `listOf("primary", "secondary", "tertiary").contains(feature["highway"])`

Can't think of a use case for having the right side as a literal.

lib/maplibre-compose-expressions/src/commonMain/kotlin/dev/sargunv/maplibrecompose/expressions/dsl/decision.kt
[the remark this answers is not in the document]
westnordost: true... , but using literals for these constructs is pretty much the default case for these constructs.

lib/maplibre-compose-expressions/src/commonMain/kotlin/dev/sargunv/maplibrecompose/expressions/dsl/math.kt
westnordost: Well, not sure, IIRC the Android MapLIbre SDK does that. (For Floats only, of course)

lib/maplibre-compose-expressions/src/commonMain/kotlin/dev/sargunv/maplibrecompose/expressions/dsl/ramps.kt
westnordost: I think it is rather the default / usual case that no expressions are used, so the overload that doesn't allow any expressions for the stops would I think be the most used one.
Having data classes for `step` and `interpolate` i.e. `stop(...)` sounds fine to me, too. It's only marginally less convenient/more verbose than the `to` (and anyway, that is just a normal infix function, you could name the function that creates a `Stop` also like that.)

sargunv: I'd be wary of shadowing `to` so would want a different infix name, or just `stop()` like the others. 

It's true that most stops/case/condition would be `const`, but also interpolate/switch is already a pretty big, verbose, construct, and so adding one const per stop doesn't hurt readability much (at least to my eye).

Compared to `get` and `has` which were small enough that the const dwarfed the get/has and made it confusing to read.

lib/maplibre-compose-expressions/src/commonMain/kotlin/dev/sargunv/maplibrecompose/expressions/dsl/collections.kt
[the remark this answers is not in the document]
sargunv: Adding an overload for List<String> sounds good to me if it doesn't shadow the regular Kotlin `contains`. 

If it does shadow, well I think we already have `const(listOf())` so it's not too bad. Should there be a `constList(varargs)`?

westnordost:
it wouldn't really be shorter, and would be less transparemt

PR #197, opened 2024-12-27
feat: add support for image sdf and painter size

### LINKED ISSUE #188: Support adding SDF images to the style
Mapbox docs explain it well, and should be applicable to MapLibre: https://docs.mapbox.com/help/troubleshooting/using-recolorable-images-in-mapbox-maps/

### LINKED ISSUE #192: Support customizing the size of Painter images
The library internally renders Painter passed into `image` to a bitmap of the intrinsic size of the painter. If the size is unspecified, we raise an exception. We should provide an optional size parameter to customize the size instead. 

With the new AST, we should be able to attach additional metadata to expression nodes. With that, we can support an optional parameter for size and use it to produce the bitmap. The reference counting cache will have to be keyed on the node instead of the painter object itself (to include size metadata in the key)

sargunv:
closes #188
closes #192

### DIFF
--- a/lib/maplibre-compose-expressions/src/commonMain/kotlin/dev/sargunv/maplibrecompose/expressions/ExpressionContext.kt
+++ b/lib/maplibre-compose-expressions/src/commonMain/kotlin/dev/sargunv/maplibrecompose/expressions/ExpressionContext.kt
@@ -1,9 +1,9 @@
 package dev.sargunv.maplibrecompose.expressions
 
-import androidx.compose.ui.graphics.ImageBitmap
-import androidx.compose.ui.graphics.painter.Painter
+import dev.sargunv.maplibrecompose.expressions.ast.BitmapLiteral
 import dev.sargunv.maplibrecompose.expressions.ast.CompiledExpression
 import dev.sargunv.maplibrecompose.expressions.ast.Expression
+import dev.sargunv.maplibrecompose.expressions.ast.PainterLiteral
 import dev.sargunv.maplibrecompose.expressions.value.FloatValue
 
 /**
@@ -21,10 +21,10 @@ public interface ExpressionContext {
   public val spScale: Expression<FloatValue>
 
   /** @return the resolved identifier for the [bitmap]. */
-  public fun resolveBitmap(bitmap: ImageBitmap): String
+  public fun resolveBitmap(bitmap: BitmapLiteral): String
 
   /** @return the resolved identifier for the [painter]. */
-  public fun resolvePainter(painter: Painter): String
+  public fun resolvePainter(painter: PainterLiteral): String
 
   /** A context where no complex types can be resolved. */
   public object None : ExpressionContext {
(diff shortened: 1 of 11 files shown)


sargunv:
* **#197** <a href="https://app.graphite.dev/github/pr/sargunv/maplibre-compose/197?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/sargunv/maplibre-compose/197?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

westnordost:
Ah, interesting, hmm, actually, there is something maplibre-compose that would somewhat cool and be again quite maybe out of scope, just as the thing about meters:

So, the deal with SDF images in MapLibre, particularly the `sdf` parameter passed to that function is that it just tells the SDK to *assume* that the passed image is already in SDF format. What is SDF format? Basically, it means that the monochrome raster image is blurred by a very specific amount of radius.

For loading the iD preset icons that are SVGs as SDF in StreetComplete, I do the "SDF conversion" myself. First, rasterize, then, blur in a very specific way. But this could be something that the framework could do itself, i.e. the `sdf` parameter could indicate that the passed bitmap should be made into a SDF. The code is here:

https://github.com/streetcomplete/StreetComplete/blob/993018694df25ca2241f7abfbb7fc8210599f870/app/src/main/java/de/westnordost/streetcomplete/util/ktx/Bitmap.kt#L13

and here https://github.com/streetcomplete/StreetComplete/blob/993018694df25ca2241f7abfbb7fc8210599f870/app/src/main/java/de/westnordost/streetcomplete/util/sdf/Sdf.kt#L12

I could declare to license that code under the BSD 3-clause license if that is something you would want to include.

Thing is, if you add SDF pictures *from code* and not from the JSON style, it is very likely that you don't have already blurred images at hand, but have normal monochrome vector icons at hand. This initially somewhat confused me, as the SDF stuff was seemingly not working correctly, until I found out that MapLibre assumes that the passed images should already be blurred.

If you don't think this automatic conversion should be included, then maybe it might be worth a thought to change the API in the way that makes it clear. E.g. have a different overload for adding SDF images?

sargunv:
Hmm, I think if bitmaps are passed the flag should just indicate "is this bitmap already an SDF". 

But if a painter is passed, we handle bitmap generation. So we can have a flag for "should this bitmap be generated as an SDF" and do that blurring. I think the naming should be slightly different for the two flags, to indicate they're different.

westnordost:
or, add a function alike 'addMonochromeIcon' or similar rather than a scond parameter (just brainstorming)

westnordost:
or, dont change the interface but supply a helper function which will do the conversion that could be put inbetween (referring to the resource and having the imagebitmap)

sargunv:
Filed https://github.com/sargunv/maplibre-compose/issues/202

westnordost:
I hereby release the two files mentioned in https://github.com/sargunv/maplibre-compose/pull/197#issuecomment-2563675917 under the BSD-3-Clause license. Another contributor contributed to that file, however, he just ran an automated lint check on it for code formatting.

PR #232, opened 2025-01-06
Reimplement Scalebar

westnordost:
https://github.com/user-attachments/assets/538ec998-5737-43f5-851a-68c64362ab94

Reimplementation of scalebar. Somewhat similar to the Google Maps one, only better :-)

- can be aligned horizontally to start, end, center or anything in-between (i.e. also layout direction aware)
- text size and style can be specified, defaults to `labelSmall`
- both scale bar and text have a consistent halo and color (defaults to material3 `surface` and `onSurface`)
- can show metric, imperial or both. Will select a reasonable default based on the user's locale or system setting (if applicable)
- automatically a (minimum) size is selected so that the text does not overlap with the scalebar. Setting a bigger size is fine, too


lib/maplibre-compose-material3/src/iosMain/kotlin/dev/sargunv/maplibrecompose/material3/util.ios.kt
```diff
@@ -0,0 +1,24 @@
...
+internal actual fun scaleBareMeasurePreference(): ScaleBarMeasure? =
+  null
+  /* TODO iOS/Mac developer:
+  when (NSLocale.currentLocale.measurementSystem) {
+    Locale.MeasurementSystem.metric -> ScaleBarMeasure.METRIC
+    Locale.MeasurementSystem.uk -> ScaleBarMeasure.IMPERIAL_AND_METRIC
+    Locale.MeasurementSystem.us -> ScaleBarMeasure.IMPERIAL
+    else -> ScaleBarMeasure.IMPERIAL_AND_METRIC
+  }
+  */
```
westnordost, line 24: @sargunv `measurementSystem` is only available from iOS 16, not sure how to only access it when on iOS 16 and above. Also, not sure about the syntax, as I don't get any help from the IDE on a non-Mac-system.

lib/maplibre-compose-material3/src/iosMain/kotlin/dev/sargunv/maplibrecompose/material3/util.ios.kt
```diff
@@ -0,0 +1,24 @@
...
+
+internal actual fun scaleBareMeasurePreference(): ScaleBarMeasure? =
+  null
+  /* TODO iOS/Mac developer:
+  when (NSLocale.currentLocale.measurementSystem) {
+    Locale.MeasurementSystem.metric -> ScaleBarMeasure.Metric
+    Locale.MeasurementSystem.uk -> ScaleBarMeasure.ImperialAndMetric
+    Locale.MeasurementSystem.us -> ScaleBarMeasure.Imperial
+    else -> ScaleBarMeasure.ImperialAndMetric
+  }
```
westnordost, line 23: @sargunv `measurementSystem` is only available from iOS 16, not sure how to only access it when on iOS 16 and above. Also, not sure about the syntax, as I don't get any help from the IDE on a non-Mac-system.

lib/maplibre-compose-material3/src/commonMain/kotlin/dev/sargunv/maplibrecompose/material3/controls/ScaleBar.kt
```diff
@@ -1,253 +1,300 @@
...
-      shadowColor = MaterialTheme.colorScheme.surface,
-    )
+/** Which measure to show on the scale bar. */
+public enum class ScaleBarMeasure {
+  /** meters / kilometers */
+  Metric,
+  /** feet / miles */
+  Imperial,
+  /** both feet / miles and meters / kilometers */
+  ImperialAndMetric;
```
sargunv, line 37: Haven't reviewed the whole thing yet but just a naming nit: the "imperial" measuring system means specifically the UK system while the US system is "US customary". Both evolved from an older English system, and today differ in some units (though feet and miles are the same for both, ignoring the "survey foot" and "survey mile" and nonsense like that).

I'd name this `Meters`, `Miles`, and `MetersAndMiles` or something like that, and update comments referencing "imperial" to reference "feet / miles".

lib/maplibre-compose-material3/src/commonMain/kotlin/dev/sargunv/maplibrecompose/material3/util.kt
```diff
@@ -0,0 +1,82 @@
...
+    ?: defaultScaleBarMeasure(Locale.current.language, Locale.current.region)
+
+/** default scale bar measure to use, depending on the locale */
+internal fun defaultScaleBarMeasure(language: String?, country: String?): ScaleBarMeasure {
+  if (!country.isNullOrEmpty()) {
+    return when (country) {
+      // US is famously imperial only
+      "US" -> {
+        ScaleBarMeasure.Imperial
+      }
```
sargunv, line 24: I'd default to both in the US as well

lib/maplibre-compose-material3/src/androidMain/kotlin/dev/sargunv/maplibrecompose/material3/util.android.kt
```diff
@@ -0,0 +1,5 @@
+package dev.sargunv.maplibrecompose.material3
+
+import dev.sargunv.maplibrecompose.material3.controls.ScaleBarMeasure
+
+internal actual fun scaleBareMeasurePreference(): ScaleBarMeasure? = null
```
sargunv, line 5: API 28" has `LocaleData.getMeasurementSystem`: https://developer.android.com/reference/android/icu/util/LocaleData#getMeasurementSystem(android.icu.util.ULocale)

not quite a "preference" since I think the user can't set it, but still probably good to rely on that to inform the default on APIs where it's available.

And then the big hardcoded list can be commented with a note for removal when minSdk >= 28

lib/maplibre-compose-material3/src/commonMain/composeResources/values-es/strings.xml
```diff
@@ -1,3 +0,0 @@
-<resources>
-  <string name="feet_symbol">pies</string>
-</resources>
```
sargunv, line 3: Was this removed due to the length of the symbol?

lib/maplibre-compose-material3/src/desktopMain/kotlin/dev/sargunv/maplibrecompose/material3/util.desktop.kt
```diff
@@ -0,0 +1,5 @@
+package dev.sargunv.maplibrecompose.material3
+
+import dev.sargunv.maplibrecompose.material3.controls.ScaleBarMeasure
+
+internal actual fun scaleBareMeasurePreference(): ScaleBarMeasure? = null
```
sargunv, line 5: should put a TODO comment here; macOS at least probably has an API for this identical to the iOS one. unsure about other platforms.

lib/maplibre-compose-material3/src/commonMain/kotlin/dev/sargunv/maplibrecompose/material3/controls/ScaleBar.kt
westnordost: Meters and Miles sounds off, because one mile is roughly 1500x of a meter.

I've looked over the wikipedia articles for US customary units, imperial units etc.

The gist of it is that regarding (feet + mile) length, the Imperial and US customary measurement system are identical through the [international yard and pound](https://en.wikipedia.org/wiki/International_yard_and_pound) agreement of 1959. So, the full name would be "US customary and British Imperial measurement units".

Suggestions:
- leave as is
- use `Meters`, `FeetAndMiles`, `Both`
- use `SI`, `US`, `UK` (like in CLDR). However, the CLDR `MeasurementSystem` refers to the measurement system in general, including weights, masses, areas, volues, ... , and is thus much more coarse.

lib/maplibre-compose-material3/src/androidMain/kotlin/dev/sargunv/maplibrecompose/material3/util.android.kt
westnordost: I've found the data source for the CLDR data on the preferred measurement system per country. It is quite coarse, much more coarse than what I put in the code:

https://github.com/unicode-org/cldr/blob/main/common/supplemental/units.xml#L478-L496

The gist of that data: Use feet+miles in `US`, use yard+miles in `GB`. So, it's worse than the data I researched.

lib/maplibre-compose-material3/src/desktopMain/kotlin/dev/sargunv/maplibrecompose/material3/util.desktop.kt
westnordost: I also think there is an API for that on MacOS. But in the `build.gradle`, the `desktop` target is actually the `jvm` target, which is disjunct from the macOs target according to this illustration:

default-hierarchy-example

(source: https://kotlinlang.org/docs/multiplatform-hierarchy.html#default-hierarchy-template )

lib/maplibre-compose-material3/src/commonMain/kotlin/dev/sargunv/maplibrecompose/material3/controls/ScaleBar.kt
westnordost: Addendum: According to the CLDR and confirmed by viewing some StreetView imagery, in the United Kingdom, they don't use feet+miles, but yards+miles ... 🤔

lib/maplibre-compose-material3/src/androidMain/kotlin/dev/sargunv/maplibrecompose/material3/util.android.kt
westnordost: > And then the big hardcoded list

(By the way, it is only big because this "spotless check format" insists on a linebreak after each country and language code :-/)

lib/maplibre-compose-material3/src/commonMain/kotlin/dev/sargunv/maplibrecompose/material3/controls/ScaleBar.kt
sargunv: Perhaps the enum has three options: `Metric`, `FeetAndMiles`, `YardsAndMiles` (or SI, US, UK), and the scale bar takes up to two of them as parameters: `primaryMeasurementSystem` and `secondaryMeasurementSystem`.

* Default in most of the world: primary is SI, secondary is null
* Default everywhere else,
  * primary is US or UK if not transitioning, SI if transitioning
  * secondary is SI if not transitioning, US or UK if transitioning

lib/maplibre-compose-material3/src/androidMain/kotlin/dev/sargunv/maplibrecompose/material3/util.android.kt
sargunv: Hmm, with my other comment on "primary" and "secondary", maybe we use this data for the default "primary", then:
* for locales with primary = US or UK, secondary default is SI
* for locales with primary = SI, there's a hardcoded list of secondary=US or UK, rest are null

lib/maplibre-compose-material3/src/commonMain/composeResources/values-es/strings.xml
sargunv: sounds good to me. I'll follow up with some spanish speaking folks in the US to confirm but it's probably reasonable to remove.

lib/maplibre-compose-material3/src/desktopMain/kotlin/dev/sargunv/maplibrecompose/material3/util.desktop.kt
sargunv: Yeah, we'd have to detect platform and call in to native APIs probably (if there's not already a wrapper library for it). Can just leave it as a TODO for now (I'll take care of it later)

sargunv:
(I can also pick this up from here and wrap up Apple implementation and the primary/secondary thing if you'd like)

lib/maplibre-compose-material3/src/desktopMain/kotlin/dev/sargunv/maplibrecompose/material3/util.desktop.kt
sargunv: note for later, [this](https://javadoc.lwjgl.org/org/lwjgl/system/macosx/ObjCRuntime.html) is a helper to make calls to ObjC from the JVM

westnordost:
I am currently looking into the yard+mile stuff

lib/maplibre-compose-material3/src/commonMain/kotlin/dev/sargunv/maplibrecompose/material3/controls/ScaleBar.kt
```diff
@@ -1,253 +1,388 @@
...
+    50000f,
+    100000f,
+    200000f,
+    500000f,
+    1000000f,
+    2000000f,
+    5000000f,
+    10000000f,
+    20000000f,
+    40000000f,
```
sargunv, line 322: these can be generated, something like:

```kt
private fun buildStops(mantissas: Set<Int>, exponents: IntRange): List<Double> = buildList {
    for (e in exponents)
        for (m in mantissas)
            add(m.toDouble() * Math.pow(10.0, e.toDouble()))
}
```

westnordost:
hmm, maybe the Measure should be an interface... experimenting with an implementation

westnordost:
Alright, done! One thing I am a bit worried about is whether the `ScaleBarMeasure` `.Metric` etc. are still considered as `@Stable` by Compose. If they are not, then this would lead to the widget being recomposed every frame, even if nothing changed. (Because the `List` of stops **might** have changed)

See also https://developer.android.com/develop/ui/compose/performance/stability/fix#immutable-collections

Should this library made to depend on kotlinx.collections.immutable to fix this?

sargunv:
> One thing I am a bit worried about is whether the ScaleBarMeasure .Metric etc. are still considered as @Stable by Compose.

Strong skipping should prevent that, right? We can also tag them with `@Immutable`

lib/maplibre-compose-material3/src/commonMain/kotlin/dev/sargunv/maplibrecompose/material3/controls/ScaleBar.kt
westnordost: Looks quite ingenious! For me at least, this would introduce a readability problem however, as I find it quite hard to understand what would be the resulting list.

This is the sort of complexity where I'd say that this needs a test case. But then, at the end, it will probably not really be less lines of code.

But anyway, if you want to go ahead with this anyway, I don't mind much.

sargunv: > For me at least, this would introduce a readability problem however, as I find it quite hard to understand what would be the resulting list.

I considered that, but I think it's mitigated with just a comment explaining it's based on scientific notation: `m*10^e`.

> if you want to go ahead with this anyway, I don't mind much.

sure, I'll take care of it; just need to tweak the above pattern for the transition between feet/yards and miles.

westnordost:
I thought strong skipping is only for automatic lambda memoization?

The [docs make it sound like](https://developer.android.com/develop/ui/compose/performance/stability/strongskipping#when-skip) I **have** to annotate the thing as `@Stable` to enable the `.equals(...)` check

If we just annotate it as `@Immutable`, we are handing out guarantees that are out of our control. (`ScaleBarMeasure` is not sealed, because it doesn't have to be, and to enable funny 1st of April measures 🤩)

sargunv:
separately, I'm tempted to experiment with https://github.com/kevincianfarini/alchemist to represent the length units and their conversion using their `Length` value class (like `Duration`); unsure if it'll be worth the extra dependency but I'll try it and see how it looks.

sargunv:
> If we just annotate it as @Immutable, we are handing out guarantees that are out of our control.

ah yeah, I was thinking just our implementations should be `@Immutable`. But I don't know, with polymorphic parameters, does the compose runtime consider immutability of the actual value at runtime, or just the declared type?

> The [docs make it sound like](https://developer.android.com/develop/ui/compose/performance/stability/strongskipping#when-skip) I have to annotate the thing as @Stable to enable the .equals(...) check

I think `@Immutable` is also considered a "stable" value by the compose compiler, not just `@Stable`.

> I thought strong skipping is only for automatic lambda memoization?

https://developer.android.com/develop/ui/compose/performance/stability/strongskipping#when-skip

> Strong Skipping is a mode available in the Compose compiler. When enabled, it changes the compiler's behavior in two ways:
> 
> Composables with unstable parameters become [skippable](https://developer.android.com/develop/ui/compose/performance/stability#functions)
> Lambdas with unstable captures are [remembered](https://developer.android.com/develop/ui/compose/state#state-in-composables)

westnordost:
> Skippable: If the compiler marks a composable as skippable, Compose can skip it during recomposition if all its arguments are equal with their previous values.

Hmm okay, sounds like we are good(?)

sargunv:
> Hmm okay, sounds like we are good(?)

yup, `object` combined with referential equality means we should be fine even without `@Immutable`

westnordost:
https://github.com/user-attachments/assets/48d844bf-6959-4a24-a596-c310361ca71c

lib/maplibre-compose-material3/src/commonMain/kotlin/dev/sargunv/maplibrecompose/material3/controls/ScaleBarMeasure.kt
```diff
@@ -0,0 +1,163 @@
...
+        5000000f,
+        10000000f,
+        20000000f,
+        40000000f,
+      )
+
+    @Composable
+    override fun getText(stop: Float): String =
+      if (stop >= 1000) {
+        "${(stop/1000f).toShortString()} ${stringResource(Res.string.kilometers_symbol)}"
```
sargunv, line 60: I just noticed the NNBSP, nice touch

PR #234, opened 2025-01-08
exploration: try out alchemist Length type

sargunv:
I like how this enables map scale arithmetic between length and dp. But there's some blockers before we can merge it:
* https://github.com/kevincianfarini/alchemist/issues/54 (hard)
* https://github.com/kevincianfarini/alchemist/issues/53 (soft)

I also want to see whether it can make implementations of our scale bar measure interface simpler (generate stops based on provided units).


sargunv:
* **#234** <a href="https://app.graphite.dev/github/pr/sargunv/maplibre-compose/234?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/sargunv/maplibre-compose/234?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

westnordost:
I don't know, it looks like an interesting finger exercise, but I don't see the point:

- it increases code complexity, and is also more code
- it adds a dependency (every additional dependency is a liability), and this dependency is very fresh (API may change rapidly, bugs)

and most importantly, what about my 🍌 bananas unit? It is certainly not supported by Alchemist! 😥

sargunv:
> it increases code complexity, and is also more code

re: "more code", it's primarily due to the addition in the expressions AST, only necessary if we want `Length` in those fill extrusion expressions specified in meters.

re: "complexity": personally I like the simplicity and compile time safety of arithmetic like `length = scale * dp` or `dp = length / scale`.

> it adds a dependency (every additional dependency is a liability), and this dependency is very fresh (API may change rapidly, bugs)

but yeah, this is my main concern, and the primary reason why this PR is just a draft. My thinking here is that it's also true of _our_ library, and this pre v1 stage is the best time to experiment with API things like this. Worst case, we push a breaking change in the future to drop the dependency (or fork it).

> and most importantly, what about my 🍌 bananas unit? It is certainly not supported by Alchemist! 😥

I actually think, with the right API, it'll be _easier_ to add new scale bar measures (including custom units, as long as your custom unit is divisible by nanometers).

Something like, for already supported units:

```kt
data object FeetYardsMiles : ScaleBarMeasure {
  // stops generated automatically based on real size of these units
  override val units = setOf(Foot, Yard, Mile)

  // optional override to provide localized symbols
  @Composable
  override fun format(length: Length, unit: LengthUnit) = when(unit) {
    Foot -> "${length.toShortString()} ${stringResource(Res.values.foot_symbol)}"
    ..
  }
}
```

and for measures with custom units:

```kt
data object Banana : LengthUnit { 
  override val nanometerScale: Long get() = 19.centimeters.toLong(Nanometer)
  override val symbol = "🍌"
}

data object BananasMeasure : ScaleBarMeasure {
  override val units = setOf(Banana)

  // default `format` uses `LengthUnit.symbol` without localization
}
```

Bananas are silly, but realistically I could see this being used by consumers of the library to create scale bars for less common units like [nautical miles](https://en.wikipedia.org/wiki/Nautical_mile), or [US survey feet / miles](https://en.wikipedia.org/wiki/Mile#US_survey)

sargunv:
Here it is, kilobananas and football fields in the latest draft, with auto generated stops.

westnordost:
Good to hear that bananas survived. 😆
Anyway, 

> only necessary if we want Length in those fill extrusion expressions specified in meters

My argument still stands, though. Is it worth the effort (, added dependencies, added code complexity) to have a different type for the height in `FillExtrudeLayer`? For example, when specifying pixel values in Compose, we also don't have a `Px` type. It's not worth to introduce types for the sake of it. You've seen that it complicates other things, such as having to provide (operator, etc.) overloads for each of those types - or deciding not to do it because it is too much effort, which then in any case makes it less comfortable to use for users than if it wasn't an own type to begin with.
The reason why there is a `Dp` type in Compose is to clearly distinguish it from pixels and make clear that it must be divided by the screen pixel density for drawing.

sargunv:
I think it's not worth the effort just for that single property. I think it is worth the effort for the scale bar API and implementation. 

So the question from there is, assuming we keep the dependency for the scale bar, is it worth also using the dependency for the map scale type and the fill extrusion property. 

If both are no, we _could_ keep this as a dependency of the material3 extensions module and kick it out of the core module (but, would it return to core if some of this scale bar logic needs to be reused in a different UI module?)

If map scale is yes, imo the fill extrusion height should use it just for consistency. 

I'm not convinced map scale should use it though; I'm leaning keep it in the material3 extensions module for now and revert the changes in core.

sargunv:
Fictional maps (see local tiles demo) are an interesting use case for custom units in the scale bar. I think for v0.8.0 it'd be good to pull this in (just as a material3 module dependency, not as a core dependency)

sargunv:
closing this out in favor of #350 (just the material3 portion of this exploration)

PR #235, opened 2025-01-08
scale bar: slight performance improvement

westnordost:
after all, the `defaultScaleBarMeasures` is now called every frame.

(If it wasn't `@Composable`, it could also be `remember`ed)

---

I was previously looking into performance improvements on the scalebar, after all, quite a lot is going on there. But basically any attempt to let the scalebar compose less often is sabotaged by the fact that any change in latitude or zoom will make it recompose anyway, as the `metersPerDp` changes.

`metersPerDp` could maybe be rounded, the trick would be to find out at which precision it starts to matter (i.e. be visible on the scalebar). Then, it would generally not recompose when scrolling around in an area, only when one zooms.


sargunv:
Haven't looked at the PR yet, but:

> metersPerDp could maybe be rounded, the trick would be to find out at which precision it starts to matter (i.e. be visible on the scalebar). Then, it would generally not recompose when scrolling around in an area, only when one zooms.

`derivedStateOf` could possibly be useful here? input is metersPerDp, outputs are bar width and text. Bar width (dp) could be rounded based on density.

PR #243, opened 2025-01-12
wip: add js bindings for sources and layers

sargunv:


Namnodorel:
I would be willing to help out with this as I need these bindings for a project. Is it correct that this basically just requires the busywork of implementing all the `actual` functions in `jsMain`?

Issue #164, opened 2024-12-21
refactor attribution info as actual State in StyleState


sargunv:
Attribution info changes as sources load or are removed from the map. It should be State to make the material3 attribution dialog implementation cleaner and more robust.

We can also use this to implement a Material3 "attribution bar" similar to MapLibre Web's default attribution UI

westnordost:
(This is probably impossible to include in 0.7.0 unless that version will be the giga-update that replaces MapLibre iOS and MapLibre Android with direct use of MapLibre native, see the linked PR.)

sargunv:
Thanks for looking in to it! makes sense to me.

Issue #77, opened 2024-12-02
The camera follow demo does not stop following on double tap zoom


sargunv:
Normally the demo stops following when any gesture happens, but the double tap zoom gesture doesn't trigger this. Affects both Android and iOS.

sargunv:
compass part fixed in https://github.com/sargunv/maplibre-compose/pull/179

double tap is still outstanding

michalgwo:
I noticed that when the camera follow animation is run, the move reason is programmatic, but when it's not, the move reason is a gesture. It may be because of the execution of `cameraState.animateTo()` every few hundred milliseconds, which overrides the double tap gesture (see: https://github.com/maplibre/maplibre-native/issues/2790). I limited `cameraState.animateTo()` executions to 1 per second and the move reason was correctly detected as a gesture. 

I'm not sure if it's important, but I also noticed that when the camera follow animation is run and I tap on the map without moving it, it detects it as a programmatic camera movement (double tap, detects 2 programmatic camera movements), but when I tap on the map when there is no animation, it doesn't detect any camera movement.

As for differentiation between the developer's and API's programmatic reasons for camera movement, it seems to be a bug in Maplibre, and it was an issue before the fork on Mapbox (see: https://github.com/mapbox/mapbox-gl-native/issues/10674#issuecomment-526181969)

As a workaround for the demo, we can stop the animation on zoom changes, but it's not a real solution, so probably not a good option.

sargunv:
Yup. We're limited by what the SDK provide here. Calling animateTo so frequently doesn't even work properly on JS.

I'm hoping to solve this by moving the gestures implementation fully into this library. We'll need to reimplement gestures anyway for desktop native, so if we make that common code I hope we can solve this and also get better integration with Compose gesture interop APIs.

sargunv:
#870 eliminates the platform compass so this is now solvable on our end. probably an api that allows a "programmatic" camera movement to register as gesture? won't fully work for android/ios until #572 but solving it for desktop/web will solve it for those too eventually.

sargunv:
thinking some more, breaking follow on compass tap is somewhat weird, and the user can already wire that up if they want. double tap zoom is handled by our compose-implemented gestures now on desktop/web, so what's left here is just duplicate of #572

PR #248, opened 2025-01-13
attribution button with bar

### LINKED ISSUE #174: Change Material3 attribution button to an attribution bar visible by default
See discussion in https://github.com/maplibre/maplibre-native/issues/2600

probably depends on #164

westnordost:
Fixes #174

https://github.com/user-attachments/assets/ba1407b3-8951-461d-86e2-6faaf145395e

I deviated from my suggestion in #174 because:
- **No dialog:** there's not really the need for a dialog, if the attribution text(s) are too long, they can just wrap to several lines
- **No Text buttons but links in text**: fully native and JS implementation is incoming. This means that parsing the attribution text+link will be redone and probably parsed directly to `AnnotatedString`s 
imagen
  which will enable to (correctly) have only parts of the attribution text link to the attribution source, as it should be and as it is done in MapLibre JS.
- **no button background**: personal taste, I guess. IMO the info button should not look like a regular (but smaller?) map button like the compass, etc. because it's rather related to the map than to the HUD-like UI on top of the map (just like the scalebar). It would IMO look weird to have a smaller mis-aligned button amongst other map buttons. Also, if the attribution text is shown fully before map move, it's quite fine to make this button somewhat unobtrusive. Anyway, I think this can be customized anyway with `IconButtonColors`? (didn't try)

I briefly tried out to give the info button 1px white halo, like the scalebar has, but in my eyes, this looked a bit disruptive, even though one would think it would look consistent with the scalebar.

### Question:
- should the text links by default be styled? I am surprised there is no default styling for text links in Compose. E.g. colored like text buttons, with underscore-effect on hover.

### TODO:
- The attribution _should_ automatically retract when the map is being interacted with (i.e. moved) by the user. But I think there is no API for that yet, right?

### DIFF
--- a/demo-app/src/commonMain/kotlin/dev/sargunv/maplibrecompose/demoapp/app.kt
+++ b/demo-app/src/commonMain/kotlin/dev/sargunv/maplibrecompose/demoapp/app.kt
@@ -166,7 +166,12 @@ fun DemoMapControls(
         modifier = Modifier.align(Alignment.TopEnd),
         onClick = onCompassClick,
       )
-      AttributionButton(styleState, modifier = Modifier.align(Alignment.BottomEnd))
+      val attributionLinks = remember { styleState.queryAttributionLinks() }
+      AttributionButton(
+        lastCameraMoveReason = cameraState.moveReason,
+        attributions = attributionLinks,
+        modifier = Modifier.align(Alignment.BottomEnd)
+      )
     }
   }
 }
--- a/demo-app/src/commonMain/kotlin/dev/sargunv/maplibrecompose/demoapp/docs/Material3.kt
+++ b/demo-app/src/commonMain/kotlin/dev/sargunv/maplibrecompose/demoapp/docs/Material3.kt
@@ -36,12 +36,9 @@ fun Material3() {
     Box(modifier = Modifier.fillMaxSize().padding(8.dp)) {
       ScaleBar(
         metersPerDp = cameraState.metersPerDpAtTarget,
-        modifier = Modifier.align(Alignment.TopStart)
+        modifier = Modifier.align(Alignment.TopStart),
       )
-      CompassButton(
-        cameraState = cameraState,
-        modifier = Modifier.align(Alignment.TopEnd)
-      )
+      CompassButton(cameraState = cameraState, modifier = Modifier.align(Alignment.TopEnd))
       val attributionLinks = remember { styleState.queryAttributionLinks() }
       AttributionButton(
         lastCameraMoveReason = cameraState.moveReason,
--- a/demo-app/src/commonMain/kotlin/dev/sargunv/maplibrecompose/demoapp/docs/Material3.kt
+++ b/demo-app/src/commonMain/kotlin/dev/sargunv/maplibrecompose/demoapp/docs/Material3.kt
@@ -6,6 +6,7 @@ import androidx.compose.foundation.layout.Box
 import androidx.compose.foundation.layout.fillMaxSize
 import androidx.compose.foundation.layout.padding
 import androidx.compose.runtime.Composable
+import androidx.compose.runtime.remember
 import androidx.compose.ui.Alignment
 import androidx.compose.ui.Modifier
 import androidx.compose.ui.unit.dp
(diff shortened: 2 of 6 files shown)


sargunv:
> The attribution should automatically retract when the map is being interacted with (i.e. moved) by the user. But I think there is no API for that yet, right?

There is, it'd be the same technique used to stop following on user gesture in the camera follow demo, though it's not reliable yet on the JS side. It has caveats that should be fixed in the future (see #77)

> should the text links by default be styled? I am surprised there is no default styling for text links in Compose. E.g. colored like text buttons, with underscore-effect on hover.

Maybe an underline by default? No strong opinion here, though it'd be nice if the links were visibly clickable

> there's not really the need for a dialog, if the attribution text(s) are too long, they can just wrap to several lines

if such wrapping is unlikely with common map styles and common displays, sounds good to me

> no button background

I agree having an unobtrusive button is best, but perhaps we can still include the background when the text is expanded? So it starts with the attribution text and button, then on gesture the text collapses into the button and the background collapses + fades.

westnordost:
I'll look into it

westnordost:
Reproducible on an Android 15 emulator in landscape mode.

westnordost:
Implemented both suggestions. The issue with `styleState.queryAttributionLinks()` is that it only queries the current situation. If it changes, the widget is not updated. As a matter of fact, the attribution info for all the styles seems to be empty, at least at the time in which the attribution button is (first) composed. Probably at that time the style didn't fully load yet.
So, it would be better if `styleState.attribution` was actually a field.

https://github.com/user-attachments/assets/29183659-531d-4d04-8a08-29e6a954e7d0

> Maybe an underline by default? No strong opinion here, though it'd be nice if the links were visibly clickable

Tried it out. While still for each attribution the whole link is clickable, this looks odd. Also, I've looked into M3 guidelines, and they don't loose a word about in-text links (and TextButtons have a different appearance).
I think if the link is colored in the `primary` color (like text links) and underlined on focus, hover and press, that would be a good default in the future, but right now it would look odd.

sargunv:
> The issue with `styleState.queryAttributionLinks()` is that it only queries the current situation. If it changes, the widget is not updated.

Yup, it would be good to have it work through State, similar to camera position and camera move reason. That's https://github.com/sargunv/maplibre-compose/issues/164

westnordost:
So merging this is actually blocked by #164, because when a user just naively employed this, no attribution would be shown.

sargunv:
Yeah I think so. 

Still, thanks for doing this! Will review in a bit

westnordost:
Regarding #250: Maybe it would make sense to allow the button to be align towards any corner. However, I didn't do this because I have not idea how to do it. When you review the code, you'll see that several things are aligned to the bottom right within the widget:
- the info button, so that it always stays in its corner, regardless of the height of the attribution info texts (which may grow to several lines)
- the texts within the attribution info field are aligned to the end
- the containing box aligns its content to the bottom right, so that the AnimatedVisibility animation also expands/retracts the content from/to there, respectively

It **could** be parametrized, but the issue is that it is a wild mix of `Arrangement`, `Alignment` and `Alignment.Horizontal`, so passing e.g. an `Alignment` will not work as it is not possible to inspect the fields of an `Alignment` (the `bias` is not part of the interface).

westnordost:
I'll try out an idea how to implement this.

sargunv:
> it is not possible to inspect the fields of an Alignment

I've done this elsewhere in the library by just asking the Alignment instance to align a 1x1 box in some convenient rect. For an example, check how I map Alignment to Android gravity or the iOS bit mask for the SDK ornaments.

sargunv:
I'm not sure if the material 3 library has tooltip or popup menu components, but if it does we could probably look at how they work too

westnordost:
**To summarize**:
- The PR is ready to merge / ready to review
- but is effectively blocked by #164

westnordost:
Hm, I am a bit discontent with this. It is weird to have a component in the map HUD layout that can expand its size to any size.

I'll look into whether/how to display the expanded attribution info in a view that overlaps all (popup / dialog), hopefully possible with a nice animation like it is now.

westnordost:
Success! Superseded by #293.

However, I wasn't able to re-create this nice appear-animation with the expanding popup bar, so I used a simple fade in / fade out.

PR #249, opened 2025-01-15
don't show scalebar when the map is not initialized yet

westnordost:


sargunv:
Leaving myself a note, if we end up with a dedicated type for Map scale as in https://github.com/sargunv/maplibre-compose/pull/234, this 0.0 should become something like MapScale.Unspecified (similar to the Size type)

Issue #250, opened 2025-01-15
Add possibility to set compass margins and image


michalgwo:
It would be nice to have the possibility to set custom compass margins (the current padding parameter of `OrnamentSettings` sets padding for all ornaments, but I need to change only compass paddings) and custom compass image.
- On android it can be achieved by [setCompassMargins()](https://maplibre.org/maplibre-native/android/api/-map-libre%20-native%20-android/org.maplibre.android.maps/-ui-settings/set-compass-margins.html) and [setCompassImage()](https://maplibre.org/maplibre-native/android/api/-map-libre%20-native%20-android/org.maplibre.android.maps/-ui-settings/set-compass-image.html?query=open%20fun%20setCompassImage(compass:%20Drawable))
- I'm not so familiar with iOS, but [compassViewMargins](https://maplibre.org/maplibre-native/ios/latest/documentation/maplibre/mlnmapview/compassviewmargins) looks like a way to set margins, I don't know about `setCompassImage()` equivalent, maybe [MLNCompassButton](https://maplibre.org/maplibre-native/ios/latest/documentation/maplibre/mlncompassbutton)? 

The first screenshot is what I have now - a black compass image, covered by my button. 
The second one is what I'm trying to achieve - a compass image consistent with other buttons, under my button.
![ss](https://github.com/user-attachments/assets/92f1e0b4-b93b-4289-938f-7232af8d509b) ![ss2](https://github.com/user-attachments/assets/bf24920d-c2c0-493f-baee-4c6810d603e6)

westnordost:
I think the ornament settings are going away, as these use the MapLibre Android/iOS SDK implementations. Sargun is working on basing maplibre-compose (for native platforms, i.e Android, iOS, ...) directly on the MapLibre C++ API.

It would be more future-proof if you used the widget in the maplibre-compose-material3 artifact.

westnordost:
By the way, I noticed that your attribution info button is in the top left corner. Is this temporary or do you think this is the best location for that? This is relevant for the compose-based implementation of the attribution info that is currently in review: https://github.com/sargunv/maplibre-compose/pull/248
Currently, the button layouts itself towards the lower end side.

michalgwo:
> It would be more future-proof if you used the widgets from the maplibre-compose-material3 artifact. Are you using material3 on compose?

That's even better, I didn't know about the maplibre-compose-material3 artifact. Thanks! Although it would be nice to have size and needle drawable customizable.

[...]

PR #251, opened 2025-01-15
Made CompassButton more customizable

michalgwo:
As discussed in #250, I made Compass Button more customizable:

- allowed setting the size of the button
- allowed setting drawable of the needle

In addition, 2 changes that seemed reasonable:
- allowed setting content description of DisappearingCompassButton
- set the default content description of CompassButton to string resource, instead of hardcoded string

### DIFF
--- a/lib/maplibre-compose-material3/src/commonMain/kotlin/dev/sargunv/maplibrecompose/material3/controls/CompassButton.kt
+++ b/lib/maplibre-compose-material3/src/commonMain/kotlin/dev/sargunv/maplibrecompose/material3/controls/CompassButton.kt
@@ -47,7 +47,7 @@ public fun CompassButton(
   colors: ButtonColors = ButtonDefaults.elevatedButtonColors(),
   contentDescription: String = stringResource(Res.string.compass),
   size: Dp = 48.dp,
-  contentPadding: PaddingValues = PaddingValues(8.dp),
+  contentPadding: PaddingValues = PaddingValues(size / 6),
   shape: Shape = CircleShape,
   needlePainter: Painter = painterResource(Res.drawable.compass_needle),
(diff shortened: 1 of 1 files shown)


westnordost (commented):
Does this actually place the chosen icon in the center when a very large size is specified (or the icon is very small)?

michalgwo:
Yes. The sizes on the screenshots are 30.dp and 300.dp. The icon is vector drawable with 24 dp height and 24 dp width, I don't think the size of the drawable matters as it's in the centre no matter how big the button is. What I noticed now, the icon may go out of bounds when the button is big enough (as you can see in the last screenshot), and barely noticeable, when the button is small, so it may be a good idea to have contentPadding customizable as well.

73e08dc7-6a9c-43f1-a324-3d286594743a 135b8a7f-b94c-41e8-a48b-2d42470030a6 ca0ed5e6-5075-4880-a293-cd72ebdd6118

michalgwo:
Now it looks better. I also made the shape of the button customizable, as it was the last thing that wasn't customizable.

 24345b36-ff72-419b-a729-9bf8e35a48fd ad68584b-4eb1-4477-bd68-5e536b5b6940

michalgwo:
Actually, it may be a good idea to make the default contentPadding depend on the size, instead of hardcoding 8.dp, any thoughts on this?
``` kotlin
contentPadding: PaddingValues = PaddingValues(size / 6)
```

sargunv:
Thanks for the PR! Looks reasonable to me, will review properly in a bit. 

> it may be a good idea to make the default contentPadding depend on the size, instead of hardcoding 8.dp, any thoughts on this?

That sounds good to me, though I don't have a strong preference either way.

Issue #39, opened 2024-11-27
Support querying vector source features


michalgwo:
It would be useful to support querying vector source features.

Functions from MapLibre Native:
- Android: [querySourceFeatures](https://maplibre.org/maplibre-native/android/api/-map-libre%20-native%20-android/org.maplibre.android.style.sources/-vector-source/query-source-features.html)
- iOS: [featuresInSourceLayersWithIdentifiers:predicate:](https://maplibre.org/maplibre-native/ios/latest/documentation/maplibre/mlnvectortilesource/featuresinsourcelayerswithidentifiers:predicate:/) (I didn't test it, but judging by the documentation, it looks like the equivalent of Android's `querySourceFeatures`

sargunv:
Also cluster inspection methods (see TODO in #79)

PR #255, opened 2025-01-17
Getting specific type of source, querying source features on android and expose sources

michalgwo:
Related to #39, but not fixing the issue entirely, as the implementations are only for Android. I don't have access to a Mac right now, so I can't implement it for iOS.

Example of usage:
``` kotlin
val vectorSource = styleState.getSource("openmaptiles") as? VectorSource
val features = vectorSource?.querySourceFeatures(setOf("landuse"), null)
```

The first commit was to make it possible to cast Source to VectorSource, GeoJsonSource, and RasterSource. Exposed `getSource()` and `getSources()` make it possible to get a source outside of the Composable function as opposed to `getBaseSource()`


sargunv:
Thanks for this, I'll take care of the iOS side before merging

michalgwo:
I just realized that I set the predicate of `querySourceFeatures` as `CompiledExpression` instead of `Expression`. I fixed this and made it non-nullable.

michalgwo:
I'm sorry for so many changes after submitting the PR. I forgot about a few things. Now, everything should be fine.

PR #256, opened 2025-01-19
Fixed log message

michalgwo:
Fixed log message in LayerManager


michalgwo:
Actually it was correct, my bad.

PR #258, opened 2025-01-21
Animate camera to LatLngBounds

michalgwo:
- Android target - works fine
- iOS target - unfortunately I don't have a Mac right now, so I can't implement it. It looks like [camera:fittingCoordinateBounds:edgePadding:](https://maplibre.org/maplibre-native/ios/latest/documentation/maplibre/mlnmapview/camera:fittingcoordinatebounds:edgepadding:) or [setVisibleCoordinateBounds ](https://maplibre.org/maplibre-native/ios/latest/documentation/maplibre/mlnmapview/setvisiblecoordinatebounds:edgepadding:animated:completionhandler:)is the way to go
- JS target - works fine, but `FitBoundsOptions` doesn't have a duration parameter, so currently it has no effect on JS. It was my first time working with JS target, so I'm unsure if I've done everything right.
- Desktop target - the other `animateCameraPosition()` is not implemented on the desktop, so I left it unimplemented as well

Example of usage: 
``` kotlin
cameraState.animateTo(
  BoundingBox(
    northeast = Position(20.009155194730226, 50.06807238031317),
    southwest = Position(19.91854897235649,50.04007893350368)
  ),
  padding = PaddingValues(50.dp),
  tilt = 0.0,
  bearing = 0.0
)
```


michalgwo:
I just realized there is a BoundingBox class, so it makes more sense to use this, instead of creating a new LatLngBounds.

sargunv:
Thanks for this!

PR #265, opened 2025-01-30
Fixed crash when style is not fully loaded

### LINKED ISSUE #252: Crash when changing styles quickly on Android
When changing styles quickly, my Android app crashes with this error:
```
          Fatal Exception: java.lang.IllegalStateException: Calling getSources when a newer style is loading/has loaded.
       at org.maplibre.android.maps.Style.validateState(Style.java:776)
       at org.maplibre.android.maps.Style.getSources(Style.java:107)
       at dev.sargunv.maplibrecompose.core.AndroidStyle.getSources(AndroidStyle.kt:27)
       at dev.sargunv.maplibrecompose.compose.engine.SourceManager.<init>(SourceManager.kt:7)
       at dev.sargunv.maplibrecompose.compose.engine.StyleNode.<init>(StyleNode.kt:8)
       at dev.sargunv.maplibrecompose.compose.engine.RememberStyleCompositionKt$rememberStyleComposition$1$1.invokeSuspend(rememberStyleComposition.kt:28)
       at kotlin.coroutines.jvm.internal.BaseContinuationImpl.resumeWith(ContinuationImpl.kt:33)
       at kotlinx.coroutines.DispatchedTask.run(DispatchedTask.kt:101)
       at androidx.compose.ui.platform.AndroidUiDispatcher.performTrampolineDispatch(AndroidUiDispatcher.android.kt:81)
       at androidx.compose.ui.platform.AndroidUiDispatcher.access$performTrampolineDispatch(AndroidUiDispatcher.android.kt:41)
       at androidx.compose.ui.platform.AndroidUiDispatcher$dispatchCallback$1.run(AndroidUiDispatcher.android.kt:57)
       at android.os.Handler.handleCallback(Handler.java:883)
       at android.os.Handler.dispatchMessage(Handler.java:100)
       at android.os.Looper.loop(Looper.java:224)
       at android.app.ActivityThread.main(ActivityThread.java:7590)
       at java.lang.reflect.Method.invoke(Method.java)
       at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:539)
       at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:950)
```

I managed to reproduce it on Style Switcher Demo, by changing styles very quickly, and got a little different error:

```
java.lang.IllegalStateException: Calling getLayers when a newer style is loading/has loaded.
	at org.maplibre.android.maps.Style.validateState(Style.java:776)
	at org.maplibre.android.maps.Style.getLayers(Style.java:268)
	at dev.sargunv.maplibrecompose.core.AndroidStyle.getLayers(AndroidStyle.kt:43)
	at dev.sargunv.maplibrecompose.compose.MaplibreMapKt$MaplibreMap$callbacks$1$1.layerNodesInOrder(MaplibreMap.kt:139)
	at dev.sargunv.maplibrecompose.compose.MaplibreMapKt$MaplibreMap$callbacks$1$1.onClick-GrvxvIE(MaplibreMap.kt:146)
	at dev.sargunv.maplibrecompose.core.AndroidMap._init_$lambda$7(AndroidMap.kt:169)
	at dev.sargunv.maplibrecompose.core.AndroidMap.$r8$lambda$3_KOftxaIQ7_JozXXAVYsr09nZY(Unknown Source:0)
	at dev.sargunv.maplibrecompose.core.AndroidMap$$ExternalSyntheticLambda4.onMapClick(D8$$SyntheticClass:0)
	at org.maplibre.android.maps.MapGestureDetector.notifyOnMapClickListeners(MapGestureDetector.java:1006)
	at org.maplibre.android.maps.MapGestureDetector$StandardGestureListener.onSingleTapConfirmed(MapGestureDetector.java:362)
	at org.maplibre.android.gestures.StandardGestureDetector$1.onSingleTapConfirmed(StandardGestureDetector.java:83)
	at android.view.GestureDetector$GestureHandler.handleMessage(GestureDetector.java:323)
	at android.os.Handler.dispatchMessage(Handler.java:107)
	at android.os.Looper.loop(Looper.java:224)
	at android.app.ActivityThread.main(ActivityThread.java:7590)
	at java.lang.reflect.Method.invoke(Native Method)
	at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:539)
	at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:950)
```

I don't know if the crash also occurs on other platforms, but on Android it can be fixed simply by checking [isFullyLoaded](https://maplibre.org/maplibre-native/android/api/-map-libre%20-native%20-android/org.maplibre.android.maps/-style/is-fully-loaded.html?query=open%20fun%20isFullyLoaded():%20Boolean) before every interaction with the style.

michalgwo:
Fixes #252 on Android. This crash doesn't seem to happen on the JS target. I don't know if the problem also occurs on iOS, I was looking at the documentation and couldn't find `isFullyLoaded` equivalent, so it may be an Android-specific crash, but it would be worth to test it also on iOS.

I'm not so familiar with the codebase here, so I'm unsure if skipping the execution of functions or returning a null/empty list is a good solution, so I would appreciate feedback on this. 

Based on my investigation of the code, it looks like all the internal executions of the style's functions happen after the style is already loaded, so it shouldn't skip any execution that should happen on a loaded style.

The crash happens on changing the style trying to execute functions of the dismissed style, so it makes sense to skip it, as it will be executed again after a new style is fully loaded. 

The cases I can think of when this crash could happen:
- after the previous style is dismissed, but before a new style is loaded
- after a new style started loading, but the initialization code from the previous style didn't finish executing yet

### DIFF
--- a/lib/maplibre-compose/src/androidMain/kotlin/dev/sargunv/maplibrecompose/core/AndroidStyle.kt
+++ b/lib/maplibre-compose/src/androidMain/kotlin/dev/sargunv/maplibrecompose/core/AndroidStyle.kt
@@ -19,11 +19,11 @@ internal class AndroidStyle(style: MLNStyle) : Style {
   private var impl: MLNStyle = style
 
   override fun addImage(id: String, image: ImageBitmap, sdf: Boolean) {
-    impl.addImage(id, image.asAndroidBitmap(), sdf)
+    if (impl.isFullyLoaded) impl.addImage(id, image.asAndroidBitmap(), sdf)
   }
 
   override fun removeImage(id: String) {
-    impl.removeImage(id)
+    if (impl.isFullyLoaded) impl.removeImage(id)
   }
 
   private fun MLNSource.toSource() =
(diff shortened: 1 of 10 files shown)


sargunv:
thanks for looking into this! with this fix, instead of crashing, we just won't add the images, layers, etc.

Do you know if the style being operated on at the time of the crash is the outgoing style or the incoming style? If it's outgoing this is fine, but if it's incoming this would be a bug.

It might be better to make a change elsewhere in the stack to ensure we don't make style changes on outgoing styles (there's already an attempt at this by setting a no-op style implementation, but I guess it's missing something)

michalgwo:
I took a different approach now, it seems to be a better solution. Previously, the crash still happened in some cases, but now I can't reproduce it anymore. It's implemented in common code, so it will affect all platforms, not only Android, and now we can be sure that `isLoaded` is `false` only on the outgoing style. The drawback of this approach is that we need to check if style `isLoaded` before every call of style's functions, otherwise, it can throw `IllegalStateException` when changing styles.

sargunv:
Could you see if the bug can be replicated with the changes in https://github.com/sargunv/maplibre-compose/pull/269? It's an approach somewhere in the middle of your two approaches here.

It's working fine on my machine, but I haven't been able to replicate the crash at all so I'm working blind here 😅

michalgwo:
> (needs feedback from @michalgwo )

The continuation of this is in #269, I think @sargunv's approach is better, but it needs a few fixes I mentioned there.

PR #266, opened 2025-01-30
wip: map snapshotter

### LINKED ISSUE #28: Support snapshots
* iOS: https://maplibre.org/maplibre-native/ios/latest/documentation/maplibre-native-for-ios/staticsnapshotexample/
* Android: https://maplibre.org/maplibre-native/android/examples/snapshotter/

We should provide snapshot functionality to a similar extent as the native SDKs.

This may also come in handy for automated testing.

michalgwo:
Fixes #28 

I implemented snapshot functionality for Android and iOS. There is no snapshotter in the MapLibre GL JS library, so I left it empty.

I also created a demo presenting snapshot functionality.

As for iOS, as I already mentioned in my previous PRs, I don't have a Mac, so I couldn't run it, but this time I tried my best to implement it for iOS as well, but as it's not tested, and I don't have any experience with iOS, it may not work :sweat_smile: Let me know if you are not interested in fixing it, so I will remove the iOS part, and leave only Android implementation.

A few remarks about the iOS implementation:
- I wasn't sure what size should I use for `CameraPosition.toMLNCamera()`, so I took my shot and used the same as the size of the snapshot
- on Android `cameraPosition` param for snapshotter options is nullable, but on iOS it's not, I'm not sure if it's the best solution, but I made it nullable in common code and on iOS created a new empty `MLNMapCamera()` when `cameraPosition` is `null`
- I'm not sure about the error handling, I couldn't find any information in the documentation on what's inside the `String` returned on Android and `NSError` returned on iOS. I also failed to make the snapshotter return error, so I have no idea what kind of error is there.
- the function for converting UIImage to ImageBitmap is not written by me, I took it from [this](https://medium.com/mobile-innovation-network/uiimage-to-imagebitmap-compose-multiplatform-497230a71f37) article


sargunv:
Thanks for this! Currently planning on reviewing next week or so.

lib/maplibre-compose/src/androidMain/kotlin/dev/sargunv/maplibrecompose/core/util/util.kt
```diff
@@ -70,6 +74,37 @@ internal fun BoundingBox.toLatLngBounds(): LatLngBounds =
...
 
+internal fun MLNCameraPosition.toCameraPosition(): CameraPosition =
+  CameraPosition(
+    target = target?.toPosition() ?: Position(0.0, 0.0),
+    zoom = zoom,
+    bearing = bearing,
+    tilt = tilt,
+    padding =
+      padding?.let {
+        PaddingValues.Absolute(left = it[0].dp, top = it[1].dp, right = it[2].dp, bottom = it[3].dp)
```
westnordost, line 85: I think there is a bug here ;-)

lib/maplibre-compose/src/jsMain/kotlin/dev/sargunv/maplibrecompose/core/JsMapSnapshotter.kt
westnordost, line 1: Is this a " to be implemented" todo or is it not possible to support it on JS? Either way, should either add a `TODO` comment or a comment that explains why it cannot be supported on JS.

lib/maplibre-compose/src/commonMain/kotlin/dev/sargunv/maplibrecompose/core/MapSnapshotter.kt
```diff
@@ -0,0 +1,20 @@
+package dev.sargunv.maplibrecompose.core
+
+import androidx.compose.ui.graphics.ImageBitmap
+import androidx.compose.ui.unit.Dp
+import io.github.dellisd.spatialk.geojson.BoundingBox
+
+public interface MapSnapshotter {
```
westnordost, line 7: Shouldn't this interface be `internal`? It seems that it is only accessed from `CameraState`, and that's where the public interface is at.

lib/maplibre-compose/src/commonMain/kotlin/dev/sargunv/maplibrecompose/compose/CameraState.kt
```diff
@@ -197,4 +199,36 @@ public class CameraState internal constructor(firstPosition: CameraPosition) {
...
+    callback: (ImageBitmap) -> Unit,
+    errorHandler: (String) -> Unit = {},
+  ) {
+    requireMap()
+      .getMapSnapshotter()
+      .snapshot(width, height, styleUri, region, cameraPosition, showLogo, callback, errorHandler)
+  }
+
+  /** Cancels the map snapshotter if it is currently active. */
+  public fun cancelSnapshotter() {
```
westnordost, line 231: I figure that this library doesn't have (and doesn't intend to) have a dependency on Kotlin coroutines?

This kind of interface very much looks like it *should* be based on coroutines if this library uses coroutines elsewhere. (I.e. `suspend fun snapshot(...)`)

lib/maplibre-compose/src/commonMain/kotlin/dev/sargunv/maplibrecompose/compose/CameraState.kt
```diff
@@ -197,4 +199,36 @@ public class CameraState internal constructor(firstPosition: CameraPosition) {
     // TODO at some point, this should be refactored to State, just like the camera position
     return requireMap().getVisibleRegion()
   }
+
+  /**
+   * Takes a snapshot of the map with the specified parameters. Available on Android and iOS only.
```
westnordost, line 204: _Available_ or _functional_?

lib/maplibre-compose/src/commonMain/kotlin/dev/sargunv/maplibrecompose/compose/CameraState.kt
```diff
@@ -197,4 +199,36 @@ public class CameraState internal constructor(firstPosition: CameraPosition) {
...
+   *
+   * @param width The width of the snapshot in Dp.
+   * @param height The height of the snapshot in Dp.
+   * @param styleUri The URI of the style to use for the snapshot.
+   * @param region The bounding box of the region to capture in the snapshot.
+   * @param cameraPosition The camera position to use for the snapshot.
+   * @param showLogo Whether to show the logo in the snapshot. Defaults to true.
```
westnordost, line 211: Regarding logo - is this legacy from MapBox?

It appears to me this is almost useless, and much more important when making an image from the map would be to print the attribution into a corner.

lib/maplibre-compose/src/androidMain/kotlin/dev/sargunv/maplibrecompose/core/AndroidMapSnapshotter.kt
```diff
@@ -0,0 +1,62 @@
...
+    height: Dp,
+    styleUri: String,
+    region: BoundingBox?,
+    cameraPosition: CameraPosition?,
+    showLogo: Boolean,
+    callback: (ImageBitmap) -> Unit,
+    errorHandler: (String) -> Unit,
+  ) {
+    with(density) {
+      if (isStarted) return
```
westnordost, line 35: Is it important that this is threadsafe? (I.e. will bad things happen if two snapshots are done in parallel?) Because it isn't threadsafe. 

To make this threadsafe, `isStarted` would either have to be atomic or the whole block would need to be synchronized. Both would be done either by using https://github.com/Kotlin/kotlinx-atomicfu or by using the platform-native structures for concurrency on each platform (`AtomicBoolean`/`synchronized` on Android/Java and ...🤷‍♂️ on iOS)

lib/maplibre-compose/src/commonMain/kotlin/dev/sargunv/maplibrecompose/compose/CameraState.kt
michalgwo: There is an option for showing the logo both on iOS and Android, so I don't see any reason why not to implement it. As for attribution, it is printed by default, and I don't see any way to disable it.

lib/maplibre-compose/src/androidMain/kotlin/dev/sargunv/maplibrecompose/core/util/util.kt
michalgwo: Right

lib/maplibre-compose/src/jsMain/kotlin/dev/sargunv/maplibrecompose/core/JsMapSnapshotter.kt
michalgwo: It's a missing upstream feature, I will add a comment

lib/maplibre-compose/src/commonMain/kotlin/dev/sargunv/maplibrecompose/core/MapSnapshotter.kt
michalgwo: Right

lib/maplibre-compose/src/commonMain/kotlin/dev/sargunv/maplibrecompose/compose/CameraState.kt
michalgwo: I think functional fits better here

lib/maplibre-compose/src/commonMain/kotlin/dev/sargunv/maplibrecompose/compose/CameraState.kt
michalgwo: It uses coroutines actually, gonna fix it

lib/maplibre-compose/src/androidMain/kotlin/dev/sargunv/maplibrecompose/core/AndroidMapSnapshotter.kt
michalgwo: It's not that important I guess, gonna remove it

michalgwo:
Thank you for your review, I fixed all the things you mentioned.

lib/maplibre-compose/src/commonMain/kotlin/dev/sargunv/maplibrecompose/core/SnapshotResponse.kt
```diff
@@ -0,0 +1,11 @@
...
+
+import androidx.compose.ui.graphics.ImageBitmap
+
+public sealed class SnapshotResponse {
+  public data class Success(val image: ImageBitmap) : SnapshotResponse()
+
+  public data class Error(val error: String) : SnapshotResponse()
+
+  public data object Cancelled : SnapshotResponse()
+}
```
westnordost, line 11: Wouldn't it be an easier interface if the `MapSnapshotter` simply returned an `ImageBitmap`?

When an error occurs, it can simply be thrown as an e.g. `SnapshotException` and documented accordingly. The cancelled thing is not necessary at all, I think? After all, when it is cancelled, there is noone left to read the result, right?

I am not entirely sure about the "best code practices" regarding Kotlin coroutines, but I have been under the impression that `suspend` functions should feel like normal functions from the point of view of the caller. A `Result` kind of enum/sealed class with a `Success` and `Error` value seems more Rust-idiomatic than Kotlin-idiomatic.

lib/maplibre-compose/src/androidMain/kotlin/dev/sargunv/maplibrecompose/core/AndroidMapSnapshotter.kt
```diff
@@ -0,0 +1,57 @@
...
+          snapshotter.start({ snapshot ->
+            cont.resume(SnapshotResponse.Success(snapshot.bitmap.asImageBitmap()))
+          }) { error ->
+            cont.resume(SnapshotResponse.Error(error))
+          }
+        }
+      } catch (e: CancellationException) {
+        snapshotter.cancel()
+        SnapshotResponse.Cancelled
+      }
```
westnordost, line 54: Check out the [`suspendCancellableCoroutine`](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines/suspend-cancellable-coroutine.html) documentation regarding cancellation. It should rather be like

```kotlin
suspendCancellableCoroutine { cont ->
    snapshotter.startTheThing()
    cont .invokeOnCancellation { snapshotter.cancel() }
}
```

lib/maplibre-compose/src/commonMain/kotlin/dev/sargunv/maplibrecompose/core/SnapshotResponse.kt
michalgwo: Thank you for your research! I will be busy in the next couple of days, but I should be back working on this next week.

westnordost (approved):
👍 

Super clean implementation

michalgwo:
> 👍
> 
> Super clean implementation

Appreciate your review! Did you test it on iOS? As I mentioned, I didn't, because I don't have a Mac right now. I will probably get one in a couple of weeks, so if there is any problem, I can fix it.

westnordost:
I didn't test anything, just read the code.

michalgwo:
I fixed the iOS implementation, so PR is fully ready for review

westnordost:
Btw in this PR or after this PR has been merged:

Should update the list of implemented features: https://github.com/sargunv/maplibre-compose/blob/main/docs/docs/index.md

michalgwo:
Good catch! Fixed

lib/maplibre-compose/src/androidMain/kotlin/dev/sargunv/maplibrecompose/compose/AndroidMapView.kt
```diff
@@ -68,6 +69,7 @@ internal fun AndroidMapView(
               mapView = mapView,
               map = map,
               scaleBar = AndroidScaleBar(context, mapView, map),
+              mapSnapshotter = AndroidMapSnapshotter(context, layoutDir, density),
```
sargunv, line 72: we need to make sure layout dir and density gets updated when it changes (see below in the `update` callback)

lib/maplibre-compose/src/commonMain/kotlin/dev/sargunv/maplibrecompose/compose/CameraState.kt
```diff
@@ -197,4 +199,29 @@ public class CameraState internal constructor(firstPosition: CameraPosition) {
...
+   * @return The resulting ImageBitmap of the snapshot.
+   * @throws IllegalStateException if the map is not initialized yet. See [awaitInitialized].
+   * @throws SnapshotException if an error occurs during snapshot generation.
+   */
+  public suspend fun snapshot(
+    width: Dp,
+    height: Dp,
+    styleUri: String,
+    region: BoundingBox? = null,
+    cameraPosition: CameraPosition? = null,
```
sargunv, line 221: are `region` and `cameraPosition` somewhat redundant?

lib/maplibre-compose/src/androidMain/kotlin/dev/sargunv/maplibrecompose/core/AndroidMap.kt
```diff
@@ -253,47 +255,14 @@ internal class AndroidMap(
...
-          padding?.let {
-            PaddingValues.Absolute(
-              left = it[0].toInt().toDp(),
-              top = it[1].toInt().toDp(),
-              right = it[2].toInt().toDp(),
-              bottom = it[3].toInt().toDp(),
-            )
-          } ?: PaddingValues.Absolute(0.dp),
-      )
-    }
```
sargunv, line 273: I guess these were unused?

sargunv: ah yeah I see the imports from util

lib/maplibre-compose/src/commonMain/kotlin/dev/sargunv/maplibrecompose/compose/CameraState.kt
```diff
@@ -197,4 +199,29 @@ public class CameraState internal constructor(firstPosition: CameraPosition) {
...
+   * @param region The bounding box of the region to capture in the snapshot.
+   * @param cameraPosition The camera position to use for the snapshot.
+   * @param showLogo Whether to show the logo in the snapshot. Defaults to true.
+   * @return The resulting ImageBitmap of the snapshot.
+   * @throws IllegalStateException if the map is not initialized yet. See [awaitInitialized].
+   * @throws SnapshotException if an error occurs during snapshot generation.
+   */
+  public suspend fun snapshot(
+    width: Dp,
+    height: Dp,
```
sargunv, line 218: Should a snapshot dimensions be in `Dp`, or in actual image pixels?

I'd expect two snapshots of the same region, exported from two different devices, to be identical, not dependent on the display density of the device they came from.

demo-app/src/commonMain/kotlin/dev/sargunv/maplibrecompose/demoapp/demos/SnapshotterDemo.kt
```diff
@@ -0,0 +1,151 @@
...
+                  cameraState.snapshot(
+                    width = 500.dp,
+                    height = 500.dp,
+                    styleUri = DEFAULT_STYLE,
+                    cameraPosition = cameraState.position,
+                  )
+
+                snapshot.value = response
+              } catch (error: SnapshotException) {
+                println("Error during snapshot generation: ${error.message}")
```
sargunv, line 121: ideally the demo should show a dialog (I just got a timeout b/c I tried the snapshot on slow train internet, and it was unclear why it stopped loading)

sargunv:
Thinking about the use cases for snapshots, I can think of two:
* implementing a "share" feature, where an image of something the user is viewing on the map is shared as an image through the system share dialog. In this case, we probably want the snapshot to include all the same styling (including programmatic layers) as the actual interactive map the user is using, though image dimensions may be different than the view
* displaying efficient static maps, where perhaps many maps would be displayed in a list, or the map doesn't need to be interactive. In this case, we probably don't want the snapshot API to be dependent on the Compose map / CameraState at all; rather it's just a KMP wrapper around the MLN snapshot APIs. Would still be good to support programmatic styles, though that might be out of scope for this first implementation.

Do you have some other use cases you're intending to use this for, where the API considerations are different?

sargunv (changes requested):
thanks for the PR! this is close; but I think the API needs some tweaks considering the above use cases before we merge

sargunv:
I can confirm the implementation does work on iOS
image

westnordost:
> implementing a "share" feature

And in context of multiplatform (i.e. desktop), a use case would be to print the map.

lib/maplibre-compose/src/commonMain/kotlin/dev/sargunv/maplibrecompose/compose/CameraState.kt
michalgwo: They indicate the place where the snaphotter should take the snapshot. On the demo for simplicity, it takes a snapshot of the currently visible region on the map, but we can take a snapshot of a totally different region as well.

lib/maplibre-compose/src/commonMain/kotlin/dev/sargunv/maplibrecompose/compose/CameraState.kt
```diff
@@ -197,4 +199,29 @@ public class CameraState internal constructor(firstPosition: CameraPosition) {
...
+    styleUri: String,
```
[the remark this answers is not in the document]
michalgwo: It should be possible. Layers added programmatically won't be visible as Snapshotter is independent of the loaded map. Android's snapshotter can set JSON style, which makes it easier to add layers in code to the snapshotter, but there is no option for that on iOS. I created an issue in maplibre-native for that (https://github.com/issues/created?issue=maplibre%7Cmaplibre-native%7C3052), which was partially resolved lately, but the snapshotter's part was skipped.

westnordost:
> it takes CameraPosition and region, but it requires only one of them.

That sounds like a good use case for a sealed interface! E.g.  

sealed interface SnapshotLocation {
  data object Camera(position: CameraPosition) : SnapshotLocation
  data object Bounds(bounds: BoundingBox) : SnapshotLocation
}

lib/maplibre-compose/src/commonMain/kotlin/dev/sargunv/maplibrecompose/compose/CameraState.kt
michalgwo: Make sense to have dimensions in pixels

westnordost:
That's... possible? after all, with tilt or camera rotation , the displayed area is not a bbox anymore. I'd presume that if a camera position is specified, at least if it has any rotation or tilt, the camera position takes precedence and the bbox is discarded(?)

lib/maplibre-compose/src/commonMain/kotlin/dev/sargunv/maplibrecompose/core/SnapshotResponse.kt
sargunv: Sorry for responding late here. My take on the Result pattern vs exceptions is similar to yours; use the result pattern when the error is expected to happen (like a network failure or disconnection) but exceptions when the error is likely a bug or exceptional case.

I believe however that the map snapshotting fails if the necessary tiles fail to download, such as due to an outage on the tile server, or a slow network? So I think this would fall under "expected" failures, and therefore the result pattern. Invalid styles I agree are more "exceptional".

lib/maplibre-compose/src/commonMain/kotlin/dev/sargunv/maplibrecompose/compose/CameraState.kt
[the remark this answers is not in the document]
sargunv: Can we pass an existing Style object (from the loaded map) to the Snapshotter?

Can we apply the style composition to the snapshotter's Style?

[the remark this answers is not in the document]
michalgwo: We can't pass it, but now I noticed that we can get the `MLNStyle` object from the `MLNMapSnapshotter` on iOS, then we can edit this style, e.g. `setStyleJSON`, `addLayer`, `addSource`. On Android, we can't get the `Style` object, but we can initiate `MapSnapshotter` with JSON, which makes it possible. We need to create JSON out of programmatically added Layers and Sources and combine it with the root style. But this would be out of the scope of this PR.

I also noticed that Android's Snapshotter has private `addSource()` and `addLayer()`, it would be an easy PR (unless there is a reason for them being private) for maplibre-native to just make them public, which would probably make it easier.

michalgwo:
> That's... possible? after all, with tilt or camera rotation , the displayed area is not a bbox anymore. I'd presume that if a camera position is specified, at least if it has any rotation or tilt, the camera position takes precedence and the bbox is discarded(?)

According to the documentation, when `region` and `CameraPosition` are set in conjunction, the `CameraPosition.target` is overridden by region. I don't know how it works under the hood, but my understanding is that the other `CameraPosition` params like tilt, rotation, etc. are applied. 

I just tested setting `region` (`BoundingBox`) with `CameraPosition`, which was rotated and tilted, and it works. The snapshot was rotated and tiled area from the `BoundingBox`. I was also trying `zoom` and `padding`, but it seems these don't have any effect on the snapshot.

sargunv:
Relevant comment at https://github.com/maplibre/maplibre-compose/issues/374#issuecomment-2972331475

My suggestion to support StyleState with programmatic styling is possible in Android but not in iOS without upstream MapLibre Native changes. So I think we can merge without that for now, and file issues to request that feature upstream (for offline tiles too actually)

I'll take this PR to the finish line and then release v0.9.0.

sargunv:
Ah nope I was wrong; programmatic styling is available in iOS: https://github.com/maplibre/maplibre-compose/issues/374#issuecomment-2972344377

So yeah, back to the original plan to land this after our style state refactor so we can use it here.

sargunv:
#834 integrates maplibre-native-ffi, which has a cpu readback function that works with the map texture directly. even a separate snapshotter I think will be built on that foundation now, as #572 moves to the same shared implementation

sargunv:
thanks for taking a swing at this, and sorry it got stuck for so long!

PR #277, opened 2025-02-19
Make rememberCameraState use rememberSaveable

Vichy97:
This should prevent camera position being reset after navigating away and then coming back. 
Addresses https://github.com/sargunv/maplibre-compose/issues/261

### DIFF
--- a/lib/maplibre-compose/src/commonMain/kotlin/dev/sargunv/maplibrecompose/compose/CameraState.kt
+++ b/lib/maplibre-compose/src/commonMain/kotlin/dev/sargunv/maplibrecompose/compose/CameraState.kt
@@ -27,10 +27,8 @@ import kotlinx.coroutines.channels.Channel
 /** Remember a new [CameraState] in the initial state as given in [firstPosition]. */
 @Composable
 public fun rememberCameraState(firstPosition: CameraPosition = CameraPosition()): CameraState {
-  val layoutDirection = LocalLayoutDirection.current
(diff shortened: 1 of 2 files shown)


sargunv:
> Any chance on a minor (or snapshot) release after this lands? The next release looks like it's a ways off but there have been some important bug fixes recently.

Yeah, I'd like to re-scope 0.7.0 for existing tweaks and bug fixes, and move feature work to v0.8.0

The snapshot release should also happen automatically within a day of merging.

sargunv:
Tested on Android and got: java.lang.IllegalArgumentException: dev.sargunv.maplibrecompose.compose.CameraState@71c521e cannot be saved using the current SaveableStateRegistry. The default implementation only supports types which can be stored inside the Bundle. Please consider implementing a custom Saver for this class and pass it to rememberSaveable().

sargunv (changes requested):
Tested on Android and got: java.lang.IllegalArgumentException: dev.sargunv.maplibrecompose.compose.CameraState@71c521e cannot be saved using the current SaveableStateRegistry. The default implementation only supports types which can be stored inside the Bundle. Please consider implementing a custom Saver for this class and pass it to rememberSaveable().

Looks like it needs a custom Saver.

Vichy97:
> Owner

My bad! I was having trouble running the demo app initially. I got it running and it seems to work with this new saver class.

Vichy97:
@sargunv could I get another review on this?

sargunv (approved):

sargunv:
This version lgtm

PR #281, opened 2025-03-02
solve typo: Interpolateable -> Interpolatable

westnordost:
A breaking change because name of public class is changed, hence v0.7.0.


westnordost:
(Not merging to `main` yet because I don't know whether @sargunv intends to do a 0.6.x release from `main` before a a 0.7.x release)

sargunv:
I'll do a 0.7 after merging some of the currently open stuff, hopefully within a week or so

PR #282, opened 2025-03-02
Camera padding fix

### LINKED ISSUE #270: Material Compass wrong behaviour when uneven padding is applied to the camera
### What happened?

When you set the padding to the camera (it happens only if padding at least on one side is different from the other sides), and then press the compass, the position of the camera changes. It should be easy to reproduce with the steps below, but please let me know if you can't, so I will attach a video.

It may be useful to see how malibre-native implemented compass click listener on [Android](https://github.com/maplibre/maplibre-native/blob/a577daefd067f6d5ba647e4c9dadcc6b49e43dc3/platform/android/MapLibreAndroid/src/main/java/org/maplibre/android/maps/MapView.java#L265). `maplibreMap.setFocalBearing()` is public, so we can easily implement this. I don't know about other platforms.

### Affected Platforms

Android

### Platform Version

Android API 29

### Library Version

0.6.0

### Sample Code

To reproduce:
1. Go to Camera State Demo and set padding to the camera, for example, `top = 10.dp`, `end = 100.dp`
``` kotlin
val cameraState = rememberCameraState(firstPosition = CameraPosition(target = CHICAGO, zoom = 15.0, padding = PaddingValues(top = 10.dp, end = 100.dp)))
```
2. Run the app and try using the compass. The deviation is small after clicking the first time, but it's bigger the second and third times, so try it a couple of times.

michalgwo:
Fixes #270

### DIFF
--- a/lib/maplibre-compose/src/androidMain/kotlin/dev/sargunv/maplibrecompose/core/AndroidMap.kt
+++ b/lib/maplibre-compose/src/androidMain/kotlin/dev/sargunv/maplibrecompose/core/AndroidMap.kt
@@ -254,21 +254,23 @@ internal class AndroidMap(
   }
 
   private fun MLNCameraPosition.toCameraPosition(): CameraPosition =
-    CameraPosition(
-      target = target?.toPosition() ?: Position(0.0, 0.0),
-      zoom = zoom,
-      bearing = bearing,
-      tilt = tilt,
-      padding =
-        padding?.let {
-          PaddingValues.Absolute(
-            left = it[0].dp,
-            top = it[1].dp,
-            right = it[2].dp,
-            bottom = it[3].dp,
-          )
-        } ?: PaddingValues.Absolute(0.dp),
-    )
+    with(density) {
+      CameraPosition(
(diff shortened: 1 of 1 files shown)


westnordost (approved):
Cool! Indeed, that's a bug, thank you for the fix!

I've looked at the source code and noticed that on the iOS and JS side, it is handled consistently. For both, there's no conversion at all between `dip` and `px`. (`IosMap.kt` for iOS, `util.kt` for JS)

Maybe because the fields of `UIEdgeInsets` on iOS and `PaddingOptions` on JS are already in device independent pixels? No idea unfortunately 🤷‍♂️. Maybe @sargunv knows and follow up with a fix for iOS and JS, if necessary. 

In any case, this fix is fine.

westnordost:
> Maybe because the fields of UIEdgeInsets on iOS and PaddingOptions on JS are already in device independent pixels? No idea unfortunately 🤷‍♂️. Maybe @sargunv knows and would follow up with a fix for iOS and JS, if necessary.

Nevermind, it looks like in iOS, "raw pixels" are not exposed on public APIs at all, i.e. everything is screen pixel-density independent already.

PR #290, opened 2025-03-17
Fix issues & add features on scalebar

westnordost:
1. **fix**: when aligned to the right, the right end stroke would wobble on a sub-pixel level.
2. **fix**: the passed modifier was used twice, leading to layout bugs
3. **feature**: (breaking change) add `haloWidth` and `barWidth` as parameters
4. **change**: default to no halo, because that's consistent with the attribution button and also consistent with e.g. the system icons (there's also no halo around those in edge-to-edge)
5. **change**: default color is `LocalContentColor.current`, halo color is derived from that. Had to change the parameter order for that.


westnordost:
I noticed that if the view in which the map and the HUD controls are shown is not **contained in some `Surface`** and the user has **dark mode** on, the colors of the _attribution button_ and the _scale bar_ are inconsistent:

The scale bar is white (with a black halo) because of this:

  ```kotlin
haloColor: Color = MaterialTheme.colorScheme.surface,
color: Color = contentColorFor(haloColor),
 ```

The attribution button is black, because

```kotlin
iconColors: IconButtonColors = IconButtonDefaults.iconButtonColors(),
```

(and if there's no `Surface`, the default colors are simply the "light" colors).

---

So, one way to make this consistent would be to make the scalebar default colors like so

  ```kotlin
haloColor: Color = backgroundColorFor(color),
color: Color = LocalContentColor,
 ```

`ColorScheme.backgroundColorFor` doesn't exist yet, so this would need to be added.

```kotlin
@Stable
internal fun ColorScheme.backgroundColorFor(contentColor: Color): Color =
  when (contentColor) {
    onPrimary -> primary
    onSecondary -> secondary
    onTertiary -> tertiary
    onBackground -> background
    onError -> error
    onPrimaryContainer -> primaryContainer
    onSecondaryContainer -> secondaryContainer
    onTertiaryContainer -> tertiaryContainer
    onErrorContainer -> errorContainer
    inverseOnSurface -> inverseSurface
    onSurface -> surface
    onSurfaceVariant -> surfaceVariant
    else -> Color.Unspecified
  }
```

Would you agree that this should be done or do you prefer the way it is now?

sargunv:
Do you have a screenshot of the issue?

westnordost:
no_surface_dark_mode

FYI this can only be achieved if
1. the app does actually display in dark mode and 
2. no parent `provides` a `LocalContentColor`, like e.g. a `Surface` or `Scaffold` does.

So, I had to modify the code of the demo app (delete the `Scaffold`, set the flag to use dark mode)

sargunv:
Ah, I understand now. Yeah, that suggested change sounds reasonable to me, with an explanatory comment in the code

PR #291, opened 2025-03-20
Sources state

### LINKED ISSUE #164: refactor attribution info as actual State in StyleState

westnordost:
Fixes #164

- Adds `public val sources: List<Source>` to `StyleState`
- Removes `getSources` and `queryAttributionLinks` from `StyleState`


westnordost:
On Android, It looks like all [`Source.getAttributions`](https://github.com/maplibre/maplibre-native/blob/e069889f7130ff3564a39d29ba3b6821c3f6fea0/platform/android/MapLibreAndroid/src/main/java/org/maplibre/android/style/sources/Source.java#L68) in the loaded style are (still) empty when the style is loaded, i.e. when either
- the `callback: Style.OnStyleLoaded` from [`MapLibreMap.setStyle`](https://github.com/maplibre/maplibre-native/blob/e069889f7130ff3564a39d29ba3b6821c3f6fea0/platform/android/MapLibreAndroid/src/main/java/org/maplibre/android/maps/MapLibreMap.java#L945) 
- but also when the `OnDidFinishLoadingStyleListener` listener from the `MapView`

is invoked. The same for iOS.

The style (e.g. https://tiles.openfreemap.org/styles/bright) at this point has been loaded, including a list of `Source`s, but they don't contain any information, probably because the tilejsons of these sources (e.g. https://tiles.openfreemap.org/planet)  referred to in the style are loaded just after, asynchronously.

Luckily, on Android, one can add another callback [`MapView.addOnSourceChangedListener`](https://github.com/maplibre/maplibre-native/blob/e069889f7130ff3564a39d29ba3b6821c3f6fea0/platform/android/MapLibreAndroid/src/main/java/org/maplibre/android/maps/MapView.java#L845) in order to be notified via a `OnSourceChangedListener` of changes to any source (e.g. it being loaded).

Unluckily, while the iOS code also [gets notified of changes to a source](https://github.com/maplibre/maplibre-native/blob/e069889f7130ff3564a39d29ba3b6821c3f6fea0/platform/ios/src/MLNMapView%2BImpl.mm#L100) ...

```objective-c++
- (void)sourceDidChange:(MLNSource *)source {
    // no-op: we only show attribution after tapping the info button, so there's no
    // interactive update needed.
}
```
...this callback is just [ignored](
https://github.com/maplibre/maplibre-native/blob/e069889f7130ff3564a39d29ba3b6821c3f6fea0/platform/ios/src/MLNMapView.mm#L6903-L6906), and (thus) also not passed onto the `MLNMapViewDelegate` so that we could update the `StyleState` with an updated list of `Source`s. 😕

So, well, I don't know. This PR will not work on iOS, it is either blocked by a to-be-posted feature request in MapLibre for iOS to please pass on the source-did-change notification to library users, or, made obsolete by your works on using MapLibre core directly.

westnordost:
Created a feature request: https://github.com/maplibre/maplibre-native/issues/3330

westnordost:
Has been implemented! Will be unblocked on the next MapLibre SDK release for iOS.

westnordost:
Unblocked, as new maplibre for iOS just dropped. Converting back to draft because now the updating on iOS is missing.

westnordost:
@sargunv / @michalgwo hm, this PR is ready, however, iOS cannot build because [MapLibre-iOS 6.13.0](https://github.com/maplibre/maplibre-native/releases/tag/ios-v6.13.0) is not found. Maybe some really obvious mistake for someone who knows his way around iOS build setup?

michalgwo:
I don't have much experience with iOS, but I guess there is a problem with Cocoapods, as https://cocoapods.org/pods/MapLibre shows Internal Server Error

sargunv:
Looks like cocoapods is missing 6.13; perhaps someone can PR [here](https://github.com/CocoaPods/Specs/tree/master/Specs/b/0/0/MapLibre) to release it there? not sure how the release process there typically works

It is available [via SPM](https://swiftpackageindex.com/maplibre/maplibre-gl-native-distribution) so an alternative is figuring out how to replace cocoapods with that (cocoapods is largely being phased out anyway, but kotlin native doesn't yet support SPM officially)

sargunv:
reported here: https://github.com/maplibre/maplibre-native/issues/3457

sargunv (approved):

PR #293, opened 2025-03-21
Attribution Button with Popup instead of Dialog

### LINKED ISSUE #174: Change Material3 attribution button to an attribution bar visible by default

westnordost:
This PR changes the attribution button to open the attribution info in a popup instead of a dialog. The popup is initially shown and is hidden when the map is moved or the info button is clicked.

Fixes #174 

Supersedes #248

Developed on top of #291, which is blocked.

---

Much of the lines of code is used (`SuperimposingPopupPositionProvider.kt`, `Arrangement.kt`, ~30 lines of code in `AttributionButton.kt`) to make the popup appear over the info icon (rather than next to it like a tooltip) and placing another info icon at the very exact same spot within the popup, so that it appears as if the popup expands from below the info button.

https://github.com/user-attachments/assets/6f6f02ee-5d69-4d79-902a-6ff38779ae76

https://github.com/user-attachments/assets/66c20023-1d4e-40b0-a5c2-0a1e1c521eab

The attribution text in just a simple popup would be just the first commit, in case you prefer that.


michalgwo:
Small suggestion. Maybe we can filter out double attributions (i.e. if 2 sources have the same attribution)? Example of such a style: `https://demotiles.maplibre.org/styles/osm-bright-gl-terrain/style.json`

How it looks like when 2 sources have the same attribution:
92c92358-fc1b-4ac1-b35d-abbd3917a318

westnordost:
Makes sense to me, but then both the label and the link must be 100% the same.

michalgwo:
There is something wrong with the paddings. On my screen, there is a little movement of the icon while opening/closing attribution:

https://github.com/user-attachments/assets/eecfc712-63d4-48a5-81af-7fad713ccb1d

It would also be nice to have padding and `Surface`'s parameters customizable, to be specific, parameters shape, color, contentColor, tonalElevation, shadowElevation, and border in this line:
https://github.com/sargunv/maplibre-compose/blob/7ad41a5125ec712f7264c20984815f2ccc2d71d9/lib/maplibre-compose-material3/src/commonMain/kotlin/dev/sargunv/maplibrecompose/material3/controls/AttributionButton.kt#L107

By padding, I mean the outer padding of the `Surface`. For example, when I have it on `TopStart`, I want to add padding to the end to make space for the compass button.

westnordost:
I noticed this too, on some configurations, especially when display cutouts that protruded into the map were involved (on the aligned side). But the position returned by  `PopupPositionProvider` is fine. I think it is just that Compose (or Android?) may decide to put the popup somewhere else after all than what the `PopupPositionProvider` returned if it doesn't fit.

And to my knowing, there is no way to *force* Compose (or Android?) to treat the placement position as highest priority, forcing the popup contents to shrink.

Anyway, if you find a solution, you could create a PR that merges into this PR, but I myself found no solution. And, well, it just amounts to a slight cosmetic defect.

> It would also be nice to have [...] Surface's parameters customizable, to be specific, parameters shape, color, contentColor, tonalElevation, shadowElevation, and border in this line:

Sure, why not. Will you create a PR that merges into this one?

> By padding, I mean the outer padding of the Surface. For example, when I have it on TopStart, I want to add padding to the end to make space for the compass button.

Uh, but the whole point of this exercise was to provide a way to make the popup *overlay* other UI elements, without affecting the layout of those other UI elements (i.e. to supersede #248)

michalgwo:
> Anyway, if you find a solution, you could create a PR that merges into this PR, but I myself found no solution. And, well, it just amounts to a slight cosmetic defect.

I was looking for a solution, but I couldn't find any

> Sure, why not. Will you create a PR that merges into this one?

Sure

> Uh, but the whole point of this exercise was to provide a way to make the popup _overlay_ other UI elements, without affecting the layout of those other UI elements (i.e. to supersede #248)

Why not give users the choice of whether they want to overlay it or place it relative to other elements?

westnordost:
> Why not give users the choice of whether they want to overlay it or place it relative to other elements?

Because the implementation is completely different. See the prior PR.

westnordost:
Notes to myself:

- The end spacer seems to be ignored if the attribution text would otherwise fill the whole screen width
- maybe add `|` between the single attribution texts for better separation? I think MapLibre-Js attribution popup does this

sargunv:
Something is off with the padding (perhaps I broke it while resolving merge conflicts)

CleanShot 2025-05-17 at 20 32 26

michalgwo:
> Something is off with the padding (perhaps I broke it while resolving merge conflicts)

Looks like a compose multiplatform bug to me. It works fine on Android, but not on iOS. Adding text style somehow stretches the composable. This line, to be specific:
https://github.com/sargunv/maplibre-compose/blob/765e522fcd026216f4ded00dbac7eca72ab5312d/lib/maplibre-compose-material3/src/commonMain/kotlin/dev/sargunv/maplibrecompose/material3/controls/AttributionButton.kt#L210
I tried setting text style as a parameter of `Text`, but the effect was the same.

My investigation:
I added `onSizeChanged` modifiers to composables in `AttributionButton` and found out that on Android `onSizeChanged` was called only once for `FlowRow` (composable below) with correct width, but on iOS it was called twice - first time with width almost equal to the width of the screen, the second time with correct width. But when width was changed a second time, the width of the outer composables remained the same.

https://github.com/sargunv/maplibre-compose/blob/765e522fcd026216f4ded00dbac7eca72ab5312d/lib/maplibre-compose-material3/src/commonMain/kotlin/dev/sargunv/maplibrecompose/material3/controls/AttributionButton.kt#L211-L214

michalgwo:
Another problem is when changing styles, the attribution is blinking

https://github.com/user-attachments/assets/a03b8aa0-a24b-47b2-876f-698a8096941f

sargunv:
closing this one out in favor of #349, but feel free to reopen if you'd still like to add a PopupAttributionButton or something

PR #307, opened 2025-04-14
Attribution Button improvements

michalgwo:
- fixed movement/extension of popup
- added padding at the end of the attribution popup
- made attribution popup customizable

Developed on top of #293

### DIFF
--- a/lib/maplibre-compose-material3/src/commonMain/kotlin/dev/sargunv/maplibrecompose/material3/controls/AttributionButton.kt
+++ b/lib/maplibre-compose-material3/src/commonMain/kotlin/dev/sargunv/maplibrecompose/material3/controls/AttributionButton.kt
@@ -4,6 +4,7 @@ import androidx.compose.animation.AnimatedVisibility
 import androidx.compose.animation.core.MutableTransitionState
 import androidx.compose.animation.fadeIn
 import androidx.compose.animation.fadeOut
+import androidx.compose.foundation.BorderStroke
 import androidx.compose.foundation.layout.Arrangement
 import androidx.compose.foundation.layout.Box
 import androidx.compose.foundation.layout.ExperimentalLayoutApi
@@ -23,6 +24,7 @@ import androidx.compose.material3.MaterialTheme
 import androidx.compose.material3.ProvideTextStyle
 import androidx.compose.material3.Surface
 import androidx.compose.material3.Text
+import androidx.compose.material3.contentColorFor
 import androidx.compose.runtime.Composable
 import androidx.compose.runtime.LaunchedEffect
 import androidx.compose.runtime.getValue
@@ -31,6 +33,8 @@ import androidx.compose.runtime.remember
 import androidx.compose.runtime.setValue
 import androidx.compose.ui.Alignment
 import androidx.compose.ui.Modifier
+import androidx.compose.ui.graphics.Color
+import androidx.compose.ui.graphics.Shape
 import androidx.compose.ui.platform.LocalLayoutDirection
 import androidx.compose.ui.text.LinkAnnotation
 import androidx.compose.ui.text.TextLinkStyles
@@ -60,18 +64,34 @@ import org.jetbrains.compose.resources.stringResource
  *   [Source.attributionLinks][dev.sargunv.maplibrecompose.core.source.Source.attributionLinks] via
  *   [StyleState.sources][dev.sargunv.maplibrecompose.compose.StyleState.sources]
  * @param modifier the Modifier to be applied to this layout node
- * @param colors Colors that will be used for the info button
+ * @param iconColors Colors that will be used for the info button
  * @param textStyle Text style used for the attribution info
  * @param textLinkStyles Text link styles that should be used for the links in the attribution info
+ * @param popupEndPadding Padding that will be applied on the end of the popup (i.e. if the popup is
+ *   aligned to the start, padding will be applied on the end, but if the popup is aligned to the
+ *   end, padding will be applied to the start)
+ * @param popupShape Shape of the popup (applied to [Surface])
+ * @param popupColor Color of the popup (applied to [Surface])
+ * @param popupContentColor Content Color of the popup (applied to [Surface])
+ * @param popupTonalElevation Tonal Elevation of the popup (applied to [Surface])
+ * @param popupShadowElevation Shadow Elevation of the popup (applied to [Surface])
+ * @param popupBorder Border of the popup (applied to [Surface])
  */
 @Composable
 public fun AttributionButton(
   lastCameraMoveReason: CameraMoveReason,
   attributions: List<AttributionLink>,
   modifier: Modifier = Modifier,
-  colors: IconButtonColors = IconButtonDefaults.iconButtonColors(),
+  iconColors: IconButtonColors = IconButtonDefaults.iconButtonColors(),
   textStyle: TextStyle = MaterialTheme.typography.bodyMedium,
   textLinkStyles: TextLinkStyles? = null,
+  popupEndPadding: Dp = 12.dp,
+  popupShape: Shape = RoundedCornerShape(24.dp),
+  popupColor: Color = MaterialTheme.colorScheme.surface,
+  popupContentColor: Color = contentColorFor(popupColor),
+  popupTonalElevation: Dp = 0.dp,
+  popupShadowElevation: Dp = 0.dp,
+  popupBorder: BorderStroke? = null,
 ) {
   if (attributions.isEmpty()) return
 
@@ -85,7 +105,7 @@ public fun AttributionButton(
   }
 
   Box(modifier) {
-    IconButton(onClick = { expanded.targetState = !expanded.currentState }, colors = colors) {
+    IconButton(onClick = { expanded.targetState = !expanded.currentState }, colors = iconColors) {
       InfoIcon()
     }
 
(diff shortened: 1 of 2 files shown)


lib/maplibre-compose-material3/src/commonMain/kotlin/dev/sargunv/maplibrecompose/material3/controls/AttributionButton.kt
```diff
@@ -98,13 +122,27 @@ public fun AttributionButton(
       val verticalAlignment = if (alignTop) Alignment.Top else Alignment.Bottom
       val horizontalArrangement =
         if (alignLeft) Arrangement.Absolute.Left else Arrangement.Absolute.Reverse
+      val isRTL = LocalLayoutDirection.current == LayoutDirection.Rtl
 
       Popup(
         popupPositionProvider = popupPositionProvider,
+        properties = PopupProperties(clippingEnabled = false),
```
westnordost, line 129: What's this for? (add comment?)

lib/maplibre-compose-material3/src/commonMain/kotlin/dev/sargunv/maplibrecompose/material3/controls/AttributionButton.kt
```diff
@@ -56,18 +64,34 @@ import org.jetbrains.compose.resources.stringResource
...
 @Composable
 public fun AttributionButton(
   lastCameraMoveReason: CameraMoveReason,
   attributions: List<AttributionLink>,
   modifier: Modifier = Modifier,
-  colors: IconButtonColors = IconButtonDefaults.iconButtonColors(),
+  iconColors: IconButtonColors = IconButtonDefaults.iconButtonColors(),
   textStyle: TextStyle = MaterialTheme.typography.bodyMedium,
   textLinkStyles: TextLinkStyles? = null,
+  popupEndPadding: Dp = 12.dp,
```
westnordost, line 88: where does this default value come from? Before it was 0, no?

lib/maplibre-compose-material3/src/commonMain/kotlin/dev/sargunv/maplibrecompose/material3/controls/AttributionButton.kt
```diff
@@ -135,6 +173,12 @@ public fun AttributionButton(
   }
 }
 
+private fun Modifier.paddingEndOfPopup(padding: Dp, alignLeft: Boolean, isRTL: Boolean): Modifier {
+  val alignEnd = if (isRTL) !alignLeft else alignLeft
+
+  return if (alignEnd) padding(end = padding) else padding(start = padding)
```
westnordost, line 179: you check the layout direction in order to cancel it and align it left/right after all.

Couldn't you just instead use [absolutePadding](https://developer.android.com/reference/kotlin/androidx/compose/ui/Modifier#(androidx.compose.ui.Modifier).absolutePadding(androidx.compose.ui.unit.Dp,androidx.compose.ui.unit.Dp,androidx.compose.ui.unit.Dp,androidx.compose.ui.unit.Dp))?

lib/maplibre-compose-material3/src/commonMain/kotlin/dev/sargunv/maplibrecompose/material3/controls/AttributionButton.kt
michalgwo: This fixes the problem of "movement of the icon" while opening/closing attribution. I found out that popup wasn't actually moving. When the text inside the popup took more space than was available, but enough to fit on the screen, the popup was extended to fit the text inside. I doubt the comment is needed here, when you look at the documentation, you can see "Whether to allow the popup window to extend beyond the bounds of the screen." which explains enough what it is for.

lib/maplibre-compose-material3/src/commonMain/kotlin/dev/sargunv/maplibrecompose/material3/controls/AttributionButton.kt
michalgwo: It was 0 before, but after I added `properties = PopupProperties(clippingEnabled = false)`, for some reason, the other side of the popup was extended outside of the screen up to a point where text still fits on the screen, so I decided to add some padding. 12dp is something I made up to prevent popup from extending outside of the screen, and it looks good. Let me know if you have a better idea for this.

For example, both screenshots below were taken when the popup was full (i.e. adding one more "i" letter, moves one word to the next line). The first one is with 0dp padding, the second one with 12dp padding.

ee483f4d-2194-40fc-9b32-11f6ff110041 340f99af-3476-4f9f-84e3-3802f0ff4e57

lib/maplibre-compose-material3/src/commonMain/kotlin/dev/sargunv/maplibrecompose/material3/controls/AttributionButton.kt
michalgwo: You are right, I didn't know about `absolutePadding`

lib/maplibre-compose-material3/src/commonMain/kotlin/dev/sargunv/maplibrecompose/material3/controls/AttributionButton.kt
westnordost: Hm well, I see. So the 12.dp is probably incidentally the margin to the right of the popup in your posted screenshot here.

So when clipping is not done automatically, i.e. Compose does not resize / reposition the popup automatically anymore, we'd need to do it ourselves. I think it is not okay to offload this to the library user.

So in this case,
- the default `popupEndPadding` as passed to `AttributionButton` should be `0.dp`
- we need to calculate the necessary padding to fit the popup onto the screen ourselves. E.g. the attribution button might be placed not directly in a corner, but at a certain distance to the corner. Imagine if the button was almost in the center, then the end padding must be almost half the screen. The `SuperimposingPopupPositionProvider` should have all the parameters it needs to be able to calculate the desired popup width and hence the padding.

(In my opinion, it is not worth the effort only to avoid a slight offset of the popup in some edge cases, but if it is done, in my opinion, it should be done proper, i.e. see above - don't require the user to fix the popup not being fully displayed on-screen by specifying some magic extra padding.)

michalgwo: Alright, now it should be fine.

I had set the popup `width` instead of `padding`, because padding was sometimes changing the `popupContentSize`, creating an infinite loop.

This is what it looks like when the popup is full with different `AttributionButton` alignment: `BottomStart`, `BottomEnd`, and `BottomCenter`

Zdjęcie WhatsApp 2025-04-17 o 14 49 44_f4266b76 Zdjęcie WhatsApp 2025-04-17 o 14 49 44_96f38fdc Zdjęcie WhatsApp 2025-04-17 o 14 49 44_275fb8c0

The implication of this approach is that, when there is more than one line of attribution, the popup extends to the bounds of the screen, instead of clipping at the end of the text as before, so I decided to align the text to the icon, which looks better to me.

Zdjęcie WhatsApp 2025-04-17 o 14 49 43_7dd080ff Zdjęcie WhatsApp 2025-04-17 o 14 49 44_8a7ff856

When there is only one line of attribution, it clips the popup as before.

Zdjęcie WhatsApp 2025-04-17 o 14 49 43_5ca75537

westnordost: Awesome! The third screenshot, is there just one attribution or is this somehow cut off? When the horizontal space isn't enough, it should of course extend vertically.

Also, I noticed that in screenshot one, two and three, the spacer padding I added at the end of the popup content seems to be missing?

That spacer is there to make the popup look "balanced", as explained in the comment:

```kotlin
// icon buttons are automatically padded to have a certain click size, which makes the
// popup appear misaligned if we don't also add some extra padding on the other side
Spacer(Modifier.width(16.dp))
```

michalgwo: > Awesome! The third screenshot, is there just one attribution or is this somehow cut off? When the horizontal space isn't enough, it should of course extend vertically.

On the third screenshot, nothing is cut off. I just wrote as much text to fill the whole line, if I added one more letter, it would be extended vertically, and the last word would be moved to the next line (same as in screenshots 4 and 5). It is a specific case, because the `AttributionButton` here is aligned to `BottomCenter`, which is not a usual place for `AttributionButton`. My point with this screenshot was to show how it will look like when `AttributionButton` is not in the corner.

> Also, I noticed that in screenshot one, two and three, the spacer padding I added at the end of the popup content seems to be missing?

I noticed this as well, but I didn't touch it. It seems like when the popup is not full (last screenshot), it applies the spacer correctly, but when the whole space on the popup is filled (screenshots 1,2,3), the text overlaps the spacer. It was like that on your branch before I started working on this.

westnordost: > I noticed this as well, but I didn't touch it. It seems like when the popup is not full (last screenshot), it applies the spacer correctly, but when the whole space on the popup is filled (screenshots 1,2,3), the text overlaps the spacer. It was like that on your branch before I started working on this.

Something I ought to look into when this PR is merged.

michalgwo: > Also, why/how is the `FlowRow` affected by the `TextAlign`? 🤔 Shouldn't it be [`Arrangement`](https://developer.android.com/reference/kotlin/androidx/compose/foundation/layout/Arrangement)?

You are right, we should use `Arrangement` to align different attributions, but when one attribution is split between 2 lines, we need to use `TextAlign`, so ideally we can keep both. It's fixed.

> Did you test that your calculation works fine both when the popup expands to the left _and_ when the popup expands to the right?

Yes, I tested RTL and LTR as well, and everything seems to be working fine.

westnordost (approved):

PR #315, opened 2025-05-06
Update map before sending map attach signal

michalgwo:
Currently `CameraState.awaitInitialized()` doesn't work as expected. When it receives `mapAttachSignal`, the `map` variable is still null, so calling `requireMap()` after `CameraState.awaitInitialized()` still crashes. It happens because the `mapAttachSignal` is send before setting `map` variable. This PR fixes this problem.

### DIFF
--- a/lib/maplibre-compose/src/commonMain/kotlin/dev/sargunv/maplibrecompose/compose/CameraState.kt
+++ b/lib/maplibre-compose/src/commonMain/kotlin/dev/sargunv/maplibrecompose/compose/CameraState.kt
@@ -33,11 +33,13 @@ public fun rememberCameraState(firstPosition: CameraPosition = CameraPosition())
 public class CameraState internal constructor(firstPosition: CameraPosition) {
   internal var map: MaplibreMap? = null
     set(map) {
-      if (map != null && map !== field) {
+      val previousField = field
+      field = map
+
+      if (map != null && map !== previousField) {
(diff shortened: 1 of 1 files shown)


lib/maplibre-compose/src/commonMain/kotlin/dev/sargunv/maplibrecompose/compose/CameraState.kt
```diff
@@ -33,11 +33,13 @@ public fun rememberCameraState(firstPosition: CameraPosition = CameraPosition())
...
-      if (map != null && map !== field) {
+      val previousField = field
+      field = map
+
+      if (map != null && map !== previousField) {
         (map as StandardMaplibreMap).setCameraPosition(position)
         mapAttachSignal.trySend(map)
       }
-      field = map
     }
```
sargunv, line 43: would setting `field = map` inside the `if` also fix the issue?

sargunv (approved):
lgtm, with one comment

lib/maplibre-compose/src/commonMain/kotlin/dev/sargunv/maplibrecompose/compose/CameraState.kt
michalgwo: If we do that, we won't be able to set the map to null

PR #269, opened 2025-02-02
Fix crash when changing styles quickly by wrapping styles in SafeStyle

sargunv:
Attempt to fix #252 with an alternative approach to #265.

PR #316, opened 2025-05-08
[iOS] Fixed crash when setting geojson while changing style

michalgwo:
When `GeoJsonSource.setUri()` or `GeoJsonSource.setData()` is called while changing styles, on iOS, the exception `MLNInvalidStyleSourceException` is thrown by the validation in Maplibre Native:

https://github.com/maplibre/maplibre-native/blob/92cf551ec1d80c71d9ebe29a737dea601fd9d9c1/platform/darwin/src/MLNShapeSource.mm#L221-L225

On Android, in the same case, Maplibre Native just skips the execution:

https://github.com/maplibre/maplibre-native/blob/92cf551ec1d80c71d9ebe29a737dea601fd9d9c1/platform/android/MapLibreAndroid/src/main/java/org/maplibre/android/style/sources/GeoJsonSource.kt#L236-L242

This PR prevents a crash on iOS and makes the behaviour on iOS consistent with Android.

I'm aware that the solution is not very elegant, but I couldn't find a better way to do that.

### DIFF
--- a/lib/maplibre-compose/build.gradle.kts
+++ b/lib/maplibre-compose/build.gradle.kts
@@ -59,7 +59,10 @@ kotlin {
   cocoapods {
     noPodspec()
     ios.deploymentTarget = project.properties["iosDeploymentTarget"]!!.toString()
-    pod("MapLibre", libs.versions.maplibre.ios.get())
+    pod("MapLibre") {
+      version = libs.versions.maplibre.ios.get()
+      extraOpts = listOf("-Xforeign-exception-mode", "objc-wrap")
+    }
   }
 
   sourceSets {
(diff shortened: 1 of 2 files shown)


sargunv:
Can we check if the source is valid before setting, or set some sort of `detached` flag upon invalidation so we don't have to do the try/catch with string matching?

michalgwo:
I was looking for a function to check if the source is valid, but I couldn't find any. Another way to fix this I can think of is like in #269, but try/catch seems cleaner to me.

michalgwo:
Now, when I think about this again, maybe it's better to implement this as in #269, this would probably need to be made on top of #269

PR #339, opened 2025-05-17
fix: recognize long press correctly on delay, not on release, on ios

### LINKED ISSUE #313: [iOS] Long Press is recognised only after the gesture ends
### What happened?

**Current behaviour:**
When I long-press on the map, then release the finger, the gesture callback (code below) is executed twice at the same time. First with status `UIGestureRecognizerStateBegan`, second one with `UIGestureRecognizerStateEnded`.
https://github.com/sargunv/maplibre-compose/blob/5f043da445ea79459f656b696d57e9dbfd64e9e4/lib/maplibre-compose/src/iosMain/kotlin/dev/sargunv/maplibrecompose/core/IosMap.kt#L109-L113 
This ultimately leads to executing `onMapLongClick` of `MaplibreMap` after the finger is released.

**Expected behaviour:**
When I long-press on the map, the gesture callback should be executed after reaching `minimumPressDuration` before releasing the finger with state  `UIGestureRecognizerStateBegan`. Then, after releasing the finger, it should be called again with state `UIGestureRecognizerStateEnded`. This will ultimately execute  `onMapLongClick` of `MaplibreMap` once the long click begins.

### Affected Platforms

iOS

### Platform Version

18.1

### Library Version

0.6.0

### Sample Code

From my investigation, it looks like the culprit is here:

https://github.com/sargunv/maplibre-compose/blob/5f043da445ea79459f656b696d57e9dbfd64e9e4/lib/maplibre-compose/src/iosMain/kotlin/dev/sargunv/maplibrecompose/core/IosMap.kt#L242-L246

After commenting out this code, the gesture recognition seems to work fine, but as I couldn't figure out why this code is here, I don't know if removing this code breaks anything else in the code.

sargunv:
Fixes #313

### DIFF
--- a/lib/maplibre-compose/src/iosMain/kotlin/dev/sargunv/maplibrecompose/core/IosMap.kt
+++ b/lib/maplibre-compose/src/iosMain/kotlin/dev/sargunv/maplibrecompose/core/IosMap.kt
@@ -105,7 +105,7 @@ internal class IosMap(
         val point = locationInView(this@IosMap.mapView).toDpOffset()
         callbacks.onClick(this@IosMap, positionFromScreenLocation(point), point)
       },
-      Gesture(UILongPressGestureRecognizer()) {
+      Gesture(UILongPressGestureRecognizer(), isCooperative = false) {
         if (state != UIGestureRecognizerStateBegan) return@Gesture
         val point = locationInView(this@IosMap.mapView).toDpOffset()
         callbacks.onLongClick(this@IosMap, positionFromScreenLocation(point), point)


sargunv:
* **#339** <a href="https://app.graphite.dev/github/pr/sargunv/maplibre-compose/339?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/sargunv/maplibre-compose/339?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #346, opened 2025-05-21
Attribution button stretch fix

michalgwo:
## Description

Marges into #293

Fixes https://github.com/maplibre/maplibre-compose/pull/293#issuecomment-2888529761

It wasn't a compose bug as I initially thought, and it was happening both on Android and iOS. It occurred on iOS, because of a different screen size. It was actually a rare case when the popup's width was larger than the screen, but not large enough to properly break line. In short. Because we were setting the width which was calculated in `SuperimposingPopupPositionProvider` based on the content size, it couldn't properly update it.

This PR fixes this.

There is a small issue with this that I couldn't figure out how to fix, but it's not a big deal, I guess. When opening a map, sometimes for a split second, you can see a change in the popup from 1 line to 2 lines.

https://github.com/user-attachments/assets/32af3036-60dd-46f7-8f22-2e8d71b41d1f

It happens in the same case as above (again, a rare case, but you will see it when you open the demo app on iPhone 16 Pro Simulator). This time it happens only on iOS.

## Test plan

Ideally check it on different screen sizes

## Checklist

This pull request primarily:

- [ ] Updates documentation.
- [x] Fixes a bug. (please link to the issue)
- [ ] Adds a feature. (please link to the issue or discussion)

- [ ] Yes (please describe)
- [x] No

- [x] Yes
- [ ] No

- [x] Android (describe the device and OS version)
- [x] iOS (describe the device and OS version)
- [ ] Desktop (describe the device and OS version)
- [ ] Web (describe the browser version)

### DIFF
--- a/lib/maplibre-compose-material3/src/commonMain/kotlin/dev/sargunv/maplibrecompose/material3/controls/AttributionButton.kt
+++ b/lib/maplibre-compose-material3/src/commonMain/kotlin/dev/sargunv/maplibrecompose/material3/controls/AttributionButton.kt
@@ -14,6 +14,7 @@ import androidx.compose.foundation.layout.Spacer
 import androidx.compose.foundation.layout.absolutePadding
 import androidx.compose.foundation.layout.padding
 import androidx.compose.foundation.layout.width
+import androidx.compose.foundation.layout.widthIn
 import androidx.compose.foundation.shape.RoundedCornerShape
 import androidx.compose.material3.Icon
 import androidx.compose.material3.IconButton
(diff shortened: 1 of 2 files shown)


sargunv:
> When opening a map, sometimes for a split second, you can see a change in the popup from 1 line to 2 lines.

I slowed down the animation; looks like that one line issue is for the full duration of the animation, and it switches to two lines when the animation is done.

https://github.com/user-attachments/assets/3fca7ceb-6866-4a2c-9b71-f8ae03084c7e

sargunv:
And yeah it seems to work properly on Android:

[Screen_recording_20250521_225531.webm](https://github.com/user-attachments/assets/49353883-a85f-4544-81b4-b50a41f32421)

michalgwo:
> I slowed down the animation; looks like that one line issue is for the full duration of the animation, and it switches to two lines when the animation is done.

`maxPopupWidth` is calculated based on `anchorBounds` (anchor is `InfoButton`). It looks like it can't get the correct `maxPopupWidth` untill the animation is finished as `InfoButton` is not on the final position yet.

westnordost:
It's cool that you looked into it, and as you and sargunv are already I'll excuse myself of reviewing this etc. since I still don't have an iOS dev set up 😅. (Just writing this in case you are waiting for a review from me or something).

One comment though: If the behavior is different on iOS and Android, wouldn't it still be a bug in MapLibre Compose?  (Additional to maybe a bug in our code) After all, iOS support is still in beta, so it is still somewhat expected that there are some (small) issues remaining.

westnordost:
I understand that the remaining issue is that during animation to the view, the popup is not laid out correctly, which might be a limitation of compose.

I do wonder, if we don't use a custom `SuperimposingPopupPositionProvider`, would this still be a problem? If we used the default, and it is not a problem anymore, what is the magic sauce the Compose people are using for the calculation?

If it turns out to be a limitation of compose, maybe it is possible to detect when the animation finished only only **then** display the popup in the first place?

Issue #236, opened 2025-01-08
Crash on using clusterProperties due to adding duplicate source


westnordost:
While trying out how the clusterProperties work to properly document it, I noticed that it just crashes if one attempts to use it.

When I add in `ClusterPointsDemo` 

```kt
clusterProperties = mapOf("range" to GeoJsonOptions.ClusterProperty("+", feature.get("current_range_meters")))
```

(and `"current_range_meters" to (bike["current_range_meters"] ?: JsonNull),` in `readGbfsData` to actually have that data in the geojson)

the Android app will just crash in native code.

Of course, MapLibre should not crash, but maybe it crashes because the API is not used in the expected manner?

sargunv:
Did some debugging with @louwers today. We tried using the `android-sdk-debug` package instead of the release `android-sdk`, and saw a regular exception (adding a source twice) instead of a segfault.

Segfault investigation will proceed in https://github.com/maplibre/maplibre-native/issues/3493, but I dug into the source added twice, and it seems Compose is recomposing and recreating the Source when we pass a value for cluster properties. It's unclear why. Additionally, the `onDispose` hook for our source, which we rely on to remove the source from the style when it's removed from the composition, is being called only after the recreated version is added.

I think we need to either:
* understand exactly what guarantees Compose makes about recomposition and `onDispose` call order, and use that understanding to ensure we always remove a source before adding it
* or, internally manage source ids with an incrementing int like we used to, so it's not a problem to add the updated source before removing the old one

I prototyped the second approach and it does resolve the issue. I think I need to spend some time digging through our old discussions of why we should / should not internally manage IDs, as there was probably some good reason we stopped doing that.

cc @westnordost in case you remember anything

sargunv:
Found an old discussion about the composable style API and copied it [to a Discussion here](https://github.com/maplibre/maplibre-compose/discussions/400) to preserve.

westnordost:
The only thing I remember is that I was giving feedback on the API from the user point of view. If an (ideal) API is not possible due to some internal constraints, I guess this is how it is.

(Compose should recompose when any of the data changes, claro. But note that it will also recompose if actually nothing changed but any of the parameters passed are not immutable or marked as @Immutable. (E.g. if any `List<>` is passed.) Then only during recomposition, it may be decide that e.g. the list didn't change after all or whatever)

PR #365, opened 2025-05-26
Document and fix clusterproperties interface

sargunv:
## Description

Supersedes #237. Partial fix for #236.

I managed to get this working on iOS, but it still crashes on Android with a segfault. Reported upstream, https://github.com/maplibre/maplibre-native/issues/3493. Interestingly, it doesn't crash if the mapper is a literal, like `const(1)`. In that case, it seems to work on Android too, so I think we've got the usage right, it's just broken upstream.

## Test plan

Take the example from the docs, and use it in the cluster properties demo (instead of point count).


sargunv:
* **#365** <a href="https://app.graphite.dev/github/pr/maplibre/maplibre-compose/365?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/maplibre/maplibre-compose/365?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #373, opened 2025-05-27
feat: support drawing Painter as SDF

### LINKED ISSUE #202: Support SDF conversion during Painter rasterization

sargunv:
## Description

Resolves #202, using implementation from StreetComplete. See https://github.com/maplibre/maplibre-compose/pull/197#issuecomment-2563675917.

Tested and working on Android and iOS. Can't test on JS and Desktop yet, as we don't have style manipulation there yet.

### DIFF
--- a/demo-app/src/commonMain/kotlin/dev/sargunv/maplibrecompose/demoapp/demos/MarkersDemo.kt
+++ b/demo-app/src/commonMain/kotlin/dev/sargunv/maplibrecompose/demoapp/demos/MarkersDemo.kt
@@ -12,7 +12,6 @@ import androidx.compose.runtime.mutableStateOf
 import androidx.compose.runtime.remember
 import androidx.compose.runtime.setValue
 import androidx.compose.ui.Modifier
-import androidx.compose.ui.graphics.Color
 import androidx.compose.ui.unit.em
 import dev.sargunv.maplibrecompose.compose.ClickResult
 import dev.sargunv.maplibrecompose.compose.MaplibreMap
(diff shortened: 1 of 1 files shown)


lib/maplibre-compose/src/commonMain/kotlin/dev/sargunv/maplibrecompose/core/util/ImageBitmap.kt
```diff
@@ -0,0 +1,5 @@
+package dev.sargunv.maplibrecompose.core.util
+
+import androidx.compose.ui.graphics.ImageBitmap
+
+internal expect fun IntArray.toImageBitmap(width: Int, height: Int): ImageBitmap
```
westnordost, line 5: Hmm, why does this need separate implementation for each platform? (Also, why do I only see implementation for JS, iOS, desktop? Where's the implementation for JVM?)

There's also [ImageBitmap.toPixelMap](https://developer.android.com/reference/kotlin/androidx/compose/ui/graphics/ImageBitmap#(androidx.compose.ui.graphics.ImageBitmap).toPixelMap(kotlin.Int,kotlin.Int,kotlin.Int,kotlin.Int,kotlin.IntArray,kotlin.Int,kotlin.Int), maybe it is more convenient to work with that and doesn't require writing platform-specific code to convert it back to `ImageBitmap`?

sargunv: Desktop is JVM.

I couldn't find any other way to set the pixels of the bitmap (other than a DrawScope or something). PixelMap also appears to be read only, unless I'm missing something?

On Android we use an Android Bitmap, on the other platforms we use a Skia Bitmap. We could I suppose create a new "non-android" source set and share the Skia implementation, but it didn't seem worth it to figure that out just for one function.

westnordost: hm, I see. And on Android it is not possible to do the same as for the other implementations (so that there is only one implementation for all)?

I wonder if this would be worth a feature request at Jetpack Compose (Multiplatform).

PR #386, opened 2025-05-28
implement offline region manager for android and ios

### LINKED ISSUE #22: Support managing offline map data
It would be nice for the consumer to be able to manage downloaded map data. At a minimum, this means clearing the cache on demand, but perhaps up to actively saving certain regions of the world?

Unsure to what extent the native SDKs support this, but if they do, we should to the same extent.

sargunv:
## Description

Resolves #22

## Test plan

See new offline region demo


demo-app/src/maplibreNativeMain/kotlin/dev/sargunv/maplibrecompose/demoapp/demos/OfflineDemo.kt
```diff
@@ -0,0 +1,108 @@
...
+                  )
+                  .setDownloadState(DownloadState.Active)
+              }
+            },
+            enabled = cameraState.position.zoom >= 10.0,
+          ) {
+            Text("Save region")
+          }
+
+          Column(modifier = Modifier.verticalScroll(rememberScrollState())) {
```
sargunv, line 79: The demo is currently quite ugly, but I'd rather polish it up in a future PR (thinking some sort of bottom sheet containing controls and list, draggable up over the map or down out of the way)

lib/maplibre-compose/src/iosMain/kotlin/dev/sargunv/maplibrecompose/compose/offline/OfflineRegionManager.kt
```diff
@@ -0,0 +1,115 @@
...
+
+  private val regionsState = mutableStateOf(emptySet<OfflineRegion>())
+
+  override val regions
+    get() = regionsState.value
+
+  // TODO observe progress notifications and update region states:
+  // https://maplibre.org/maplibre-native/ios/latest/documentation/maplibre/mlnofflinepackprogresschangednotification
+  // https://maplibre.org/maplibre-native/ios/latest/documentation/maplibre/mlnofflinepackmaximummapboxtilesreachednotification
+  // https://maplibre.org/maplibre-native/ios/latest/documentation/maplibre/mlnofflinepackerrornotification
```
sargunv, line 34: Just this TODO left, then it's ready for review

lib/maplibre-compose/src/iosMain/kotlin/dev/sargunv/maplibrecompose/compose/offline/OfflineRegionManager.kt
```diff
@@ -0,0 +1,115 @@
...
+    keyPath: String?,
+    ofObject: Any?,
+    change: Map<Any?, *>?,
+    context: kotlinx.cinterop.CPointer<out kotlinx.cinterop.CPointed>?,
+  ) {
+    if (
+      ofObject == impl &&
+        keyPath == "packs" &&
+        change != null &&
+        change.contains(NSKeyValueChangeSetting)
```
sargunv, line 50: Note to self: I suspect this is wrong; gotta read up on https://developer.apple.com/library/archive/documentation/Cocoa/Conceptual/KeyValueObserving/Articles/KVOBasics.html

lib/maplibre-compose/src/maplibreNativeMain/kotlin/dev/sargunv/maplibrecompose/compose/offline/OfflineRegionManager.kt
```diff
@@ -0,0 +1,26 @@
...
+
+  public suspend fun delete(region: OfflineRegion)
+
+  public suspend fun invalidate(region: OfflineRegion)
+
+  public suspend fun invalidateAmbientCache()
+
+  public suspend fun clearAmbientCache()
+
+  public suspend fun setMaximumAmbientCacheSize(size: Long)
```
westnordost, line 23: This would probably also need some explanation what an "ambient cache" is. First thing that comes up in a google search is MapBox documentation:

> When a map is loaded and viewed on a device, the tiles are placed in the device's ambient cache. The ambient cache stores tiles that have already been visited by an end user so that the next time the user returns, the map loads nearly instantaneously.

lib/maplibre-compose/src/maplibreNativeMain/kotlin/dev/sargunv/maplibrecompose/compose/offline/OfflineRegionManager.kt
```diff
@@ -0,0 +1,26 @@
...
+
+  public fun setOfflineMapboxTileCountLimit(limit: Long)
```
westnordost, line 25:
```suggestion
  public fun setOfflineTileCountLimit(limit: Long)
```

I know, the standard format of vector tiles is **M**apBox **V**ector **T**iles, but the file format shouldn't matter, right? After all, MapLibre can also handle raster tiles.

lib/maplibre-compose/src/maplibreNativeMain/kotlin/dev/sargunv/maplibrecompose/compose/offline/OfflineRegionManager.kt
westnordost, line 1: Many methods in `OfflineRegionManager` are `suspend`ing. That's probably because MapLibre has callbacks to inform the caller about that it has been finished. But are they cancelable? I keep forgetting - what if the method is called in for example the UI coroutine scope, and that scope is cancelled (because the app is sent to the background or whatever). What happens with the execution of the method? Is it canceled (Exception thrown)? Or does it just not report back to the now cancelled UI scope?

lib/maplibre-compose/src/maplibreNativeMain/kotlin/dev/sargunv/maplibrecompose/compose/offline/OfflineRegion.kt
westnordost, line 1: From the name, I expected this class to represent an offline region. But reading the interface, it looks like it is a class to control and observe the download of an offline region.

lib/maplibre-compose/src/maplibreNativeMain/kotlin/dev/sargunv/maplibrecompose/compose/offline/DownloadProgress.kt
```diff
@@ -0,0 +1,19 @@
+package dev.sargunv.maplibrecompose.compose.offline
+
+public sealed interface DownloadProgress {
+  public data object Unknown : DownloadProgress
+
+  public data class Normal(
```
westnordost, line 6:
```suggestion
  public data class State(
```
? `DownloadProgress.Normal` sounds odd, IMO.

lib/maplibre-compose/src/maplibreNativeMain/kotlin/dev/sargunv/maplibrecompose/compose/offline/DownloadProgress.kt
```diff
@@ -0,0 +1,19 @@
...
+    val completedResourceCount: Long,
+    val completedResourceBytes: Long,
+    val completedTileCount: Long,
+    val completedTileBytes: Long,
+    val downloadState: DownloadState,
+    val isRequiredResourceCountPrecise: Boolean,
+    val requiredResourceCount: Long,
```
westnordost, line 13: I wonder why there is no `requiredTileCount`. `resources` probably refer to fonts, icons and such?

lib/maplibre-compose/src/iosMain/kotlin/dev/sargunv/maplibrecompose/core/util/util.kt
westnordost, line 1: This is a good case for adding a unit test: Convert to and from and see if the data remains identical. Also for testing edge cases (e.g. zero-length).

westnordost (commented):
So far, I've only looked at the interface

lib/maplibre-compose/src/maplibreNativeMain/kotlin/dev/sargunv/maplibrecompose/compose/offline/OfflineRegionManager.kt
sargunv: It's what the SDK calls it. I'm honestly unsure if the limit applies to all sources or if it's some vestigial remnant of Mapbox 

But yeah, best to rename it in our API I think

lib/maplibre-compose/src/maplibreNativeMain/kotlin/dev/sargunv/maplibrecompose/compose/offline/DownloadProgress.kt
sargunv: How about "Healthy"? As opposed to unknown, error, etc

lib/maplibre-compose/src/maplibreNativeMain/kotlin/dev/sargunv/maplibrecompose/compose/offline/DownloadProgress.kt
sargunv: I'm unsure of the difference between tile and resource here. These are 1:1 with what the MLN provides

demo-app/src/maplibreNativeMain/kotlin/dev/sargunv/maplibrecompose/demoapp/demos/OfflineDemo.kt
```diff
@@ -0,0 +1,92 @@
...
+            cameraState = cameraState,
+            styleState = styleState,
+            ornamentSettings = DemoOrnamentSettings(),
+          )
+          DemoMapControls(cameraState, styleState)
+        }
+
+        Column(modifier = Modifier.weight(0.5f)) {
+          val coroutineScope = rememberCoroutineScope()
+          val offlineManager = rememberOfflineRegionManager()
```
westnordost, line 52: I thought for a moment, "uh wait, how to access the offline region manager?", I didn't see anything in the interface. But then, I saw this.

I think this is the wrong location. I think `rememberOfflineRegionManager` doesn't make sense because it doesn't really make sense to do this in the composition.

You mentioned before that managing offline regions has nothing to do really with the UI, and I think you are right. Such things should not happen in the UI code but would typically happen in some ViewModel or Controller. I.e. on click, the button would call e.g. `viewModel.saveForOffline(bbox)` and then the ViewModel would (launch a coroutine outside of the UI scope) handle it imperatively, separately from the composition. And the ViewModel must have the means to access an OfflineRegionManager.

I guess, by the way, this is equally true for the MapSnapshotter.

lib/maplibre-compose/src/maplibreNativeMain/kotlin/dev/sargunv/maplibrecompose/compose/offline/DownloadProgress.kt
sargunv: If I had to guess, I'd say resource count is tile count plus other resources

demo-app/src/maplibreNativeMain/kotlin/dev/sargunv/maplibrecompose/demoapp/demos/OfflineDemo.kt
sargunv: On Android, we need a Context. I get it with LocalContext.current. Can we get that outside the composition, without the user passing it in (and therefore losing the common API)

lib/maplibre-compose/src/maplibreNativeMain/kotlin/dev/sargunv/maplibrecompose/compose/offline/DownloadProgress.kt
```diff
@@ -0,0 +1,19 @@
...
+  ) : DownloadProgress
+
+  public data class Error(val reason: String, val message: String) : DownloadProgress
+
+  public data class TileLimitExceeded(val limit: Long) : DownloadProgress
```
westnordost, line 18: Wouldn't this be an error? Where's the difference between an Error (with reason="Tile limit exceeded") and this one?

sargunv: I think the download can proceed once other regions are deleted

The MLN SDK differentiates between errors and this state

lib/maplibre-compose/src/maplibreNativeMain/kotlin/dev/sargunv/maplibrecompose/compose/offline/DownloadProgress.kt
westnordost: hm...

demo-app/src/maplibreNativeMain/kotlin/dev/sargunv/maplibrecompose/demoapp/demos/OfflineDemo.kt
westnordost: (And I don't know whether there's an easier method, i.e. asking compose outside of the composition for it or something)

lib/maplibre-compose/src/maplibreNativeMain/kotlin/dev/sargunv/maplibrecompose/compose/offline/OfflineRegionManager.kt
sargunv: Yup, it's the automatically cached offline tiles

lib/maplibre-compose/src/maplibreNativeMain/kotlin/dev/sargunv/maplibrecompose/compose/offline/OfflineRegionManager.kt
sargunv: I tried it by adding a `delay()`. It's not cancellable (because I used `suspendCoroutine` instead of `suspendCancellableCoroutine`). When I back out of that screen, or dismiss the app, no exception is thrown (so it seems we just don't report back).

lib/maplibre-compose/src/maplibreNativeMain/kotlin/dev/sargunv/maplibrecompose/compose/offline/OfflineRegion.kt
sargunv: In MLN, it's OfflineRegion (Android) or OfflinePack (iOS). It represents a particular region of the map that was downloaded, contains its progress, and has methods to mutate it. It's what gets created or deleted using the methods in `OfflineRegionManager`.

demo-app/src/maplibreNativeMain/kotlin/dev/sargunv/maplibrecompose/demoapp/demos/OfflineDemo.kt
sargunv: Okay, I've done that. So the common API is dependent on the Composition, but there's a platform-specific API on both iOS and Android that's not. The one on Android takes `Context` as a parameter.

lib/maplibre-compose/src/iosMain/kotlin/dev/sargunv/maplibrecompose/core/util/util.kt
sargunv: done

demo-app/src/maplibreNativeMain/kotlin/dev/sargunv/maplibrecompose/demoapp/demos/OfflineDemo.kt
sargunv: Nvm, went ahead and redid the demo in this PR

demo-app/src/maplibreNativeMain/kotlin/dev/sargunv/maplibrecompose/demoapp/demos/OfflineDemo.kt
westnordost: > I haven't worked much with view models, but my understanding is they're typically instantiated within the composition

Depends on the dependency injection framework one uses, if any. The example you have above wouldn't work with Koin, IIRC. In Koin it would look like

```kotlin
@Composable
fun Example() {
  val viewModel = koinViewModel<ExampleViewModel>()
  // ...
}
```

Even if somehow possible, it feels wrong to me if the view would pass any data to the view model. The view model holds the data and the logic, and the flow of that should be unidirectional. I.e. the view subscribes to updates of the view model, the view model in turn knows nothing about its subscribers / the view.

Also, independent of the view model, there might also be a component deeper in the app's code, independent of view model or view, that's managing offline regions. You wouldn't want to have to pass the context from the view to the view model and then from the view model to that component. (Because that component should be independent of the view completely.)

westnordost (commented):
Impressive demo! Cool that you added all those tests.

demo-app/src/maplibreNativeMain/kotlin/dev/sargunv/maplibrecompose/demoapp/demos/OfflineDemo.kt
westnordost: So, anyway, to get it across better, here's my set up:

Somewhere in the UI, there is a button labelled "download data here".

On press, the view model is told to download the data at the shown bbox.

The view model tells a class named `DownloadController` to download the area.

The `DownloadController` enqueues a a `Worker` (Android-only construct) to work on the download in the background.

The `Worker` owns a reference to a class named `Downloader`. When told to start working by the Android system, it just tells the `Downloader` to do its thing (for the given bounding box).

Now, the `Downloader` downloads the OpenStreetMap data, other data for the region and then also the tile data for MapLibre.

In other words, this couldn't be further removed from the view.

sargunv:
I definitely want to start tackling automated tests for the library soon, especially with all the recent iOS crashes in specific scenarios and the complex logic in the attribution button alignment that's tedious to test

The snapshotter, when finished, could help us test style stuff better too perhaps.

lib/maplibre-compose/src/maplibreNativeMain/kotlin/dev/sargunv/maplibrecompose/compose/offline/OfflineTilePack.kt
```diff
@@ -0,0 +1,25 @@
...
+  public val metadata: ByteArray?
+
+  /** The pack's current progress. */
+  public val progress: DownloadProgress
+
+  /** Resume downloading if the pack is paused. */
+  public fun resume()
+
+  /** Pause downloading if the pack is downloading. */
+  public fun suspend()
```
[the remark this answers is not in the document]
sargunv: Ahh yeah, I missed this when renaming to "paused"

demo-app/src/maplibreNativeMain/kotlin/dev/sargunv/maplibrecompose/demoapp/demos/OfflineDemo.kt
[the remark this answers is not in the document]
sargunv: Maybe parts of this are suitable for the library itself? Like the ListItem in the material3 module, or the region to geometry, or even the source, in the core module. 

I imagine many apps will have a list of downloaded regions, and many apps will want to highlight downloaded areas on the map somehow.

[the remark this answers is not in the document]
sargunv: (Also, this demo was largely written by Junie, the Jetbrains AI coding agent thing lol)

lib/maplibre-compose/src/maplibreNativeMain/kotlin/dev/sargunv/maplibrecompose/compose/offline/OfflineTilePack.kt
```diff
@@ -0,0 +1,25 @@
+package dev.sargunv.maplibrecompose.compose.offline
+
+/** Represents a collection of resources necessary for viewing a region offline. */
+public expect class OfflineTilePack {
```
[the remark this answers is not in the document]
sargunv: Yup exactly, it's similar to that pattern. In that case we wrap it additionally with Composables and only expose the composable API for now. 

I think someday that'll change, we'll want an optional imperative API available for cases where the Composable one is too restrictive. And then it'll be similar to this one.

demo-app/src/maplibreNativeMain/kotlin/dev/sargunv/maplibrecompose/demoapp/demos/OfflineDemo.kt
```diff
@@ -0,0 +1,322 @@
...
+        offlineManager.regions.forEach { pack ->
+          val def = pack.definition
+          feature(
+            geometry =
+              when (def) {
+                is TilePackDefinition.TilePyramid -> {
+                  val bounds = def.bounds
+                  polygon {
+                    ring {
+                      +bounds.southwest
```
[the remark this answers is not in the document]
sargunv: I generally prefer nesting over extracting functions when those functions are used only once, because I don't like hunting up and down a file or multiple files to understand the true implementation of a thing. 

Still, it's probably excessive here. Will cut it down a bit.

lib/maplibre-compose/src/maplibreNativeMain/kotlin/dev/sargunv/maplibrecompose/compose/offline/OfflineTilesManager.kt
```diff
@@ -0,0 +1,72 @@
...
+ * An offline tile pack represents a collection of resources needed to display a region offline,
+ * including map tiles, styles, and other assets. It allows you to selectively download regions of
+ * the map to be made available offline.
+ *
+ * The ambient cache is a temporary storage mechanism used to improve map loading performance and
+ * reduce network requests. It caches map tiles and other resources that the map renders, allowing
+ * them to be retrieved faster on subsequent map views or when zooming into previously viewed areas.
+ * The ambient cache is distinct from offline packs, which are used for persistent offline access.
+ */
+public interface OfflineTilesManager {
```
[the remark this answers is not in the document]
sargunv: added a `@throws` note, but without information on the circumstances (since I'm not sure). Would accept a PR to add that later if somebody investigates it

lib/maplibre-compose/src/maplibreNativeMain/kotlin/dev/sargunv/maplibrecompose/compose/offline/OfflineTilePack.kt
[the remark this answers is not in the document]
sargunv: Oh also, it's not hard for us to move `resume` and `pause` into the manager. I can def do that.

westnordost (commented):
Hmm, I am still unsure about the whole `getOfflineManager` stuff. I guess what I personally would need to do to ensure that `getOfflineManager(context)` is first called on the UI thread is to call it in the `onCreate` method of the Android `Application` class. Then, later, I can call it anywhere I like.

It is just kind of an awkward API. But so is the current MapLibre-native on Android API that requires one to call `MapLibre.getInstance(context)` anywhere on the UI thread once before using the MapLibre view anywhere in the app. And *then*, I still have to pass the context, even though `MapLibre` already got initialized... 🙄

Well, maybe it can't be helped 🤷

By the way, 

```kotlin
companion object {
  private var manager: AndroidOfflineManager? = null

  internal fun getInstance(context: Context): AndroidOfflineManager =
  manager ?: AndroidOfflineManager(context.applicationContext).also { manager = it }
}
```

is not threadsafe, i.e. it is possible that `AndroidOfflineManager` is constructed twice by two concurrent threads. But I think, this is harmless because in the end, both will use the same underlying `MLNOfflineManager` instance. (And if not, it is a race condition in MapLibre-native and should be fixed there)

sargunv:
Hmm, actually I think ending up with multiple OfflineManager instances is a bug, for the same reason that multiple OfflinePack instances for a native pack is a bug.

So we should make that thread safe.

sargunv:
> (Are you sure? OfflinePack contains data. OfflineManager doesn't contain data. )

I was thinking OfflineManager would be responsible for maintaining uniqueness of OfflinePack, but I actually implemented without that. So nope, OfflineManager can have duplicates now.

PR #406, opened 2025-06-04
wip: material map style

sargunv:


sargunv:
* **#406** <a href="https://app.graphite.dev/github/pr/maplibre/maplibre-compose/406?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/maplibre/maplibre-compose/406?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

lib/maplibre-compose-material3/src/commonMain/kotlin/dev/sargunv/maplibrecompose/material3/style/MaterialMap.kt
westnordost, line 1: Oh, as part of the material3 library, not just a demo?

I think it would be better as just a demo, for several reasons:
1. A style is always tied to a certain map tile provider. I.e. this `MaterialMapStyle` only works with openfreemap (OpenMapTiles?). But the library should be agnostic of style
2. It's somewhat of a toy style. Design-wise, creating a proper style is much more involved. Which either means
  a) this style will not really be used by anyone
  b) or, this will be an endless source of change requests to improve the style

sargunv: Not 100% sure if it'll be a demo or part of the library.

If its a part of the library, I'd want it to be:
* source-agnostic (like, some adapter interface with provided implementations for common open sources)
* relatively minimal, useful primarily as a base map for apps adding their own data and layers on top (so, more like Positron, less like Liberty)

For now, just experimenting.

Issue #72, opened 2024-12-01
Benchmark and/or optimize performance of updating GeoJson data


sargunv:
Continuation of https://github.com/sargunv/maplibre-compose/issues/42 specifically for GeoJsonSource updation.

When the user passes a URL to the GeoJsonSource, the native SDK is fully responsible for acquiring the data.

But when the user passes some Spatial-K GeoJson objects, our library must convert those to something the native SDK understands. We currently do this by serializing these objects to JSON, and passing that JSON into the native SDK. This probably has significant overhead. The Android and iOS SDKs both have native classes representing GeoJson; is it more performance to manually convert to those? Can Spatial-K serialization be optimized? How much of the work is done in serialization, and how much in the native SDK deserialization?

There may be some use cases (see camera follow demo) where geojson is updated frequently, perhaps every frame. There may be other use cases (see clustering demo) where large geojson is inserted. This demo currently hangs for a moment upon loading.

We should benchmark performance of updating geojson data programatically, and find optimizations.

westnordost:
If using the platform specific data classes for GeoJSON data instead of GeoJson strings turn out to be better for performance, this might lead to replacing Spatial-K with an own actual/expect wrapper around each what the iOS and Android SDK has to offer in terms of GeoJson data classes (unless Spatial-K is already exactly that).

However, keep in mind, before going with this option, that when eventually the C API is released, this would again necessitate to re-do this part (maybe using Spatial-K again, maybe something else).

What I am reading from your comment, it sounds like the current approach is the approach that is most future proof way in this regard. (Just to keep this in mind.)

michalgwo:
I did some investigation.

### 1. Benchmarking
It looks like serialization takes a significant amount of time.

The code I ran:
``` kotlin
val time = measureTime {
    json = feature.json()
}
```

Size of GeoJson: 1,2MB
Time on Xiaomi Redmi Note 13: ~800ms
Time on iPhone SE 2022: ~230ms (what a huge difference :open_mouth:)

### 2. Solutions
- We can skip serializing at least on one platform (preferably Android, as even low-end iPhone is much faster) by implementing our own actual/expect wrapper with typealias, which will perform the same as native on one platform, but we will still need to convert it on the other platform.
[...]

PR #411, opened 2025-06-06
Setting serialized GeoJson as a source

michalgwo:
## Description

This PR allows the user to pass serialized GeoJson to `rememberGeoJsonSource`.

Use cases:
- Setting `GeoJsonSource` when the user has only serialized GeoJson
- GeoJson can be serialized before passing to the `rememberGeoJsonSource`. It gives more control over the process of serialization. As it can take some time to serialize it, the user may want to serialize it in the background thread, cache the serialized string, and display a loading indicator during serialization. In the `ClusteredPointsDemo` I demonstrated serializing in the background thread (which eliminates UI freeze during the loading) and displaying a loading indicator. Caching doesn't make sense in this case, but it's possible now.

## Test plan

Run `ClusteredPointsDemo` and check if everything is smooth.

## Checklist

This pull request primarily:

- [x] Adds a feature. Partially related to #72

- [x] Yes (please describe)

Adding a new overload of `rememberGeoJsonSource`, created conflicting overloads, so I needed to wrap `Uri` into the data class, so now we need to do:
``` kotlin
rememberGeoJsonSource(id = "id", uri = Uri(Res.getUri(ROUTES_FILE)))
```
instead of
``` kotlin
rememberGeoJsonSource(id = "id", uri = Res.getUri(ROUTES_FILE))
```

- [x] Yes

- [x] Android 14 - Xiaomi Redmi Note 13
- [x] iOS 18.5 - iPhone SE 2022

### DIFF
--- a/lib/maplibre-compose/src/commonMain/kotlin/dev/sargunv/maplibrecompose/compose/source/rememberGeoJsonSource.kt
+++ b/lib/maplibre-compose/src/commonMain/kotlin/dev/sargunv/maplibrecompose/compose/source/rememberGeoJsonSource.kt
@@ -4,6 +4,7 @@ import androidx.compose.runtime.Composable
 import androidx.compose.runtime.key
 import dev.sargunv.maplibrecompose.core.source.GeoJsonOptions
 import dev.sargunv.maplibrecompose.core.source.GeoJsonSource
+import dev.sargunv.maplibrecompose.core.source.Uri
 import io.github.dellisd.spatialk.geojson.GeoJson
 
 /**
@@ -15,7 +16,7 @@ import io.github.dellisd.spatialk.geojson.GeoJson
 @Composable
 public fun rememberGeoJsonSource(
   id: String,
-  uri: String,
+  uri: Uri,
   options: GeoJsonOptions = GeoJsonOptions(),
 ): GeoJsonSource =
   key(id, options) {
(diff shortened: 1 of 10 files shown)


lib/maplibre-compose/src/commonMain/kotlin/dev/sargunv/maplibrecompose/compose/source/rememberGeoJsonSource.kt
```diff
@@ -42,3 +43,22 @@ public fun rememberGeoJsonSource(
...
+  id: String,
+  data: String,
+  options: GeoJsonOptions = GeoJsonOptions(),
+): GeoJsonSource =
+  key(id, options) {
+    rememberUserSource(
+      factory = { GeoJsonSource(id = id, data = data, options = options) },
+      update = { setData(data) },
+    )
+  }
```
sargunv, line 64: <p class='graphite__hidden'><i>[Re: lines +47 to +65]</i></p>
It's probably too easy to accidentally pass a URL string to this. Can we replace the `Uri` data class with a sealed interface of the possible geojson data sources? something like:

```kt
public sealed interface GeoJsonData {
  public data class Uri(...)
  public data class JsonString(...)
  public data class GeoJson(...)
}
```
<p class='graphite__hidden'>See this comment inline on <a href="https://app.graphite.dev/github/pr/maplibre/maplibre-compose/411?utm_source=unchanged-line-comment">Graphite</a>.</p>

michalgwo: Sure, I also renamed the `GeoJsonSource` constructor parameters, so we don't have `data.data` and it's now consistent with the parameters names of `setData()`

sargunv (approved):

PR #427, opened 2025-06-13
fix attribution text layout

sargunv:
## Description

Fixes incorrect layout direction, and also avoid breaking inside an attribution (we can still do separate lines for multiple attributions if they're long enough).

I'm not super happy with the scrolling solution, but anything else I tried looked wonky in certain cases.

## Test plan

CleanShot 2025-06-12 at 20 10 34@2x

image

image

### DIFF
--- a/lib/maplibre-compose-material3/src/commonMain/kotlin/dev/sargunv/maplibrecompose/material3/controls/AttributionButton.kt
+++ b/lib/maplibre-compose-material3/src/commonMain/kotlin/dev/sargunv/maplibrecompose/material3/controls/AttributionButton.kt
@@ -192,7 +192,9 @@ public fun ExpandingAttributionButton(
           exit = collapse(animationAlignment),
         ) {
           Box(Modifier.padding(start = 0.dp, end = 16.dp, top = 8.dp, bottom = 8.dp)) {
-            expandedContent(attributions)
+            CompositionLocalProvider(LocalLayoutDirection provides layoutDir) {
+              expandedContent(attributions)
+            }
           }
         }
       }
--- a/lib/maplibre-compose-material3/src/commonMain/kotlin/dev/sargunv/maplibrecompose/material3/controls/AttributionButton.kt
+++ b/lib/maplibre-compose-material3/src/commonMain/kotlin/dev/sargunv/maplibrecompose/material3/controls/AttributionButton.kt
@@ -8,11 +8,13 @@ import androidx.compose.animation.core.animateDpAsState
 import androidx.compose.animation.expandIn
 import androidx.compose.animation.shrinkOut
 import androidx.compose.foundation.BorderStroke
+import androidx.compose.foundation.horizontalScroll
 import androidx.compose.foundation.layout.Arrangement
 import androidx.compose.foundation.layout.Box
 import androidx.compose.foundation.layout.FlowRow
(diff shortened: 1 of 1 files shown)


westnordost (commented):
huh? Isn't it the job of the `FlowRow` to expand the area so that all the `Text`s fit in? (I.e. upstream bug if something is wrong)

westnordost:
> avoid breaking inside an attribution

Why would you do that? If it is long, why not break it? I wouldn't expect a horizontal scroll on a bullet point list either (but long bullet points being line-breaked)

sargunv:
To my eye, it ended up looking odd with a break because the surface expands to the full width of the screen, but the break happens much sooner and so leaves an odd empty space on the side. It's particular to the specific text and display size, but the OpenFreeMap attribution on my Z Fold and the phone triggered that odd looking case. 

It looked something like:

```kt
[OpenFreeMap (c) OpenMapTiles Data from         ]
[OpenStreetMap                                  ]
```

Where I'd prefer:

```kt
[OpenFreeMap (c) OpenMapTiles Data from]
[OpenStreetMap                         ]
```

or even, with the ability to edit the attribution string to insert non-breaking spaces:

```kt
[OpenFreeMap (c) OpenMapTiles]
[Data from OpenStreetMap     ]
```

sargunv:
Anyways, I'm not really happy with the horizontal scroll either, and if we can fix it to expand to the actual width of the text and no more, I'd be happy with removing maxLines here.

Still, it's easy for the user to override this behavior if they prefer the alternative.

westnordost:
But that's an unusually long map attribution text anyway. Legal-wise, `(c) OpenMapTiles (c) OpenStreetMap` should be fine, too. (I hate this requirement of OpenMapTiles to contain attribution to it, it's just a tile schema, for gods sake!)

Anyway, seems to me that the best solution would be to not use `FlowRow` at all, but instead concatenate the attribution HTMLs of all sources in a single annotated string. (Maybe add `'|'` as separators though?)

sargunv:
If you come up with some version that looks better on common devices with our example styles I'd be happy to change it again

I do personally prefer the look of not breaking within a single attribution, but it's easy enough for the user to customize this component so I'm not attached to that being provided by the library as long as what we do provide by default works well with the styles in the demo app.

westnordost:
I was about to make a PR for this, but realized that it changes nothing because that OpenFreeMap attribution is **one** attribution text. I.e. that attribution text still breaks in the way you don't want it to.

```kt
/**
 * A composable function that displays a collection of attribution links as a flow layout.
 *
 * @param attributions A list of HTML strings representing the attributions that need to be
 *   displayed as links. See: [dev.sargunv.maplibrecompose.core.source.Source.attributionHtml].
 * @param linkStyles Optional style for hyperlinks. Default is primary color and underlined.
 * @param attributionSpacing Which HTML string to use to separate the different attribution text(s)
 */
@Composable
public fun AttributionLinks(
  attributions: List<String>,
  linkStyles: TextLinkStyles? = AttributionButtonDefaults.linkStyles(),
  attributionSpacing: String = " ",
  modifier: Modifier = Modifier,
) {
  val text = htmlToAnnotatedString(attributions.joinToString(attributionSpacing), linkStyles)
  Text(text, modifier)
}
```

PR #444, opened 2025-06-17
Update CONTRIBUTING.md

sargunv:
updating contributing guide 

would be nice to have some architecture overview too, but that's something for later

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -32,8 +32,11 @@ As there's no stable LSP for Kotlin Multiplatform, you'll want to use either
 IntelliJ IDEA or Android Studio for developing MapLibre Compose. In addition to
 the IDE, you'll need some plugins:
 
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -1,28 +1,62 @@
...
 ```properties
-# Replace the paths with the actual paths on your machine
+# Replace the path with the actual path on your machine
 sdk.dir=/Users/username/Library/Android/sdk
 ```
 
-Follow
-[Jetbrains's instructions](https://www.jetbrains.com/help/kotlin-multiplatform-dev/multiplatform-setup.html#get-help)
-for setting up an environment for KMP development.
+As there's no stable LSP for Kotlin Multiplatform, you'll want to use either
```
westnordost, line 31: What's an LSP?

sargunv: https://en.m.wikipedia.org/wiki/Language_Server_Protocol

Used by editors like vscode, neovim, etc to provide language intelligence like an IDE.

westnordost: Ah

PR #456, opened 2025-06-19
add ComputedSource

sargunv:
## Checklist

This pull request primarily:

- [ ] Updates documentation.
- [ ] Fixes a bug. (please link to the issue)
- [ ] Adds a feature. (please link to the issue or discussion)

- [ ] Yes (please describe)
- [ ] No

- [ ] Yes
- [ ] No

- [ ] Android (describe the device and OS version)
- [ ] iOS (describe the device and OS version)
- [ ] Desktop (describe the device and OS version)
- [ ] Web (describe the browser version)


sargunv:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/maplibre/maplibre-compose/456?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>
> <a href="https://graphite.dev/docs/merge-pull-requests">Learn more</a>

* **#456** <a href="https://app.graphite.dev/github/pr/maplibre/maplibre-compose/456?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/maplibre/maplibre-compose/456?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#455** <a href="https://app.graphite.dev/github/pr/maplibre/maplibre-compose/455?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#454** <a href="https://app.graphite.dev/github/pr/maplibre/maplibre-compose/454?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

westnordost:
Cool, I didn't know this existed. This could be potentially very interesting for my project, too. In StreetComplete, I display a jungle of pins from a GeoJson source which I need to update very frequently. For performance reasons, what I do is to actually only add to the GeoJson whatever is currently in view and clear data that is outside the view (so that the GeoJson remains small).
A ComputedSource sounds like a proper solution for this kind of thing.

Unfortunately, I find the (Android) documentation for the custom geometry source somewhat lacking, which is why I cannot really comment on whether your implementation is correct. It might be worthwhile to either create a demo for testing or create a prototype of your train/scooter tracker app to test whether the API works the way you think it does.

For example, one is supposed to supply the feature collection via the GeometryTileProvider, but then, there are methods on the CustomGeometrySource like `setTileData` and `invalidateRegion` etc.

westnordost:
> For example, one is supposed to supply the feature collection via the GeometryTileProvider, but then, there are methods on the CustomGeometrySource like setTileData and invalidateRegion etc.

And I imagine this might be a bit problematic in the context of compose, where there's supposed to be somewhat of a unidirectional data source. With the architecture in the Android API, I gather that a data source both has to implement `GeometryTileProvider` and also kind of tell its parent (`CustomGeometrySource`) of data updates (i.e. call `invalidate`, and whatever `setTileData` is supposed to do).

sargunv:
I think calling those additional methods is optional, like if you need to update data outside of the normal tile load / unload as the map pans around?

But yeah, I'm not sure yet. I intend to try it out in a demo of some sort

westnordost:
Anyway, as this library is only wrapping iOS, Android APIs at the moment, maybe nothing to worry about. I am just thinking further in the sense that this kind of "tell source to update because data changed" sounds like something that should be done using the composition system, i.e. "automatic application of data updates".

sargunv:
> sounds like something that should be done using the composition system, i.e. "automatic application of data updates".

Yeah, we do exactly that for the geojson source right now.

> Maybe that's what volatile is about. 

We don't even support setting the `volatile` flag programmatically yet (I can't find the relevant API for iOS; unsure if it exists).

As we build experience with this API, we can refine it better for Compose

Issue #416, opened 2025-06-08
Expose map loaded state


michalgwo:
### Feature Description

When the internet connection is slow, loading the map can take some time. It would be nice to have the possibility to display information that it's loading (e.g. loading indicator), for that purpose we need to expose whether the map is loaded/rendered.

### Target Platforms

MapLibre Native - Android, MapLibre Native - iOS

### MapLibre Native/JS Support

Yes

### MapLibre Documentation Links

There are 2 listeners/delegates. I'm don't know what's the difference. While testing, they were called at the same time.
 
- Android: [OnDidFinishLoadingMapListener](https://maplibre.org/maplibre-native/android/api/-map-libre%20-native%20-android/org.maplibre.android.maps/-map-view/-on-did-finish-loading-map-listener/index.html) and [OnDidFinishRenderingMapListener](https://maplibre.org/maplibre-native/android/api/-map-libre%20-native%20-android/org.maplibre.android.maps/-map-view/-on-did-finish-rendering-map-listener/index.html)
- iOS: [mapViewDidFinishLoadingMap](https://maplibre.org/maplibre-native/ios/latest/documentation/maplibre/mlnmapviewdelegate/mapviewdidfinishloadingmap:/) and [mapViewDidFinishRenderingMap](https://maplibre.org/maplibre-native/ios/latest/documentation/maplibre/mlnmapviewdelegate/mapviewdidfinishrenderingmap:fullyrendered:)

### Implementation

Yes

### Development Platforms

Android

sargunv:
The "did finish loading map" event I think is when the various TileJSON sources have all finished loading for the first time

Not sure about "did finish rendering map".

I guess we'd want a new `MapState` for this kind of stuff?

sargunv:
I forget how the API works for getting the frame rate in the frame rate demo, but that's a good candidate for folding into MapState too

PR #470, opened 2025-06-20
Exposed finished and failed map loading functions

andreas-umbricht:
## Description

I came accross [this issue here](https://github.com/maplibre/maplibre-compose/issues/416). Until such a map state is being created (which I assume will take a while), I would still already be happy to have some feedback on wether the map has loaded or not. This can be refined later on or expanded with different kind of callbacks or just simply added to a map state.

## Test plan

I simply extended the demos and looked if the callbacks are effectively propagated to the client.

## Checklist

- Android: Samsung S21 Android 12, Pixel 8 Android 15
- iOS: iPhone 16 Pro iOS 18.3


sargunv (approved):
lgtm, just tweaked naming and kdoc

PR #471, opened 2025-06-20
Correct iconPadding type: It is a DpPaddingValue

westnordost:
It looks like this was an oversight before:

[`icon-padding`](https://maplibre.org/maplibre-style-spec/layers/#icon-padding) is not of type `number` (like [`text-padding`](https://maplibre.org/maplibre-style-spec/layers/#text-padding)) but of type [`padding`](https://maplibre.org/maplibre-style-spec/types/#padding) (like [`icon-text-fit-padding`](https://maplibre.org/maplibre-style-spec/layers/#icon-text-fit-padding)).

It is a breaking change, as the interface changes.

I have not tested the changes. (But I can say from usage in my project that the linked documentation is correct, it is a padding, not a single value).

Also, `PaddingValues.Absolute(0.dp)` is wrong in the sense that `left` is set to `0.dp` and the rest is set to the default. (Which incidentally is also `0.dp`, so this doesn't result in a bug)

### DIFF
--- a/lib/maplibre-compose-expressions/src/commonMain/kotlin/dev/sargunv/maplibrecompose/expressions/Defaults.kt
+++ b/lib/maplibre-compose-expressions/src/commonMain/kotlin/dev/sargunv/maplibrecompose/expressions/Defaults.kt
@@ -12,7 +12,10 @@ import dev.sargunv.maplibrecompose.expressions.value.ColorValue
 import dev.sargunv.maplibrecompose.expressions.value.ListValue
 import dev.sargunv.maplibrecompose.expressions.value.StringValue
 
-public val ZeroPadding: PaddingValues.Absolute = PaddingValues.Absolute(0.dp)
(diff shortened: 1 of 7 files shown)


lib/maplibre-compose-expressions/src/commonMain/kotlin/dev/sargunv/maplibrecompose/expressions/Defaults.kt
```diff
@@ -12,7 +12,10 @@ import dev.sargunv.maplibrecompose.expressions.value.ColorValue
 import dev.sargunv.maplibrecompose.expressions.value.ListValue
 import dev.sargunv.maplibrecompose.expressions.value.StringValue
 
-public val ZeroPadding: PaddingValues.Absolute = PaddingValues.Absolute(0.dp)
+public val ZeroPadding: PaddingValues.Absolute = PaddingValues.Absolute(0.dp, 0.dp, 0.dp, 0.dp)
+
+public val DefaultIconPadding: PaddingValues.Absolute =
```
sargunv, line 17: nice catch

sargunv (approved):

Issue #461, opened 2025-06-19
Rework demos


sargunv:
https://github.com/maplibre/maplibre-compose/pull/457#discussion_r2157433872

@westnordost:

> It would be much cooler to show an actual map semi-transparently overlayed over the normal map, maybe with a slider to control transparency. E.g. an old public domain map. Since this PR is already merged and the author of this library is from Seattle, maybe he wants to add a map from Seattle instead?
> 
> [commons.wikimedia.org/wiki/Category:Old_maps_of_Seattle](https://commons.wikimedia.org/wiki/Category:Old_maps_of_Seattle)
> 
> (A map not oriented towards north or slightly skewed would be extra cool to show the capabilities)

sargunv:
#476 merges all the demos onto the same map, so I think they should be reworked to be a bit less chaotic when on the map all together:

Rework demos

TODOs:

- [ ] ImageSource as historic map overlay
- [ ] Merge markers and animated Layers demo into a single "Amtrak" demo with routes and stations
- [ ] Make location demo more robust (fix bearing)
- [ ] Showcase Compose animations for layer properties on the location puck instead of Amtrak routes
- [ ] Simplify Lime bikeshare clustering demo visuals

sargunv:
Stack #965 reworks the demos, different from the checklist above but it does include the castelloplan demo

PR #497, opened 2025-07-10
auto assign source IDs

### LINKED ISSUE #236: Crash on using clusterProperties due to adding duplicate source

sargunv:
## Description

Fixes #236 and similar crashes when recreating sources by auto generating source IDs, which ensures we never have an ID conflict and we can remove sources after adding their replacements.

## Test plan

Re-enabled the cluster property demo, and also testing toggling demo visibility (which used to crash)

## Checklist

Yes, the `id` param is no longer provided to the `remember*Source` functions.

### DIFF
--- a/demo-app/src/commonMain/kotlin/dev/sargunv/maplibrecompose/demoapp/demos/ClusteredPointsDemo.kt
+++ b/demo-app/src/commonMain/kotlin/dev/sargunv/maplibrecompose/demoapp/demos/ClusteredPointsDemo.kt
@@ -20,9 +20,11 @@ import dev.sargunv.maplibrecompose.demoapp.DemoState
 import dev.sargunv.maplibrecompose.expressions.dsl.asNumber
 import dev.sargunv.maplibrecompose.expressions.dsl.asString
 import dev.sargunv.maplibrecompose.expressions.dsl.const
+import dev.sargunv.maplibrecompose.expressions.dsl.convertToNumber
 import dev.sargunv.maplibrecompose.expressions.dsl.feature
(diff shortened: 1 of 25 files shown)


demo-app/src/commonMain/kotlin/dev/sargunv/maplibrecompose/demoapp/demos/ClusteredPointsDemo.kt
```diff
@@ -70,25 +72,21 @@ object ClusteredPointsDemo : Demo {
...
-          //   ),
+          clusterProperties =
+            mapOf(
+              "total_range" to
+                GeoJsonOptions.ClusterPropertyAggregator(
+                  mapper = feature["current_range_meters"].convertToNumber(),
+                  reducer =
+                    feature.accumulated().asNumber() + feature["total_range"].convertToNumber(),
+                )
+            ),
```
sargunv, line 89: I want to tweak this cluster property to maybe sum up the different _types_ of vehicles and show the sums in the cluster somehow. Will leave that for #461.

For now, it's enough that this no longer crashes.

PR #511, opened 2025-07-21
use new clusterMinPoints GeoJson option in MapLibre-native

westnordost:
https://github.com/maplibre/maplibre-native/issues/261 implemented the `clusterMinPoints` property for GeoJsonOptions.

This PR passes the already present option in our `GeoJsonOptions` data class through to native.

To my knowledge, this is not a breaking change. I haven't tested the change.

This change requires a newer version of maplibre-android.


michalgwo:
> To my knowledge, this is not a breaking change. I haven't tested the change.

It doesn't compile when you use older versions of MapLibre Native than v6.17.0, so it is breaking. It's worth mentioning in the changelog. I lost a bunch of time figuring out why it doesn't compile.

PR #517, opened 2025-07-23
only apply bounding box when it changes, to prevent halting the camera animation without cause

### LINKED ISSUE #508: Camera animation stops when performing any other action at the same time on iOS
### What happened?

When I animate the camera to a position and perform any other action at the same time, the camera animation stops. For example, I want to close the keyboard while animating the camera. The code below should zoom to 13, but it stops at 5.
``` kotlin
scope.launch {
    cameraState.animateTo(
        finalPosition = cameraState.position.copy(
            target = Position(19.938830100326083, 50.061658517346814),
            zoom = 13.0
        ),
        duration = 1000.milliseconds
    )
}
keyboardController?.hide()
```

https://github.com/user-attachments/assets/a0cde2cb-4c0d-4a08-8ddb-9e11c517a0d7

I experience the same behaviour when adding a marker to the map while animating the camera.

The issue occurs only on iOS, library v0.10.1. It works on Android, and on v0.9.1, it worked on both mobile platforms. I didn't test it on the Web and Desktop.

### Affected Platforms

- [ ] Android
- [x] iOS
- [ ] Desktop (JVM)
- [ ] Browser (JS)
- [ ] Browser (Wasm)

### Platform Version

iOS 26.0

### Library Version

0.10.1

### Sample Code

I created a sample project: https://github.com/michalgwo/MLCTest/tree/camera-animation-bug-ios

To reproduce the issue:
- run and open the app on iOS
- click on the text field in the top-left corner to show the keyboard
- click on the "Animate camera" button - it should animate to Krakow, Poland with zoom 13, but it stops midway

sargunv:
## Description

this is an attempt to resolve #508.

I added some logging and noticed the camera bounding box was being set whenever the animation stopped. We can't make that BoundingBox class as @Immutable or @Stable because it comes from a library, but we can remember the last set state ourselves and only update if it actually changes. 

## Test plan

Tested with the repro code from #508; animation no longer halts when keyboard collapses.

### DIFF
--- a/lib/maplibre-compose/src/iosMain/kotlin/dev/sargunv/maplibrecompose/core/IosMap.kt
+++ b/lib/maplibre-compose/src/iosMain/kotlin/dev/sargunv/maplibrecompose/core/IosMap.kt
@@ -266,7 +266,10 @@ internal class IosMap(
     mapView.maximumZoomLevel = maxZoom
   }
 
+  private var lastBoundingBox: BoundingBox? = null
+
   override fun setCameraBoundingBox(boundingBox: BoundingBox?) {
+    if (boundingBox == lastBoundingBox) return
     mapView.setMaximumScreenBounds(
       boundingBox?.toMLNCoordinateBounds()
         ?: MLNCoordinateBoundsMake(


sargunv:
Ah nice catch. Will fix.

michalgwo:
> When the `BoundingBox` is not null, it also halts the animation on Android.

What about Android?

PR #520, opened 2025-07-24
android: only set bounding box when it changes

sargunv:
# 517 and #518 fixed this for ios, but it's an issue on android too.


michalgwo:
It resolved the issue, thanks!

Issue #218, opened 2025-01-02
Integrate with Maplibre Native Core


sargunv:
The MapLibre Android and MapLibre iOS SDKs are convenient, but they do a lot of work and hide some internals that would be better for us to implement in compose and KMP. Additionally, there's no such SDK for desktop platforms, and the current web view solution for desktop is problematic. 

If we need to build on the C++ core for desktop, it's good opportunity to do the same for all three native platforms and move more code (gestures, styling, controls, etc) all into common code. For web, we can then disable anything not related to rendering, display under the Compose canvas, and operate with all the same common controls and gestures as native. 

## List of issues that would be solved or aided with this

* Reducing platform specific differences:
	* much of #217
	* #20
	* #164 
	* #355
* Better desktop platform:
	* #219 
	* #239 
* Reimplementing controls solves all of:
	* #77
	* #126
	* #124
	* Indirectly: #210
	* Though of course most of this can be done just by disabling the existing SDK controls and/or PRing improvements upstream.
	
## Alternative

Why not just keep using the iOS and Android SDKs and just disable SDK ornaments, gestures, and animations?

* We'd still need to integrate with the native core on desktop (unless we want to keep the web view around)
* Android/iOS/Desktop would all need dedicated implementations of the whole source/layer glue code, only to each integrate with the same native core underneath

## Downsides of integrating with the native core

* Desktop becomes up to _six_ platforms instead of the one (JVM) now: 
  * Linux, macOS, and Windows, each on x64 and ARM. Though we won't necessarily need to support them all (macOS x64 is being phased out, Windows on ARM is still pretty niche)
  * Upstream seems to support amd64 and arm64 on Linux and macOS, but just amd64 on Windows
  * Need to figure out packaging for that
  	* bundling ~6 native libraries in a "fat" universal jar isn't ideal, though okay for a first pass if it works
* Need to maintain redundant code for Android / iOS integration that already exists upstream
  * TBD: how much exactly?

## Kotlin to C++ integration requirements

* On Android and Desktop, we can do JNI to C++
	* Or https://github.com/jnr/jnr-ffi for simpler integration to C bindings
* Kotlin Native can interop with C or ObjC bindings (also Swift with `@objc`)

So the most straightforward solution is probably to write C bindings for MLN and use it on both sides.

## Platform integration specifics

(mostly guesswork here as I'm still learning about the native core)

### Android

[...]

PR #544, opened 2025-08-11
Native Core integration for desktop target

sargunv:
TODOs after merging:
- file issues for known bugs and follow ups
  - bug: `render()` segfaults on Linux
  - feature: implement vulkan backends on all three platforms
  - build: configure desktop demo app [native distribution properties](https://www.jetbrains.com/help/kotlin-multiplatform-dev/compose-native-distribution.html#platform-specific-options)
  - build: better caching for faster desktop builds
  - build: eliminate submodules in favor of FetchContent (without breaking vcpkg)
  - build: demo app desktop native distributions: notarize for macos, deb/rpm (or appimage?) for linux, exe for windows  

## Description

First step for #218.

This is a hefty PR. It sets up maplibre native core integration for the desktop target, replacing the previous embedded chromium + maplibre js solution. Native core integration was prototyped in https://github.com/sargunv/maplibre-native-jvm-demo. This PR brings that code into maplibre-compose.

Changes:
* Pulled in the code from https://github.com/sargunv/maplibre-native-jvm-demo
  * Except the Vulkan backend; I'll follow up with that later.
* Set up gradle build for C++ natives of all variants (OS + architecture + renderer)
* Integrate just enough of it with maplibre-compose to get to feature parity with the old webview desktop implementation
* Fix some build/format scripts that were broken when developing on Windows or Linux (I mainly develop on macOS)
* Update relevant docs
* Update demo app menu with a layout usable without Compose overlays on desktop+web (no bottom sheet)

Module structure:
- maplibre-compose (added proper desktop support): the main library
- maplibre-compose-material3 (unchanged): material3 extensions for the main library
- maplibre-js-bindings (renamed): Kotlin/JS bindings for MapLibre GL JS.
   - Will someday also include Kotlin/WASM bindings.
- maplibre-native-bindings (new): Kotlin/JVM bindings for MapLibre Native.
   - Will someday also include Android and Kotlin/Native for iOS and macOS bindings.
- maplibre-native-bindings-jni (new): C++ sources providing the native shared library for maplibre-native-bindings on JVM.
   - Distributed as a separate artifact with a classifier per os/arch/renderer tuple.
   - Will someday include be used for Android JNI too, but in that case will be distributed directly in the maplibre-native-bindings AAR.

## Test plan

`just run-desktop` and test it out

Current status:
- `macos-aarch64-metal`: Works
- `macos-amd64-metal`: Works
- `windows-aarch64-opengl`: Works
- `windows-amd64-opengl`: Works
- `linux-aarch64-opengl`: Segfault during render()
- `linux-amd64-opengl`: Segfault during render()
- `*-*-vulkan`: Not implemented


sargunv:
> You should probably involve MapLibre maintainer(s) @louwers and other frequent contributors to maplibre-native in the review of the C++ code

I posted in the OSMUS slack to solicit some reviews of the demo repo when I first put that up, but didn't get any reviews. Planning on sharing this PR there again when it's ready. I'm hoping there will be more interest in reviewing the code that's actually merging. 

There's some refactoring I want to do from this AI generated MVP to something more maintainable, especially around the rendering backend. I already have some idea of changes to make there. But focusing on getting the demo code integrated vertically first. 

So I don't want to solicit reviews here yet; code is still changing heavily. 

> I wonder, what does lib/maplibre-native-jni do exactly, and how does it relate to lib/kotlin-maplibre-native?

* kotlin-maplibre-native (planning to rename to maplibre-native-kotlin) exposes Kotlin APIs to Maplibre Native. Currently just Kotlin JVM for desktop, but someday also Kotlin JVM for Android and Kotlin Native for iOS and macOS. You can think of it as analogous to the MapLibre Android SDK, but a much more direct 1:1 binding to the native core (except around rendering). 
* maplibre-native-jni provides the C++ native implementation for maplibre-native-kotlin. On maven, this jar will have a classifier per native target (3x OS times 2x arch times 2x renderer). And the jar will include the .dylib, .dll, or .so for that target. It's a similar pattern to like, lwjgl natives. 
  * Later for Android, I expect mostly the same code (cmake project) will be used minus the AWT bits, but the binaries will be embedded in the .aar of maplibre-native-kotlin.

> Should this maybe go into a separate repo (under the maplibre org) at one point or is it solely interesting for maplibre-compose?

It's interesting for any Kotlin JVM user of MapLibre, which to my knowledge right now is just maplibre-compose, but that could change with this library becoming available. That's why the name doesn't include "compose". It would be reasonable to split it off into another repo eventually, though it's easier to develop when it's all together. 

The same applies to kotlin-maplibre-js (planning to rename to maplibre-js-kotlin). 

> Finally, since you started the direct integration of maplibre-native, that means that soon-ish you'll remove the Android and the iOS maplibre libraries as dependencies, right?

Maybe not so soon, but eventually yeah. First we gotta get native core integration on desktop to parity with the mobile targets. Then implement the Android Surface renderer, and then for iOS (and macOS) write a second native core integration for Kotlin Native cinterop instead of JNI. 

> Is then (that kotlin PR in) https://github.com/maplibre/maplibre-java still relevant?

TBD. If we still want a pure Kotlin geojson library it's still just as relevant. If we want to simply wrap the C++ geojson library it's not. Really not so different from our choices before the native core integration, except before it's "wrap the Android and iOS library". 

I lean pure Kotlin because managing heap allocations across the JNI boundary is inconvenient, but there is the performance benefit of skipping the translation from Kotlin object to native object for large geojson data.

louwers (commented):
Getting errors like this when trying to build on macOS

/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk/System/Library/Frameworks/CoreFoundation.framework/Headers/CFURL.h:21:9: error: non-defining declaration of enumeration with a fixed underlying type is only permitted as a standalone declaration; missing list of enumerators? [-Welaborated-enum-base]
   21 | typedef CF_ENUM(CFIndex, CFURLPathStyle) {
      |         ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

sargunv:
Oh interesting, which XCode version? It builds in CI and on my ARM Mac, and is configured just like maplibre-native itself. 

I'm on the latest XCode (16.4 I think?)

sargunv:
Oh, the preset also has MLN_WITH_WERROR=OFF, so I think that warning shouldn't be failing it. Did you somehow enable Werror in your environment?

louwers:
XCode 16.4  & Homebrew clang version 20.1.7

I can try Clang 17

PR #638, opened 2025-10-13
Location layer

### LINKED ISSUE #27: Support displaying user location
Both native SDKs have some support for location:
* iOS: https://maplibre.org/maplibre-native/ios/latest/documentation/maplibre-native-for-ios/locationprivacyexample/
* Android: https://maplibre.org/maplibre-native/android/api/-map-libre%20-native%20-android/org.maplibre.android.location/index.html

We should investigate what exactly functionality they provide, and provide it in this project to a similar extent.

kodebach:
## Description

Fixes #27 as discussed in #292

`BasicLocationLayer` (actually multiple layers) implements the location puck, including indicators for accuracy and bearing. `LocationLayer` is a thin wrapper around `BasicLocationLayer` that sets default colours from `MaterialTheme`.

`LocationTrackingEffect` is a wrapper around `LaunchedEffect` that enables updating the `CameraState` based on location changes. This works well enough for most use cases, but is not suitable for a navigation app, because location tracking is currently not very smooth during fast movement. For smoother tracking we would have to extrapolate positions based on the last known position, speed and bearing.

`GeoLocator` is the main interface for the platform-specific APIs for location monitoring. A default implementation for based on the platform APIs `LocationManager` and `CLLocationManager` is provided for Android and iOS respectively. On Android there is also an implementation based on the `LocationEngine` from `org.maplibre.gl:android-sdk` to enable easier migration, as well as a `FusedGeoLocator` based on `FusedLocationProviderClient`.

Desktop is unsupported and will likely remain that way (unless somebody knows how to implement it). Web could be supported, but is not implemented right now.

## Test plan

So far I only tested manually using the new `UserLocationDemo`. More extensive testing will follow.

## Checklist

There should be no breaking changes. Outside the demo app all code lives in new files except for:
- `GestureOptions` gained a few new variants, existing options haven't changed
- `maplibre-compose` now depends on `org.jetbrains.androidx.lifecycle:lifecycle-runtime-compose` to allow pausing location updates when the app is not visible. If to avoid this dependency, we could move all location stuff to a new `maplibre-compose-location` module.

- Android: Emulator, API 35 and API 36
- iOS: not yet tested
- Desktop: unsupported
- Web: not yet supported


westnordost:
Well, simply from a very selfish point of view, it would be nice to have a multiplatform library that enables listening to (precise) location updates using native APIs (i.e. no GMS or only optionally) even if it is just comes down to a few hundred lines of wrapper code (for iOS and Android only, at least). (I need this for my own project 😛).
Plus, of course, this will grow further because 1. JS/Web has its own API, 2. JVM/Desktop has its own API, or even maybe several ones because Java itself is lacking it, so depending on where the Java is running (Windows, Linux, Mac), one has to speak to yet different APIs on native, 3. for general use, one might need/want a more generic API exposed rather than essentially just `StateFlow<Location?>`

I mean, does speak anything *against* outsourcing this part into an own project or module? (IMO it makes sense to have it in an own project/repo because it doesn't have anything *directly* to do with maplibre-compose)

kodebach:
IMHO creating a new location module _does not_ make much sense. Either we keep it as is, implementing just the bare minimum necessary for use in maplibre, or we use [Compass](https://github.com/jordond/compass) and provide an Android implementation of the Compass `geolocator` without Play Services (either as our own module or as a PR upstream).

> for general use, one might need/want a more generic API exposed rather than essentially just `StateFlow<Location?>`

Exactly, so a general user should use something like Compass. We could maybe make this more clear, by renaming `Geolocator` to `LocationFlowProvider`. That should make it clear that it's not a generic interface, but just the way maplibre-compose gets its locations.

In any case, I have no desired to design or maintain a  more generic API. Especially since the platform APIs are not entirely compatible. The best example for this is the desired accuracy for locations: iOS uses 7 levels, the Android platform has 5 levels (all deprecated) plus 3 orthogonal levels for power usage (themselves not deprecated, but all APIs using them are) and GMS uses 4 levels combining accuracy and power in an opaque way. If you want to give users full control, without reaching for platform-specific code, thats gonna be an interesting challenge.

In short the task of a generic library would fall to somebody else.

sargunv:
Fair enough, a separate module def doesn't need to be part of this PR.

lib/maplibre-compose/src/commonMain/kotlin/org/maplibre/compose/location/GeoLocator.kt
```diff
@@ -0,0 +1,39 @@
...
+  public val location: StateFlow<Location?>
+}
+
+public enum class DesiredAccuracy {
+  Highest,
+  High,
+  Balanced,
+  Low,
+  Lowest,
+}
```
sargunv: would be good to add some documentation comments indicating how we map these to platform accuracies

lib/maplibre-compose/src/commonMain/kotlin/org/maplibre/compose/layers/BasicLocationLayer.kt
```diff
@@ -0,0 +1,293 @@
...
+ * @param showBearing whether to show an indicator for [Location.bearing]
+ * @param bearingSize the size of the bearing indicator
+ * @param bearingColor the color of the bearing indicator
+ * @param showBearingAccuracy whether to show an indicator for [Location.bearingAccuracy]
+ * @param onClick a [LocationClickHandler] to invoke when the main location indicator dot is clicked
+ * @param onClick a [LocationClickHandler] to invoke when the main location indicator dot is
+ *   long-clicked
+ */
+@Composable
+public fun BasicLocationLayer(
```
sargunv: naming suggestion: how about `LocationPuck` or `LocationPuckLayers`?

"location puck" -> name it for the thing it renders

plural "layers" (if present at all) -> indicating it's not the same as the fundamental layer types

And `id` to `idPrefix`. Then we can get rid of that NOTE line in the kdoc

sargunv (commented):
Overall looks good! Left some minor comments (and I haven't tested it out yet)

sargunv:
Are you able to test on iOS, or would you need me to validate that piece?

sargunv:
Agreed re: permissions. If we later carve it out into its own library, KMP permissions utilities might become in scope, but def not right now beyond throwing the exception.

Re: iOS and XCode, I still use XCode 16.4 for Kotlin projects because that's the version listed at https://www.jetbrains.com/help/kotlin-multiplatform-dev/multiplatform-compatibility-guide.html#version-compatibility. I use [Xcodes](https://www.xcodes.app) to manage keeping multiple versions installed. I'll update the contributing guide to add some detail here.

Re: development team id, yeah last I looked into this, the only solution I found was to use git clean and smudge filters to scrub the development team id in the git object db 😕. Haven't gotten around to trying that out.

Vichy97:
If my project already implements a location puck via Compass, is it worth migrating to this built in one? What are the main differences?

kodebach:
If you already have an implementation that works for you, I don't see a reason to switch as long as the Compass library is still maintained and you're willing to maintain your own location puck code.

The focus of this PR was actually the `LocationPuck` composable. The `LocationProvider` part (which is what Compass gives you) was only a necessary side effect. If you're using Compass throughout your app, I'd actually recommend your own translation between `LocationPuck` and Compass to keep the location requests consistent in the whole app. Right now, you'd need to implement `LocationProvider` on top of Compass (shouldn't be too hard, it's just a `StateFlow`). But with #737 it gets even easier, then `LocationPuck` takes a `Location` directly, so you just need to map the location data from Compass.

PR #651, opened 2025-10-25
upgrade to org.maplibre.spatialk v0.4.0

sargunv:
now that we have feature generics, some code samples and APIs could be revised. But for now I'm leaving it as is, with type args equivalent to what we had before generics.


westnordost:
Looks good to me. I like that you had to fix the polygon in the test due to stricter validity checking of the input args when constructing geojson data types.

westnordost (approved):

PR #652, opened 2025-10-26
Add RasterDemSource

jgillich:
## Description

Adds the RasterDemSource, which is used by HillshadeLayer. Currently Android-only because I don't own a Mac PC.

## Test plan

Confirmed working with this snippet (it renders hills around Austria)

```kotlin
      HillshadeLayer(
        id = "hillshade123",
        source =
          rememberRasterDemSource(
            "https://demotiles.maplibre.org/terrain-tiles/tiles.json",
            tileSize = 256,
          ),
      )
```

## Checklist

No

- Android:  16

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -20,9 +20,7 @@ jobs:
       - run: |
           git config user.email "41898282+github-actions[bot]@users.noreply.github.com"
           git config user.name "github-actions[bot]"
-          git add .
-          git commit -m 'Apply mise fix'
-          git push
+          git diff --exit-code || (git add . && git commit -m 'Apply mise fix' && git push)
           git diff --exit-code
 
   lint-hk:


lib/maplibre-compose/src/desktopMain/kotlin/org/maplibre/compose/sources/RasterDemSource.kt
```diff
@@ -0,0 +1,18 @@
...
+    this.impl = TODO()
+  }
+
+  public actual constructor(
+    id: String,
+    tiles: List<String>,
+    options: TileSetOptions,
+    tileSize: Int,
+  ) : super() {
+    this.impl = TODO()
```
westnordost: Maybe at least use `TODO("Not implemented")` or similar for clarity, after all, a TODO produces a runtime exception if that code is reached, IIRC.

lib/maplibre-compose/src/jsMain/kotlin/org/maplibre/compose/sources/RasterDemSource.kt
westnordost: The JS documentation would be here:

https://maplibre.org/maplibre-gl-js/docs/API/classes/RasterDEMTileSource/

lib/maplibre-compose/src/androidMain/kotlin/org/maplibre/compose/sources/RasterDemSource.kt
westnordost: Raster DEM sources have the additional property [`encoding`](https://maplibre.org/maplibre-style-spec/sources/#encoding_1) which is missing here. Could be represented with

```kt
/** The encoding used by a Raster DEM source. */
public sealed interface RasterDemEncoding {

  /** Mapbox Terrain RGB tiles. See https://www.mapbox.com/help/access-elevation-data/#mapbox-terrain-rgb for more info */
  public data object Mapbox : RasterDemEncoding

  /** Terrarium format PNG tiles. See https://aws.amazon.com/es/public-datasets/terrain/ for more info. */
  public data object Terrarium : RasterDemEncoding

  /** 
    Custom format using the given [redFactor], [blueFactor], [greenFactor] and [baseShift] parameters.
    Currently only supported in JS.
  */
  public data class Custom(
    /** Value that will be multiplied by the red channel value when decoding. */
    val redFactor: Float = 1f,
    /** Value that will be multiplied by the blue channel value when decoding. */
    val blueFactor: Float = 1f,
    /** Value that will be multiplied by the green channel value when decoding. */
    val greenFactor: Float = 1f,
    /** Value that will be added to the encoding mix when decoding. */
    val baseShift: Float = 0f,
  ) : RasterDemEncoding
}


```

PR #712, opened 2025-11-12
Device orientation support

jgillich:
## Description

Adds a LocationProvider that enhances another LocationProvider with the bearing values of the device rotation sensor. It provides accurate bearing values when the fused location provider is unable to do so (commonly due to lack of movement).

iOS is still TODO

cc @kodebach in case you have some thoughts as the author of the location package

## Test plan

Provider is enabled in demo app for testing

- Android: 16


kodebach (commented):
The `bearing` from `LocationManager` and the `heading`/`orientation` from `SensorManager` are to different things. The bearing is based on direction of travel, while orientation is based on compass heading.

I chose to use `bearing` partly because it is available through `LocationManager`, but also because it seems more relevant to our use case. If we want to support using compass heading, the logic should probably be integrated into the normal `LocationProvider` with an option to switch between `bearing` and `heading`. On iOS both are provided by `CLLocationManager` and for the GMS module there is `FusedOrientationProviderClient`.

I'll leave the decision on how to handle this to @sargunv

jgillich:
> The bearing from LocationManager and the heading/orientation from SensorManager are to different things. The bearing is based on direction of travel, while orientation is based on compass heading.

Indeed, they are different things, but you don't get useful readings from location while standing still and you don't get useful readings from compass while moving.  This combines them in a way that makes sense for _most_ applications - but perhaps not all. We could add this a separate `heading` value on `Location`, I might've been overly focused on my own use case :smile:

sargunv:
> The bearing from LocationManager and the heading/orientation from SensorManager are to different things.

I haven't looked into the code yet, but I'd love to understand (1) what's different, and (2) in what use cases one might prefer one over the other.

My assumptions, may be wrong: one is the direction I'm traveling, the other is the direction the phone is facing. If that's correct, I definitely see use cases for both.
1. When connected to a vehicle, the direction the vehicle is facing
2. When navigating, the direction I'm traveling
3. Otherwise, the direction I'm facing

Very curious how other maps combine these

jgillich:
> one is the direction I'm traveling, the other is the direction the phone is facing

That is my understanding too. Navigation apps will generally assume that your phone is facing in the direction you're traveling, so they're merging these in similar ways (Ride with GPS and Cyclers are two examples that do this). But some apps may want to implement their own logic for handling them, so the current approach isn't flexible enough

How about a `HeadingProvider` that can be passed to `LocationProvider` to compute a dedicated `heading` on `Location`?

lib/maplibre-compose/src/androidMain/kotlin/org/maplibre/compose/location/AndroidSensorEnhancedLocationProvider.kt
```diff
@@ -0,0 +1,134 @@
...
+        override fun onAccuracyChanged(sensor: Sensor?, accuracy: Int) {
+          accuracyDegrees = accuracyToDegrees(accuracy)
+        }
+      }
+
+    val sensor =
+      sensorManager.getDefaultSensor(Sensor.TYPE_ROTATION_VECTOR)
+        ?: throw IllegalStateException("Rotation vector sensor is not available")
+
+    sensorManager.registerListener(listener, sensor, SensorManager.SENSOR_DELAY_NORMAL)
```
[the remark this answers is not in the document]
DestBro: I'm late to the party, but it looks like @kodebach is trying to limit "spamish" sensor values and get less but more relevant updates, ideally without duplicates/repetitions, right? Just wanted to mention `Flow::distinctUntilChanged`—if you're using a Flow, it’s worth checking out too.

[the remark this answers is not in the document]
kodebach: `.flowOn` moves the processing of the flow, but the listener is invoked by `SensorManager` and completely separate from the flow processing (that's why you need the `trySend`). The listener is still invoked on the main thread, so `trySend` and all the rotation matrix stuff happens on the main thread every time we get a sensor update.

`buffer(1)` (or even better `conflate()`) does discard values, if downstream isn't ready to process them. But it will still attempt updates and in our case cause recomposition as fast as possible. IMO it would be better to use `sample` and limit the number of updates per time interval.

> There is one limitation here, you can't read more often than SENSOR_DELAY_NORMAL, which is ~200ms I think. I'll add a note in the docs

The [Android Docs](https://developer.android.com/develop/sensors-and-location/sensors/sensors_overview#:~:text=The%20delay%20that%20you%20specify%20is%20only%20a%20suggested%20delay.) actually say that the delay is only a suggestion and that you will normally receive updates faster. [Further down] (https://developer.android.com/develop/sensors-and-location/sensors/sensors_overview#sensors-rate-limiting) there is a mention of a maximum update rate of 200 Hz (i.e. a delay of 5 ms) `HIGH_SAMPLING_RATE_SENSORS` permission.

kodebach:
> It may also make sense to separate location providers (position, course, speed) from orientation providers

I think that's an unnecessary complication. Apart from Android (and maybe desktop if we ever support it) all platforms will probably only ever have a single implementation for location and orientation providers. Even on Android with the GMS special case, I'd don't think there would be big use case in combining platform location with GMS orientation or vice versa. If someone really needs it, they can always implement their own `LocationProvider` since the interface is public.

I think separating them just introduces unnecessary combining logic, unless we completely separate the two and only combine them at the `UserLocationState` level like this:

```kotlin
public class UserLocationState internal constructor(
  locationState: State<Location?>,
  orientationState: State<Orientation?>,
) {
  public val location: Location? by locationState
  public var orientation: Orientation? by orientationState
}


@Composable
public fun rememberUserLocationState(
  locationProvider: LocationProvider,
  orientationProvider: OrientationProvider? = null,
  headingEnabled: Boolean = false,
  lifecycleOwner: LifecycleOwner = LocalLifecycleOwner.current,
  minActiveState: Lifecycle.State = Lifecycle.State.STARTED,
  coroutineContext: CoroutineContext = EmptyCoroutineContext,
): UserLocationState {
  val orientationState =
    orientationProvider?.orientation?.collectAsStateWithLifecycle(
      lifecycleOwner = lifecycleOwner,
      minActiveState = minActiveState,
      context = coroutineContext,
    )
  val locationState =
    locationProvider.location.collectAsStateWithLifecycle(
      lifecycleOwner = lifecycleOwner,
      minActiveState = minActiveState,
      context = coroutineContext,
    )
  return remember(locationState, orientationState) {
    UserLocationState(locationState, orientationState ?: mutableStateOf(null))
  }
}
```

That has pros and cons. On the pro side, it makes it clear that these are separate things that might update at different rates. On the con side, it also means that `LocationPuck` and user code may need to combine/synchronise the updates themselves.

demo-app/src/androidMain/kotlin/org/maplibre/compose/demoapp/demos/GmsLocationDemo.kt
```diff
@@ -32,19 +34,31 @@ object GmsLocationDemo : Demo {
...
       val locationState = rememberUserLocationState(locationProvider)
 
       LocationPuck(
         idPrefix = "gms-location",
-        locationState = locationState,
+        location = locationState.location,
+        bearing =
+          locationState.location?.let { location ->
+            val courseAccuracy = location.course?.inaccuracy ?: 180.degrees
+            val orientationAccuracy = location.orientation?.inaccuracy ?: 180.degrees
```
kodebach:
```suggestion
            val courseAccuracy = location.course?.inaccuracy ?: Double.POSITIVE_INFINITY.degrees
            val orientationAccuracy = location.orientation?.inaccuracy ?: Double.POSITIVE_INFINITY.degrees
```

While `180.degrees` produces the correct result, `Double.POSITIVE_INFINITY.degrees` makes the intention more clear for the demo

demo-app/src/commonMain/kotlin/org/maplibre/compose/demoapp/demos/UserLocationDemo.kt
```diff
@@ -74,11 +75,12 @@ object UserLocationDemo : Demo {
...
       idPrefix = "user-location",
-      locationState = state.locationState,
+      location = state.locationState.location,
       cameraState = state.cameraState,
       accuracyThreshold = 0f,
-      showBearing = bearingUpdate != BearingUpdate.IGNORE,
-      showBearingAccuracy = bearingUpdate != BearingUpdate.IGNORE,
+      bearing =
+        if (bearingUpdate != BearingUpdate.IGNORE) state.locationState.location?.orientation
+        else null,
```
kodebach:
```suggestion
      bearing = state.locationState.location?.orientation?.takeUnless { bearingUpdate == BearingUpdate.IGNORE }
```

IMHO this is one of the few use cases for `takeUnless`

lib/maplibre-compose/src/commonMain/kotlin/org/maplibre/compose/location/LocationRequest.kt
```diff
@@ -0,0 +1,20 @@
...
+ *   Defaults to true.
+ * @param course True if the course should be included in the location update, false otherwise.
+ *   Defaults to true.
+ * @param orientation True if the orientation should be included in the location update, false
+ *   otherwise. Defaults to false.
+ */
+public data class LocationRequest(
+  val position: Boolean = true,
+  val speed: Boolean = true,
+  val course: Boolean = true,
```
kodebach: Do these use separate APIs from `position` on any platform? If not, they are essentially free and there should be no reason to toggle them separately.

lib/maplibre-compose/src/commonMain/kotlin/org/maplibre/compose/location/LocationPuck.kt
```diff
@@ -49,59 +49,62 @@ import org.maplibre.compose.util.ClickResult
...
-  val bearingAccuracyPainter =
-    rememberBearingAccuracyPainter(
-      sizes = sizes,
-      colors = colors,
-      bearingAccuracy = locationState.location?.bearingAccuracy ?: 0.0,
-    )
-  val locationSource = rememberLocationSource(locationState)
+  val locationSource = rememberLocationSource(location, bearing)
+
+  val positionAccuracy = location?.position?.inaccuracy?.inMeters?.toFloat() ?: 0f
```
kodebach:
```suggestion
  val positionAccuracy = location?.position?.inaccuracy?.inMeters?.toFloat() ?: Float.NEGATIVE_INFINITY
```

So we never show the accuracy layer if we don't have a value, even if someone passes a negative `accuracyThreshold`.

lib/maplibre-compose/src/commonMain/kotlin/org/maplibre/compose/location/LocationPuck.kt
```diff
@@ -263,21 +272,27 @@ private fun rememberBearingAccuracyPainter(
 }
 
 @Composable
-private fun rememberLocationSource(locationState: UserLocationState): GeoJsonSource {
+private fun rememberLocationSource(
+  location: Location?,
+  bearing: BearingMeasurement?,
```
kodebach:
```suggestion
  position: PositionMeasurement?,
  bearing: BearingMeasurement?,
  timestamp: TimeMark,
```

This one is private it should definitely should use what it needs, there is no "nicer public API" argument here

kodebach (changes requested):
I don't like the nullable `position` in `Location`. A location without a position doesn't make sense IMO. Same goes for toggling speed and course separately from position in `LocationRequest`. AFAIK these come from the same API on all platforms, i.e. we get all of them for the same cost. No reason to control them separately.

The API otherwise LGTM except for a few minor details. What I forgot to mention before is, that `UserLocationState` also exists to ensure we can pass around a `remember`ed value that doesn't change with ever location/orientation update, which let's us avoid recomposition.

sargunv:
> Apart from Android (and maybe desktop if we ever support it)

Devices without GPS receivers may want to use some network geolocation provider.

Desktop will probably have three platform providers (macOS Core Location, Windows.Devices.Geolocation, Geoclue), maybe encapsulated into one universal "platform" provider for convenience.

> it makes it clear that these are separate things that might update at different rates

This would be the strongest justification to separate them I think. I don't have a solid opinion on this yet, will have to play with it a bit. 

I'll also be out this weekend, so can take a look next week

lib/maplibre-compose-gms/src/androidMain/kotlin/org/maplibre/compose/gms/FusedLocationProvider.kt
```diff
@@ -80,6 +107,20 @@ constructor(
 
           awaitClose { locationClient.removeLocationUpdates(callback) }
         }
+        .let { flow ->
+          if (!request.orientation) {
+            return@let flow
+          }
+          flow.zip(orientation) { location, (orientation, orientationAccuracy) ->
```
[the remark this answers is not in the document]
kodebach: True, but you can force the orientation flow to emit at least once, by calling `send(null)` similar to how the location flow calls `send(lastLocation)`. But yeah, doesn't matter, if we split it

sargunv:
Ah sorry, you're correct. 

For clarity, the states:
- location data is available. We have a timestamp, position and optionally altitude, bearing, speed, and their accuracies. 
- location data is unavailable.

And separately, orientation data:
- orientation data is available. We have a timestamp, bearing and optionally an accuracy.
- orientation data is not available.

DestBro:
Also i found a bug with the current LocationPuck:
image

This red semi-circle should not even appear since the "bearing" is turned off

jgillich:
:point_up:

Issue #716, opened 2025-11-14
MacOS/JVM: crash when the map is removed from the UI during a transition


TheLortex:
### What happened?

If a transition is happening while the map composable is removed, the program crashes with a SIGSEGV:

```
# A fatal error has been detected by the Java Runtime Environment:
#
#  SIGSEGV (0xb) at pc=0x0000000140b06d8c, pid=85827, tid=87335
#
# JRE version: OpenJDK Runtime Environment JBR-21.0.5+8-762.40-nomod (21.0.5+8) (build 21.0.5+8-b762.40)
# Java VM: OpenJDK 64-Bit Server VM JBR-21.0.5+8-762.40-nomod (21.0.5+8-b762.40, mixed mode, sharing, tiered, compressed oops, compressed class ptrs, g1 gc, bsd-aarch64)
# Problematic frame:
# C  [libmaplibre-jni.dylib+0xe6d8c]  mbgl::Map::cancelTransitions()+0x10
#
```

While programming the repro, I couldn't obtain the exact same issue, but a probably related one:
```
#  SIGSEGV (0xb) at pc=0x000000013ca9f3a8, pid=86115, tid=90131
#
# JRE version: OpenJDK Runtime Environment JBR-21.0.5+8-762.40-nomod (21.0.5+8) (build 21.0.5+8-b762.40)
# Java VM: OpenJDK 64-Bit Server VM JBR-21.0.5+8-762.40-nomod (21.0.5+8-b762.40, mixed mode, sharing, tiered, compressed oops, compressed class ptrs, g1 gc, bsd-aarch64)
# Problematic frame:
# C  [libmaplibre-jni.dylib+0xeb3a8]  non-virtual thunk to mbgl::Map::Impl::onTileAction(mbgl::TileOperation, mbgl::OverscaledTileID const&, std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char>> const&)+0x28
``` 

### Affected Platforms

- [ ] Android
- [ ] iOS
- [x] Desktop (JVM)
- [ ] Browser (JS)
- [ ] Browser (Wasm)

### Platform Version

macOS 26.0.1

### Library Version

0.12.1

### Sample Code

```
@Composable
fun BugRepro() {
    val cameraState = rememberCameraState()
    val coroutineScope = rememberCoroutineScope()
    var showMap by remember { mutableStateOf(true) }
    Column {
        Button(onClick = {
            coroutineScope.launch {
[...]

Issue #572, opened 2025-09-01
Android/iOS native core integration


sargunv:
#544 added basic native core integration for desktop. Once that matures and catches up to our existing MapLibre Android/iOS wrappers, we should look at using our own native core integration for Android and iOS too, to eliminate API differences between the different SDKs.

* Do this only after #570 is complete.
* The **maplibre-native-bindings** module should become a KMP module with Android and Native (iOS) targets in addition to the existing JVM target (for naming, see #549)
* For Android, we should be able to re-use most of the **maplibre-native-bindings-jni** module, except integrating with an Android Surface instead of an AWT Canvas. But rather than publishing the C++ natives as their own artifact, they should be integrated in to the .aar for **maplibre-native-bindings**.
* For iOS, we'll need new Kotlin/Native cinterop bindings to MapLibre Native. We can do this through ObjC (for apple platforms) or C (for all native platforms). Compose Multiplatform today only supports Apple platforms for Native, but there's other folks working on Linux and probably Windows, so it's worth trying to write a C ABI.
* If we get a working C ABI for native, it's worth investigating whether we can use that to streamline or replace the JNI integration. But we have working JNI right now, so I wouldn't abandon that to build the C one first.

Tldr: order of operations is:
* Get desktop (JVM) to parity with mobile
* Add Android support to our JNI bindings
* Add iOS support via (hopefully) C bindings
* Investigate porting JVM and Android to the same C bindings

sargunv:
Android native core integration is merged. #985 will do iOS

Issue #550, opened 2025-08-14
consider decoupling the map lifecycle from the map view


sargunv:
### Feature Description

Currently, if you remove and reinsert the composable containing a map, the map will be destroyed and reinitialized, and there's nothing the user can do to retain it. With the Android and iOS SDKs, that's necessary, because the map is tied to its View, which is owned by the composable. 

It's hard to reason about a composable's life cycle, and composables should be designed to be cheap to reinitialize. So in an ideal world, it should be possible for the user to take control of the map lifecycle, for example retaining it in a ViewModel or a Compose 1.10 `retain` block regardless of whether it's displayed in the UI at the moment. And have it be cheap to attach an already initialized map to the UI. This will enable easier testing and a more predictable map life cycle.

With native core integration, we control the renderer frontend ourselves, and so have the opportunity to decouple the map from the surface it renders to. Unsure if this is possible with MapLibre JS; can we detach and re-attach a Canvas from the map?

Prerequisites:
- complete native core integration for desktop, Android, and iOS
- investigate how such a change would affect web

PR #736, opened 2025-12-30
Implement GeoJsonSource for web

dellisd:
## Description

This is a pass at implementing a source for the web version. The js bindings behave somewhat differently from the native/mobile bindings in the sense that the underlying source implementation objects can only _really_ be created using a reference to the `Map` object. (This will also apply to layers when those are implemented).

I've tried to get around this by "binding" the source implementations to the maplibre-compose object after the source has been added to the map, but this means that some method calls won't work _until_ the source has been added to a map which I fear will introduce some lifecycle headaches when trying to call these methods.

It's also possible this might not end up being an issue in practice, but I haven't been able to test those behaviours yet. 

Similarly, many maplibre-gl-js functions return `Promise`s while their native/mobile equivalents are synchronous calls. This effectively makes them impossible to implement under the current APIs.

I wanted to open this PR sooner than later to raise this potential issue in case there are other thoughts (or maybe this has already been raised before?)

## Test plan

TBD

## Checklist

No

- Web: Chrome 143.0.7499.170

### DIFF
--- a/lib/maplibre-compose/src/jsMain/kotlin/org/maplibre/compose/sources/ComputedSource.kt
+++ b/lib/maplibre-compose/src/jsMain/kotlin/org/maplibre/compose/sources/ComputedSource.kt
@@ -1,11 +1,19 @@
 package org.maplibre.compose.sources
 
+import org.maplibre.kmp.js.stylespec.sources.SourceSpecification
 import org.maplibre.spatialk.geojson.BoundingBox
 import org.maplibre.spatialk.geojson.FeatureCollection
 
 public actual class ComputedSource : Source {
 
   @Suppress("UNREACHABLE_CODE") override val impl: Nothing = TODO()
+
+  override val spec: SourceSpecification
+    get() = TODO("Not yet implemented")
+
+  override fun bind(source: org.maplibre.kmp.js.source.Source) {
+    TODO("Not yet implemented")
+  }
 
   public actual constructor(
     id: String,
--- a/lib/maplibre-compose/src/jsMain/kotlin/org/maplibre/compose/sources/GeoJsonSource.kt
+++ b/lib/maplibre-compose/src/jsMain/kotlin/org/maplibre/compose/sources/GeoJsonSource.kt
@@ -1,17 +1,36 @@
 package org.maplibre.compose.sources
 
 import kotlinx.serialization.json.JsonObject
+import org.maplibre.compose.util.jso
+import org.maplibre.kmp.js.stylespec.sources.GeoJSONSourceSpecification
+import org.maplibre.kmp.js.stylespec.sources.GeoJsonDataDefinition
 import org.maplibre.spatialk.geojson.Feature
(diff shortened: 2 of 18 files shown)


sargunv:
Thanks for working on this!

> I've tried to get around this by "binding" the source implementations to the maplibre-compose object after the source has been added to the map, but this means that some method calls won't work until the source has been added to a map which I fear will introduce some lifecycle headaches when trying to call these methods.

I'm a bit short on time to dig into this deeply, but this sounds relevant to the future planned decoupling of the style manager from the map in order to support the snapshotter (#266). Snapshotter is native-only, but since the API is tightly coupled to the map on web, that'll be harder to do.

Will have to think more about approaches to deal with it. My initial, unvalidated line of thought:
- on maplibre-gl-js, the style definition (JSON) is a separate object from the `Style` instance itself
- for the snapshotter on native, we only _really_ need the style definition
- similarly, our Composeable style manager primarily (entirely) updates the paint/layout properties of the style - the definition.
- so, most likely, we should separate the concept of a style definition from a style instance, and couple only the latter to a map. This'd be a major refactor, and might be tricky to get right on native
- I don't want to make the native code _too_ awkward just to support JS. If it can be done with a reasonable end state, I think a major refactor is worth it. But if the refactor leads to a really weird implementation for Native, I'd rather drop JS support

sargunv:
Short term, my intention is to make progress towards JS by just working on fleshing out the bindings first, without the maplibre-compose integration (#243)

Map lifecycle management needs a rethinking anyway (see all the weirdness around SafeStyle, and the dispose issues on desktop (#716) so it's good we're hitting these constraints now so we can consider them when rethinking the map lifecycle.

Also relevant:
- #550 
- Some ask earlier about initializing map state outside Compose, like in a view model, for which I can't find the discussion now

mtpdog:
@sargunv Could you please clarify why the layer (e.g. SymbolLayer) needs 'source' as a parameter not a sourceId? The same source could theoretically be used to draw several layers. My understanding is that it would be better to split Source and Layer instances and handle them separately (update source data in one place and layer rendering in another) with only a sourceId being kept as a layer-to-source link. I believe this is how it is done in Native.

jgillich:
@mtpdog You can reuse sources with multiple layers. Also I believe this question is off-topic here

sargunv:
> underlying source implementation objects can only really be created using a reference to the Map object

circling back, since #834 this is now also true on desktop, and will soon be true on ios and android through #572. and the native api is _much_ more under our control than it was before, so there's a good chance we can reshape things to be convenient for compose and for a common api.

not 100% sure yet if the web variant will end up on maplibre-native-ffi too. it's technically possible but the toughest integration, and maplibre gl js is still the ideal fit for web.

will spend some time diving into ml-gl-js's api surface now that I've redone native, hopefully can align things in a way that gets web unstuck finally.

sargunv:
superseded by #848

PR #737, opened 2026-01-05
Add orientation provider

jgillich:
Continuation of #712

* New orientation provider returns device orientation
* rememberUserLocationState will return orientation if orientationProvider is provided
* Updated demos to display orientation
* Renamed location bearing to course
* Added measurement types (value+accuracy) for location members
* LocationPuck accepts BearingMeasurement to support both course and orientation
* Use spatialk/units for location provider minDistance

TODO:

* ~~GMS Demo crashes when you hit close~~
* ~~FusedOrientationProvider emits much more frequently than the provided `MinUpdateInterval` - should we `.sample`? I wonder if `FusedLocationProvider` does this too~~
* iOS


sargunv:
Thanks again for working on this! Will hold off on reviewing as it's in draft mode, but feel free to @ me anytime when you'd like a review

lib/maplibre-compose-gms/src/androidMain/kotlin/org/maplibre/compose/gms/FusedLocationProvider.kt
```diff
@@ -57,7 +56,7 @@ constructor(
           val callback =
             object : LocationCallback() {
               override fun onLocationResult(result: LocationResult) {
-                result.locations.forEach { trySendBlocking(it.asMapLibreLocation()).getOrThrow() }
+                result.locations.forEach { trySend(it.asMapLibreLocation()) }
```
jgillich, line 59: This had the potential to throw a `CancellationException` (`ForgottenCoroutineScopeException`) if values are sent after cancellation. As far as I know, there are only two failure scenarios, either the channel is closed or the channel is not being collected and the buffer exceeds capacity, neither of which justify throwing.

lib/maplibre-compose/src/commonMain/kotlin/org/maplibre/compose/location/UserLocationState.kt
```diff
@@ -3,31 +3,85 @@ package org.maplibre.compose.location
...
+ * @param coroutineContext The [CoroutineContext] to use for collecting updates. Defaults to
+ *   [EmptyCoroutineContext].
+ * @return A remembered [UserLocationState] instance.
+ */
+@OptIn(FlowPreview::class)
 @Composable
 public fun rememberUserLocationState(
   locationProvider: LocationProvider,
+  orientationProvider: OrientationProvider? = null,
+  samplePeriod: Duration? = 1.seconds,
```
jgillich: I added a sampling period here to avoid doubling recompositions when orientationProvider is used. It is up to the user to sync up all the periods, potential pitfall if they miss this one. Not sure if it would be better to remove the default value entirely

lib/maplibre-compose/src/commonMain/kotlin/org/maplibre/compose/location/OrientationProvider.kt
```diff
@@ -0,0 +1,26 @@
...
+  public override val orientation: StateFlow<Orientation?> = MutableStateFlow(null)
+}
+
+@Composable
+public expect fun rememberDefaultOrientationProvider(
+  updateInterval: Duration = 1.seconds
+): OrientationProvider
+
+@Composable
+public fun rememberNullOrientationProvider(): OrientationProvider {
```
jgillich, line 23: Do we want/need this? Copied it from LocationProvider

jgillich:
This is ready for review btw @sargunv (and @kodebach in case you want to)

lib/maplibre-compose-gms/src/androidMain/kotlin/org/maplibre/compose/gms/FusedLocationProvider.kt
kodebach: Seems reasonable

lib/maplibre-compose/src/commonMain/kotlin/org/maplibre/compose/location/LocationPuck.kt
```diff
@@ -50,34 +50,44 @@ import org.maplibre.compose.util.ClickResult
...
+ * @param showBearingAccuracy Whether to show the bearing accuracy indicator.
+ * @param onClick A [LocationClickHandler] to invoke when the main location indicator dot is
+ *   clicked.
+ * @param onLongClick A [LocationClickHandler] to invoke when the main location indicator dot is
+ *   long-clicked.
  */
 @Composable
 public fun LocationPuck(
   idPrefix: String,
   locationState: UserLocationState,
```
kodebach: I think this should also be:

```suggestion
  position: PositionMeasurement?
  speed: SpeedMeasurement?
```

That makes it more clear that `position`, `speed` and `bearing` can be from different sources. It also simplifies using `LocationPuck` if you have your own location source, shared throughout our app.

lib/maplibre-compose/src/commonMain/kotlin/org/maplibre/compose/location/LocationPuck.kt
kodebach, line 96: We don't need `showBearing` anymore, if `bearing` is a separate parameter. Instead of `showBearing = false` you can just pass `bearing = null` now.

lib/maplibre-compose/src/commonMain/kotlin/org/maplibre/compose/location/LocationTrackingEffect.kt
```diff
@@ -88,8 +92,15 @@ public enum class BearingUpdate {
   /** ignore changes in bearing and reset orientation to point north */
   ALWAYS_NORTH,
 
-  /** update camera orientation based on location bearing */
-  TRACK_LOCATION,
+  /** update camera rotation based on location course (direction of movement) */
+  TRACK_COURSE,
+
+  /** update camera rotation based on device orientation (heading) */
+  // TODO TRACK_ORIENTATION,
```
kodebach: Will this be part of the PR?

lib/maplibre-compose/src/commonMain/kotlin/org/maplibre/compose/location/LocationTrackingEffect.kt
```diff
@@ -88,8 +92,15 @@ public enum class BearingUpdate {
...
+
+  /** update camera rotation based on course if available, otherwise orientation */
```
kodebach: Comment is wrong, currently is exactly the same as `TRACK_COURSE`.

lib/maplibre-compose/src/commonMain/kotlin/org/maplibre/compose/location/OrientationProvider.kt
kodebach: The `NullLocationProvider` exists, so you can call `rememberUserLocationState` with it. Which was necessary, because `LocationPuck` needs a `UserLocationState` and you can't construct that without calling `rememberUserLocationState`.

Since `orientationProvider` is nullable in `rememberUserLocationState` I don't think `NullOrientationProvider` is needed. However, we should probably be consistent, so either keep both with non-null parameters in `rememberUserLocationState` or remove both and make the parameters nullable.

lib/maplibre-compose/src/iosMain/kotlin/org/maplibre/compose/location/IosOrientationProvider.kt
```diff
@@ -0,0 +1,11 @@
+package org.maplibre.compose.location
+
+import androidx.compose.runtime.Composable
+import kotlin.time.Duration
+
+@Composable
+public actual fun rememberDefaultOrientationProvider(
+  updateInterval: Duration
+): OrientationProvider {
+  TODO()
```
kodebach: On iOS (compass) orientation is also provided through `CLLocationManager` (https://developer.apple.com/documentation/corelocation/getting-heading-and-course-information).

I think we can just make `IosLocationProvider` implement both `LocationProvider` and `OrientationProvider` and add `enableLocation`/`enableOrientation` constructor parameters. Alternatively, you can also duplicate the whole class.

kodebach (changes requested):
I think making `UserLocationState` a parameter of `LocationPuck` was actually a mistake and we should fix that. The new `bearing` parameter would defeat the point of passing location as a state object anyway.

lib/maplibre-compose/src/commonMain/kotlin/org/maplibre/compose/location/LocationPuck.kt
jgillich: `rememberLocationSource` further down accesses `location.timestamp`, that's why I left it like it is

lib/maplibre-compose/src/commonMain/kotlin/org/maplibre/compose/location/LocationTrackingEffect.kt
jgillich: Umm it already is, you seem to have clicked on an individual commit  :smiley:

lib/maplibre-compose/src/commonMain/kotlin/org/maplibre/compose/location/LocationPuck.kt
jgillich: Ah this is also outdated, at HEAD it takes `Location?` and not `UserLocationState`

lib/maplibre-compose/src/commonMain/kotlin/org/maplibre/compose/location/LocationPuck.kt
```diff
@@ -50,34 +50,44 @@ import org.maplibre.compose.util.ClickResult
...
+ * @param onClick A [LocationClickHandler] to invoke when the main location indicator dot is
+ *   clicked.
+ * @param onLongClick A [LocationClickHandler] to invoke when the main location indicator dot is
+ *   long-clicked.
  */
 @Composable
 public fun LocationPuck(
   idPrefix: String,
-  locationState: UserLocationState,
+  location: Location?,
```
kodebach, line 89: Continuing from https://github.com/maplibre/maplibre-compose/pull/737#discussion_r2755077806

Hm, yeah passing the timestamp also as a separate parameter would be a bit weird too. The timestamp isn't even really used, other than to check against `oldLocationThreshold`. So in this scenario it should really be an `isOutdated`.

Not sure, which compromise is better. We can probably leave it for now.

lib/maplibre-compose/src/commonMain/kotlin/org/maplibre/compose/location/LocationPuck.kt
```diff
@@ -50,34 +50,44 @@ import org.maplibre.compose.util.ClickResult
...
+  bearing: BearingMeasurement? = location?.course,
   cameraState: CameraState,
```
kodebach:
```suggestion
  cameraState: CameraState,
  bearing: BearingMeasurement? = location?.course,
```

non-optional parameters should be listed first

lib/maplibre-compose/src/commonMain/kotlin/org/maplibre/compose/location/OrientationProvider.kt
jgillich: Ah I see, I switched it over to default to `NullOrientationProvider`

lib/maplibre-compose/src/commonMain/kotlin/org/maplibre/compose/location/LocationTrackingEffect.kt
```diff
@@ -109,20 +144,49 @@ internal class LocationChangeCollector(private val onEmit: suspend LocationChang
   ) {
     val cameraState = this
 
+    val target = currentLocation.location?.position?.value ?: cameraState.position.target
+    val course =
+      currentLocation.location?.course?.value?.clockwiseRotationTo(Bearing.North)?.inDegrees
+    val courseAccuracy = currentLocation.location?.course?.accuracy ?: 180.degrees
```
kodebach: I'd use `Double.POSITIVE_INFINITY.degrees` just to be absolutely sure it's always bigger, even if the other value is somehow out of bounds.

lib/maplibre-compose/src/commonMain/kotlin/org/maplibre/compose/location/UserLocationState.kt
kodebach: I think a better solution would be to add a

```kotlin
val samplePeriod: Duration
```

to the `*Provider` interfaces. The you can just use the max of those for the `flow.sample` call (or the code I suggest below).

But after thinking about it, I'm actually not sure what you're trying to avoid here. The trivial solution (a second `collectAsStateWithLifecycle`) shouldn't lead to unnecessary recompositions, assuming you want a recomposition every time either location or orientation changes. That recomposition should also just affect the composables that _read_ `location` or `orientation` from the returned `UserLocationState`, i.e. normally just `LocationPuck`.

lib/maplibre-compose/src/commonMain/kotlin/org/maplibre/compose/location/UserLocationState.kt
```diff
@@ -3,31 +3,85 @@ package org.maplibre.compose.location
...
+      .collectAsStateWithLifecycle(
+        initialValue = null to null,
+        lifecycleOwner = lifecycleOwner,
+        minActiveState = minActiveState,
+        context = coroutineContext,
+      )
+
+  return remember(state) {
+    UserLocationState(mutableStateOf(state.first), mutableStateOf(state.second))
+  }
```
kodebach: Pretty sure this `combine`/`sample` stuff makes it worse than before. `combine` and `sample` are both pretty complex functions that do a lot behind the scenes. Your also now constructing a new `Pair` object every time either value changes. The two additional `mutableStateOf` probably also don't help. 

If you want to sync the sampling of location and orientation, I'd suggest something like:

```kotlin
  val state = remember { UserLocationState() }

  LaunchedEffect(x, y, lifecycleOwner.lifecycle, minActiveState, coroutineContext) {
    var currentLocation = locationProvider.location.value
    var currentOrientation = orientationProvider.orientation.value
      
    suspend fun collect() {
      x.onEach { currentLocation = it }.launchIn(this)
      y.onEach { currentOrientation = it }.launchIn(this)
      while(isActive) {
        state.location = currentLocation
        state.orientation = currentOrientation
        delay(samplePeriod)
      }
    }

    lifecycleOwner.lifecycle.repeatOnLifecycle(minActiveState) {
      if (context == EmptyCoroutineContext) {
        collect()
      } else {
        withContext(coroutineContext) {
          collect()
        }
      }
    }
  }

  return state
```

plus

```kotlin
public class UserLocationState
internal constructor() {
    /** The user's current or last known location */
    public var location: Location? by mutableStateOf(null)
        internal set
    
    /** The device's current or last known orientation */
    public var orientation: Orientation? by mutableStateOf(null)
        internal set
}
```

kodebach (changes requested):
This time for the correct commit

lib/maplibre-compose/src/commonMain/kotlin/org/maplibre/compose/location/LocationPuck.kt
jgillich: Two reasons for this:

1. We also have `showBearingAccuracy: Boolean`, for consistency's sake it would be odd to turn off one via `null` and the other via `false`. We could remove both of these flags and only show them unless null, but:
2. We may, at some point in the future, want to indicate that bearing is unknown rather than disabled

lib/maplibre-compose/src/commonMain/kotlin/org/maplibre/compose/location/UserLocationState.kt
jgillich: If you have location at 1s and orientation at 1s, your UI will recompose every 0.5s on average. That is unnecessary for most apps. In my navigation app, I have a bunch of effects that respond to location changes, and keeping energy usage as low as possible is important to me. Although it is possible that this ordeal of combine/sample uses more energy than not doing it at all, so I'll see about optimizing it

kodebach: If you really want to reduce energy consumption, I would tune the parameters of the location/orientation providers so they emit fewer or less accurate updates in the first place. The most expensive part (in terms of energy) is definitely reading and processing the sensor values.

lib/maplibre-compose/src/commonMain/kotlin/org/maplibre/compose/location/UserLocationState.kt
jgillich: I implemented something similar to this. Since we're using `StateFlow`, `.first()` returns with not delay and we don't need the `onEach` construct

lib/maplibre-compose/src/commonMain/kotlin/org/maplibre/compose/location/UserLocationState.kt
```diff
@@ -1,33 +1,89 @@
...
+    locationProvider,
+    orientationProvider,
+    lifecycleOwner.lifecycle,
+    minActiveState,
+    coroutineContext,
+  ) {
+    suspend fun collect() {
+      while (isActive) {
+        locationState.value = locationProvider.location.first()
+        orientationState.value = orientationProvider.orientation.first()
```
kodebach:
```suggestion
        locationState.value = locationProvider.location.value
        orientationState.value = orientationProvider.orientation.value
```

Since it's `StateFlow`s you can just use `.value` and that makes it _really_ obvious that it cannot suspend.

kodebach (approved):
LGTM now

lib/maplibre-compose/src/commonMain/kotlin/org/maplibre/compose/location/UserLocationState.kt
jgillich: Oh, TIL

jgillich:
Actions is broken again. Dammit Github, get it together :man_facepalming:

lib/maplibre-compose/src/commonMain/kotlin/org/maplibre/compose/location/UserLocationState.kt
jgillich: `.value` actually does not work because of `SharingStarted.WhileSubscribed` passed to `stateIn`. Do you know what's the best solution to this, shall we use `.first()` or `SharingStarted.Eagerly`?

jgillich: ~~I'm leaning towards `.first()` with a dynamic `stopTimeoutMillis` based on the interval~~

Edit: actually just a permanent subscriber is best

kodebach: If `value` doesn't work, `first()` would suspend until a value is available. So you would need a `launch`/`async` anyway, to avoid blocking orientation updates until the first location is available and vice-versa. The `onEach`+`launchIn` option definitely seems best.

lib/maplibre-compose/src/commonMain/kotlin/org/maplibre/compose/location/UserLocationState.kt
```diff
@@ -1,33 +1,121 @@
...
+ *    indicates the direction the top of the device is pointing.
+ *
+ * It compares the accuracy of these two measurements and returns the one with the smallest accuracy
+ * value (i.e., the most precise). If a measurement has no accuracy specified (`null`), it is
+ * treated as having infinite (the worst possible) accuracy.
+ *
+ * @return The [BearingMeasurement] with the highest accuracy, or `null` if both [Location] and
+ *   [Orientation] are `null` or do not provide a bearing.
+ */
+public fun UserLocationState.minAccuracyBearing(): BearingMeasurement? {
```
kodebach:
```suggestion
public fun UserLocationState.mostAccurateBearing(): BearingMeasurement? {
```

While `minAccuracy` is technically correct (because the value actually describes the *in*accuracy), I always find it sounds like "least accurate".

If we add this, I'd also suggest we use it as the default in `LocationPuck`. I think that makes the sense for most use cases.

jgillich: We don't have UserLocationState in LocationPuck anymore, and since orientation is off by default, I think it's best to require users to opt-in to this behavior. I added a docs page that shows how to do it

westnordost:
I've begun reviewing this but noticed that there are a few open review comments by @kodebach . Can you check whether all of them are outdated or resolved and mark them respectively here on github?

kodebach:
From my POV everything has been resolved. Github doesn't let me mark the remaining threads as resolved though...

lib/maplibre-compose/src/commonMain/kotlin/org/maplibre/compose/location/Orientation.kt
```diff
@@ -0,0 +1,7 @@
+package org.maplibre.compose.location
+
+import kotlin.time.TimeMark
+import kotlinx.serialization.Serializable
+
+@Serializable
+public data class Orientation(val orientation: BearingMeasurement? = null, val timestamp: TimeMark)
```
westnordost: A Kotlin [TimeMark](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin.time/-time-mark/) is not a timestamp, though. That class was new to me, so I had to read up on it.

A `TimeMark` works like that you mark a time, get a `TimeMark`, and then can ask the `TimeMark` how much time has passed since it was marked. A timestamp would be, what was it, ... milliseconds since 1970-01-01 00:00.0000 or something?

lib/maplibre-compose/src/commonMain/kotlin/org/maplibre/compose/location/Measurements.kt
westnordost, line 1: On the naming, `*Measurement` and thus a `Location` consisting of several `*Measurement`s (Speed, Position, Bearing), I was initially confused by this. I've not seen this naming used on any platform.

You are German, right?

Measurement can mean **Messwert** (i.e. measured value, reading), but _also_ **Maßeinheit** (i.e. measuring unit, measure). The former is meant here, but my association was with the latter.

Also, a `Location` doesn't necessarily *have to* come from some (GNSS) sensor, one could set that programmatically, after all. Or, well, if it always should reflect a measurement, then I guess `Location` would need to be renamed to `LocationMeasurement`, too.

But as I wrote, I would avoid `Measurement`. Maybe just `*WithPrecision`? Or inline the precision values in `Location` again so that `*Measurement` is really just the return type of the `*Provider`s but does not transfer to places where `Location` is used?

Well, what do you think? 🤔

lib/maplibre-compose/src/iosMain/kotlin/org/maplibre/compose/location/IosOrientationProvider.kt
westnordost: What about this? Is it not possible for some reason to implement an orientation provider on iOS right now? If yes, a comment that explains why / links to issues would be good.

lib/maplibre-compose/src/commonMain/kotlin/org/maplibre/compose/location/OrientationProvider.kt
```diff
@@ -0,0 +1,26 @@
...
+import kotlinx.coroutines.flow.MutableStateFlow
+import kotlinx.coroutines.flow.StateFlow
+
+public interface OrientationProvider {
+  public val orientation: StateFlow<Orientation?>
+}
+
+public class NullOrientationProvider : OrientationProvider {
+  public override val orientation: StateFlow<Orientation?> = MutableStateFlow(null)
+}
```
westnordost: This could be an `object`.

```suggestion
object NullOrientationProvider : OrientationProvider {
  public override val orientation: StateFlow<Orientation?> = MutableStateFlow(null)
}
```

And then, `rememberNullOrientationProvider()` is really superfluous, an `object` doesn't need to be `remember`ed. Same goes for the `NullLocationProvider`, actually. I see it was contributed by you earlier.

lib/maplibre-compose/src/commonMain/kotlin/org/maplibre/compose/location/LocationPuck.kt
```diff
@@ -50,35 +50,45 @@ import org.maplibre.compose.util.ClickResult
...
+ * @param onLongClick A [LocationClickHandler] to invoke when the main location indicator dot is
+ *   long-clicked.
  */
 @Composable
 public fun LocationPuck(
   idPrefix: String,
-  locationState: UserLocationState,
+  location: Location?,
   cameraState: CameraState,
+  bearing: BearingMeasurement? = location?.course,
```
westnordost: I don't quite understand why this parameter exists. The `bearing` is part of `location`.

If for some reason the user wants to show a bearing that deviates from the `location.bearing`, he can just do so by passing `location = myLocation.copy(bearing = myCustomBearing)`.

Or, if the location puck actually only uses position + bearing of the `Location` (i.e. not speed), then it may make sense to have these two as parameters, and not location.

lib/maplibre-compose/src/commonMain/kotlin/org/maplibre/compose/location/LocationPuck.kt
westnordost: Can't the bearing accuracy be set to `null` in `BearingMeasurement` if the user doesn't want to show the bearing accuracy? Seems like the more straightforward solution, otherwise one could argue where to draw the line - position unknown vs position disabled, position accuracy unknown vs position accuracy not set, etc.

lib/maplibre-compose/src/commonMain/kotlin/org/maplibre/compose/location/LocationPuck.kt
```diff
@@ -267,21 +280,23 @@ private fun rememberBearingAccuracyPainter(
 }
 
 @Composable
-private fun rememberLocationSource(locationState: UserLocationState): GeoJsonSource {
+private fun rememberLocationSource(
+  location: Location?,
+  bearing: BearingMeasurement?,
```
westnordost: as noted earlier, I find it a bit weird that there is `location.bearing` and also `bearing`

westnordost (commented):
I have some comments mostly about some interface-stuff. I didn't look too closely at those callback flows.

One thing I wonder about, though, is, isn't the orientation already part of the `Location`?

lib/maplibre-compose/src/commonMain/kotlin/org/maplibre/compose/location/Orientation.kt
kodebach: Since the decision to use `TimeMark` was originally mine, I'll answer this

It's a kind of timestamp. It's just relative measured from system boot instead of having a fixed zero-date.

However, it's exactly what we need here. With an absolute timestamp i.e. `Instant` you know when the location was supposedly captured, but you cannot reliably now _how long ago_ that was, which is what we need here. We just want to know how long ago the location was captured, so we know if we should consider it outdated. We don't care what the actual time was.

With an `Instant` you need to check the current time and subtract from the `Instant` you have. However, system time can be changed by the user. The only truly monotonic time source (at least on Android) is `SystemClock.elapsedRealtimeNanos()`. That's why the Android `Location` class provides both `getTime()` and `getElapsedRealtimeNanos()`.


If this is just about the name of the property. IMO the type `TimeMark` makes it clear what the value is. If it was just a `long`, names like in `android.location.Location` would make more sense, yes.

lib/maplibre-compose/src/iosMain/kotlin/org/maplibre/compose/location/IosOrientationProvider.kt
kodebach: I thought this was resolved, because I commented on the wrong commit, but seem it's still open. See https://github.com/maplibre/maplibre-compose/pull/737#discussion_r2754962022 for info on how to implement it

It should be done before the PR is merged IMO

lib/maplibre-compose/src/commonMain/kotlin/org/maplibre/compose/location/LocationPuck.kt
kodebach: Yeah, I'm also wasn't 100% happy with the parameters (see https://github.com/maplibre/maplibre-compose/pull/737#discussion_r2755130154 and https://github.com/maplibre/maplibre-compose/pull/737#discussion_r2754866568).

Having users call `copy` is definitely not the right way to do it. That would allocate a new `Location` object every time just for the `LocationPuck` to split it apart again. Since location and bearing might be changing quite often, this is a potential performance issue. It's also bad for Compose's change tracking, because now the `location` parameter changes when just the `bearing` is updated and everything that depends on `location` needs to recompose instead of just the stuff that uses `bearing`.

IMO the correct way in terms of Compose API philosophy would actually be to split the `location` parameter further into the parts we actually use. But that's kinda annoying to call for users, so we'd need an overload that takes `Location` and splits it and adapts the click handler (without the `location` parameter the current click handler can't be called).


```kotlin
@Composable
public fun LocationPuck(
  idPrefix: String,
  location: Location?,
  cameraState: CameraState,
  oldLocationThreshold: Duration = 30.seconds,
  accuracyThreshold: Float = 50f,
  colors: LocationPuckColors = LocationPuckColors(),
  sizes: LocationPuckSizes = LocationPuckSizes(),
  showBearing: Boolean = true,
  showBearingAccuracy: Boolean = true,
  onClick: LocationClickHandler? = null,
  onLongClick: LocationClickHandler? = null,
)

@Composable
public fun LocationPuck(
  idPrefix: String,
  position: PositionMeasurement?,
  bearing: BearningMeasurement?,
  cameraState: CameraState,
  isOldLocation: Boolen, // alternatively pass the TimeMark and oldLocationThreshold, but that's also a little awkward
  accuracyThreshold: Float = 50f,
  colors: LocationPuckColors = LocationPuckColors(),
  sizes: LocationPuckSizes = LocationPuckSizes(),
  showBearing: Boolean = true,
  showBearingAccuracy: Boolean = true,
  onClick: (() -> Unit)? = null,
  onLongClick: (() -> Unit)? = null,
)
```

That's not significantly better than the current API so I abandoned the idea.

lib/maplibre-compose/src/commonMain/kotlin/org/maplibre/compose/location/LocationPuck.kt
kodebach: > Can't the bearing accuracy be set to `null` in `BearingMeasurement`

That would need a `copy` (see https://github.com/maplibre/maplibre-compose/pull/737#discussion_r3130408069) or a custom provider, both of which seem kinda bad.

lib/maplibre-compose/src/commonMain/kotlin/org/maplibre/compose/location/Orientation.kt
westnordost: maybe it is a kind of timestamp, but I'd argue that a timestamp is generally understood as *the* (unix) timestamp. My suggestion would simply be to rename the value to timeMark or similar.

kodebach: I mean sure, I don't mind if it's renamed, but it seems bit nitpicky.

BTW "timestamp" may imply "specific point in (probably UTC) time", but  certainly doesn't automatically mean Unix timestamp. If timestamp means Unix timestamp (i.e. some size integer counting seconds or milliseconds since 1970-01-01) to you, you haven't dealt with many date/time APIs yet.
For example Apple's `NSDate` uses 2001-01-01 as it's epoch, counting seconds as a `double`. Win32 (in some places) counts 100 nanosecond intervals since 1601-01-01 as a 64 bit integer. Excel counts days from 1900-01-01 as a `double`. GPS doesn't even use the UTC timescale and counts weeks since 1980-01-06.

westnordost: okay

lib/maplibre-compose/src/commonMain/kotlin/org/maplibre/compose/location/LocationPuck.kt
jgillich: There are now two sources for bearing, course (travel direction) and device orientation. One might use either one or a combination of both. As far as I can tell, this is the best way to do it.

lib/maplibre-compose/src/commonMain/kotlin/org/maplibre/compose/location/LocationPuck.kt
jgillich: But there isn't :)

https://github.com/jgillich/maplibre-compose/blob/c9aef7bc60c28676e999add0a5adfa198d7178ab/lib/maplibre-compose/src/commonMain/kotlin/org/maplibre/compose/location/Location.kt#L20

lib/maplibre-compose/src/iosMain/kotlin/org/maplibre/compose/location/IosOrientationProvider.kt
jgillich: I don't have an iOS development environment. But I can see if I can build a smoke-tested version

lib/maplibre-compose/src/commonMain/kotlin/org/maplibre/compose/location/Measurements.kt
jgillich: This naming was originally [proposed](https://github.com/maplibre/maplibre-compose/pull/712#issuecomment-3525343935) by Sargun but it seems fine to me? "Maßeinheit" does not translate to measurement according to [dict](https://m.dict.cc/deutsch-englisch/Ma%C3%9Feinheit.html)

lib/maplibre-compose/src/commonMain/kotlin/org/maplibre/compose/location/OrientationProvider.kt
jgillich: The original location feature was built by @kodebach and I was following its design. This is a class with its own remember simply for cohesion with other providers.

lib/maplibre-compose/src/commonMain/kotlin/org/maplibre/compose/location/Measurements.kt
sargunv: I would consider what you call Maßeinheit to be "unit of measure" or just "measure" but not "measurement"

That said, I do like the suggestion for `PositionWithPrecision` and so on; conveys the point of this type much better than just the noun did.

lib/maplibre-compose/src/commonMain/kotlin/org/maplibre/compose/location/OrientationProvider.kt
sargunv: I think I do prefer object, and so will probably eventually make a breaking change on the Location side to be an object as well (probably bundled with whatever next wave of breaking changes I do). So slight preference to just get it to object here now.

sargunv:
lgtm pending the iOS side and a couple of nits.

I can implement the iOS side myself when I have some time (soon) and then merge.

~~but I'd appreciate it if you can make a first pass (even if frontier-llm generated and not validated) so I only have to fix/validate from there.~~

edit nvm, done!

lib/maplibre-compose/src/commonMain/kotlin/org/maplibre/compose/location/LocationPuck.kt
sargunv: agreed with pretty much all of the above

I'm also not 100% happy with the API but it's reasonable for now

I'm generally not too concerned about short-lived allocations (`.copy()`) though; the Android/JVM GC at least is well optimized for that kind of usage. Kotlin/Native less so but even that is not bad these days.

It's been a while since I profiled it but even on extremely low end android phones I was not able to observe a performance difference from optimizing these kinds of allocations in Composables recomposing every frame. (still there's _maybe_ some tiny energy usage impact)

lib/maplibre-compose/src/commonMain/kotlin/org/maplibre/compose/location/Measurements.kt
westnordost: Hmm okay, then I was wrong.
I got the info from https://dict.leo.org/german-english/measurement

lib/maplibre-compose/src/commonMain/kotlin/org/maplibre/compose/location/LocationPuck.kt
westnordost: Jetpack Compose does this .copy stuff itself pretty frequently, e.g. something like `LocalContentColor.current.copy(alpha = ContentAlpha.medium)` and the like.

westnordost: It's funny how a whole generation of (Java) devs has been brought up to shun copying due to it being described as expensive for the GC (me included) to then be told that it doesn't matter anymore (for local allocations, I presume).

lib/maplibre-compose/src/commonMain/kotlin/org/maplibre/compose/location/Measurements.kt
sargunv: it's entirely possible some other English speakers use "measurement" that way

if I came across it I'd be able to infer the meaning, just feel it's slightly odd, but also that's what British English feels like 🤷

lib/maplibre-compose/src/commonMain/kotlin/org/maplibre/compose/location/OrientationProvider.kt
kodebach: You could do

```kotlin
fun rememberNullOrientationProvider() = NullOrientationProvider
```

still a breaking change because of the `class` -> `object` change, but keeps some backwards compatibility

lib/maplibre-compose/src/commonMain/kotlin/org/maplibre/compose/location/LocationPuck.kt
kodebach: I should've left out the part about allocating new objects. That's (probably) fine.

The bigger issue is definitely that it breaks the change detection of the compose runtime and could cause unnecessary recomposition. Especially with something as frequently changing as a live location/bearing you really want to avoid that.

> Jetpack Compose does this `.copy` stuff itself pretty frequently

True, `Color` is very bad example though. That's a highly optimized `value class` not a `data class`. The `copy` call compiles to a handful of shifts and other integer operations.

Another option would be `LocationPuckState` class with `MutableState` properties for the various parts. Then these could be independently tracked by compose and easily updated. Something like:

```kotlin
@Composable
public fun LocationPuck(
  idPrefix: String,
  state: LocationPuckState,

  // maybe move these 3 also into LocationPuckState
  oldLocationThreshold: Duration,
  accuracyThreshold: Float,
  cameraState: CameraState,

  colors: LocationPuckColors = LocationPuckColors(),
  sizes: LocationPuckSizes = LocationPuckSizes(),
  onClick: LocationClickHandler? = null,
  onLongClick: LocationClickHandler? = null,
)

@Composable
fun rememberLocationPuckState(
  location: Location?,
): LocationPuckState {
  return remember { LocationPuckState(location?.position, location?.course) }
}

class LocationPuckState(
  position: PositionMeasurement?,
  bearing: BearingMeasurement?,
) {
  var position by mutableState(position)
  var bearing by mutableState(bearing)
}
```

EDIT: I realize that's similar to what we had with `UserLocationState` but I'm just mentioning for completeness.

lib/maplibre-compose/src/commonMain/kotlin/org/maplibre/compose/location/Measurements.kt
kodebach: >  I do like the suggestion for `PositionWithPrecision` and so on

If we're discussing semantics, it should be `PositionWithAccuracy` then, since we're stating how close to the true value the position is not how close repeated measurements are to each other.

sargunv (approved):

Issue #753, opened 2026-03-02
test: verify user layer presence in style.getLayers()


sargunv:
> [!NOTE]
> **AI-generated issue.** This issue was created by Claude Code (Opus 4.6). It may contain inaccuracies — please verify before acting on it.

## Summary

Two integration tests were simplified because user-added layers never appeared
in `style.getLayers()` during the test window:

- `loadsEmptyStyleAndAddsLayers` — should verify `test-line` layer is present
- `layersRestoredAfterStyleSwitch` — should verify `user-line` is restored

Both currently only assert `assertNoLogErrors()`. Ideally they would also verify
the layers are actually in the native style.

## Root cause

The child `Composition` created by `rememberStyleComposition` via
`Composition(MapNodeApplier(rootNode), compositionContext)` schedules its initial
apply through the parent `Recomposer`. `ComposeUiTest.waitForIdle()` doesn't
flush these manually-created sub-compositions, so layers are never applied within
the test's observable window.

Debug output showed only `[org.maplibre.annotations.points]` (the SDK's default
annotation layer) — user layers were never present.

Additional constraints:
- `runOnUiThread` inside `waitUntil` causes deadlocks (Espresso sync contention)
- Manual polling with `Thread.sleep` + `waitForIdle` avoids deadlocks but layers
  still don't appear

## Possible approaches

1. **Custom TestRecomposer** that flushes sub-compositions (complex, fragile)
2. **Longer timeouts with frame-based waiting** (may still not work)
3. **Assert on `StyleState.sources` instead** (works — already used in
   `styleStateSources` test — but only verifies sources, not layers)
4. **Accept current `assertNoLogErrors()` coverage** (pragmatic — the key
   regression value is crash/error detection)

## Context

Discovered during development of the integration test harness on the
`sargunv/explore-testing` branch.

Issue #752, opened 2026-03-02
feat: iOS integration test harness


sargunv:
> [!NOTE]
> **AI-generated issue.** This issue was created by Claude Code (Opus 4.6). It may contain inaccuracies — please verify before acting on it.

## Summary

Build an iOS integration test harness equivalent to the Android one, enabling
regression tests to run against the real native MapLibre iOS SDK on a simulator.

The Android integration test harness (on `sargunv/explore-testing`) provides
`withMap` + `MapTestScope` for rendering a real `MaplibreMap`, switching styles,
capturing logs, and asserting on crashes/errors. All 11 regression tests
(6 lifecycle + 5 expression) currently only run on Android.

## Why this is needed

iOS-specific bugs (e.g., #437) can only be caught by tests running on the actual
iOS native SDK. The existing `commonTest` tests use `FakeStyle` and bypass the
native SDK entirely.

## The blocker

`runComposeUiTest` on iOS uses `SkikoComposeUiTest`, an offscreen Skia renderer
that cannot host `UIKitView`. Since MapLibre renders on iOS through `UIKitView`
interop, the map never initializes.

Tracked upstream: [JetBrains compose-multiplatform#4687](https://github.com/JetBrains/compose-multiplatform/issues/4687)

## Possible approaches

1. **Native XCTest harness with Compose interop** — Swift/Obj-C test code
   bootstraps a real `UIWindow` + `UIViewController` hosting the Compose scene
2. **Custom Kotlin/Native test runner** — Create a `UIWindow` in Kotlin test
   code with manual Compose scene setup
3. **Wait for JetBrains** — Upstream fix to `SkikoComposeUiTest` (unknown timeline)

## Additional requirements

- `CapturingLogWriter` needs a KMP-compatible replacement (currently uses
  `java.util.Collections.synchronizedList`; iOS would need `kotlinx.atomicfu`
  or `Mutex`-guarded list)
- iOS native log routing goes through
  `MLNLoggingConfiguration.sharedConfiguration.setHandler()` (different from
  Android's `MLNLogger.setLoggerDefinition()`)
- CI infrastructure already runs `iosSimulatorArm64Test`

## Acceptance criteria

- [ ] iOS test harness can render a real `MaplibreMap` with `UIKitView`
- [ ] `withMap` equivalent works on iOS simulator
- [ ] At least one regression test passes on iOS
- [ ] `CapturingLogWriter` works on iOS (KMP-compatible thread safety)
[...]

Issue #751, opened 2026-03-02
test: verify onMapLoadFinished callback fires (#315)


sargunv:
> [!NOTE]
> **AI-generated issue.** This issue was created by Claude Code (Opus 4.6). It may contain inaccuracies — please verify before acting on it.

## Summary

Add an integration test explicitly verifying that the `onMapLoadFinished`
callback on `MaplibreMap` fires after style loading completes.

Historical bug #315 involved initialization ordering issues where
`onMapLoadFinished` did not fire or fired at the wrong time.

## What the test needs

- Set `onMapLoadFinished` callback on `MaplibreMap`
- Verify callback fires within timeout
- Assert on the callback having been invoked

## Notes

The current test harness implicitly validates style loading works (via `withMap`
calling `waitForStyleLoad()` which checks `StyleState`), but doesn't explicitly
test the `onMapLoadFinished` callback path. These may diverge in the future.

This was listed in the brainstorm but dropped during plan scoping. It's a
straightforward addition to `StyleLifecycleTest`.

## Context

Part of the integration test harness started in the `sargunv/explore-testing`
branch.

Reproduces: #315

Issue #750, opened 2026-03-02
test: programmatic vector source with null options on iOS (#437)


sargunv:
> [!NOTE]
> **AI-generated issue.** This issue was created by Claude Code (Opus 4.6). It may contain inaccuracies — please verify before acting on it.

## Summary

Add an integration test verifying that creating a programmatic vector source
with default/null options does not crash on iOS.

Historical bug #437 showed a crash when the native SDK expected non-null options
but the Kotlin binding passed null.

## What the test needs

- Create a programmatic vector source with default options
- Verify no crash on iOS simulator
- Run in the `iosSimulatorArm64Test` CI job

## Why it's deferred

Blocked on iOS integration test infrastructure. `runComposeUiTest` on iOS uses
`SkikoComposeUiTest` (offscreen Skia renderer) which cannot host `UIKitView`.
Since MapLibre renders on iOS through `UIKitView` interop, the map never
initializes in the Compose test framework.

An iOS integration test harness that creates a real Compose scene with `UIKitView`
interop is a prerequisite. See the iOS test harness issue.

## Context

iOS-only test. Depends on an iOS integration test harness being built first.

Reproduces: #437
Related: [JetBrains compose-multiplatform#4687](https://github.com/JetBrains/compose-multiplatform/issues/4687)

PR #754, opened 2026-03-02
feat(testing): add Android integration tests for style lifecycle and expression regressions

sargunv:
> [!NOTE]
> **AI-assisted PR.** This PR was generated with Claude Code (Opus 4.6). See the AI Planning Documents section for the brainstorm and plan that guided implementation.

## Summary

- Add an Android integration test harness (`withMap` + `MapTestScope`) that renders a real `MaplibreMap` composable against the real native MapLibre SDK on an emulator
- Write 11 regression tests covering the two historically dominant bug categories: style lifecycle crashes (6 tests) and expression compilation failures (5 tests)
- Fix a real bug discovered during testing: `rememberStyleComposition` didn't call `style.unload()` before `composition.dispose()`, causing `CalledFromWorkerThreadException` during teardown
- Remove stale `warnIfUnloaded` println noise from `SafeStyle`

## Bug fix: `rememberStyleComposition` teardown ordering

During test teardown, `composition.dispose()` runs on the coroutine dispatcher (not the UI thread), triggering `CalledFromWorkerThreadException` from native SDK calls in `MapNodeApplier.onClear()`. Fixed by calling `style.unload()` before `composition.dispose()` in the `finally` block, making `SafeStyle` short-circuit all native calls during disposal.

## Tests added

**Style lifecycle regression tests** (`StyleLifecycleTest`):

| Test | Reproduces |
|------|-----------|
| `loadsEmptyStyleAndAddsLayers` | General |
| `switchStyleWithoutCrash` | #244, #252 |
| `rapidStyleSwitchWithoutCrash` | #269, #336 |
| `layersRestoredAfterStyleSwitch` | #506, #516 |
| `getBaseSourceReturnsNull` | #501 |
| `styleStateSources` | #353 |

**Expression regression tests** (`ExpressionRegressionTest`):

| Test | Reproduces |
|------|-----------|
| `switchExpressionMultipleCases` | #310, #327 |
| `switchExpressionZeroCases` | #639 |
| `textSizeInterpolation` | #555, #573 |

## Test plan

- [x] All 11 new tests pass on Android emulator (API 34)
- [x] Existing desktop host tests pass (no regressions)
- [x] Compilation succeeds
- [x] CI passes (`connectedDebugAndroidTest` on API 35 + API 24)
- [x] Manually test style changes with demo app

## Future work

- #748 — layer click dispatch order test
- #749 — camera state survives config change test
- #750 — vector source null options on iOS
- #751 — `onMapLoadFinished` callback test
- #752 — iOS integration test harness
- #753 — user layer presence verification in `getLayers()`

---

## AI Planning Documents

These documents guided the implementation of this PR.

<details>
<summary>Brainstorm: Integration Testing for MapLibre Compose</summary>

# Integration Testing for MapLibre Compose

**Date:** 2026-03-01 **Status:** Brainstorm

## What We're Building

A test harness that runs the real `MaplibreMap` composable against real native
MapLibre SDKs on Android (emulator) and iOS (simulator), focused on catching the
two historically dominant bug categories:

1. **Style lifecycle crashes** — race conditions between async native style
   loading and Compose composition (9 bugs, 6 crashes historically)
2. **Expression compilation failures** — expressions that the native SDK rejects
   or silently mishandles, only detectable on a live map (11 bugs, 3 crashes
   historically)

Tests live in the library module (`androidDeviceTest` / `iosTest`) with access
to internal APIs. CI infrastructure for emulators and simulators already exists.

## Why This Approach

**Full vertical with real `MaplibreMap` composable, not just `StyleNode` +
`FakeStyle`.**

The existing tests bypass the Compose composition pipeline entirely — they drive
`StyleNode` imperatively with a `FakeStyle`. This misses the exact class of bugs
that has caused the most crashes: initialization ordering, `SafeStyle`
invalidation during style switches, recomposition instability from missing
`@Immutable` annotations, and `MapNodeApplier` lifecycle issues.

Approach B (real native `Style` without the composable) was considered but
rejected because it can't test style switching, initialization ordering, or
recomposition-driven bugs.

### Test location: library module (`androidDeviceTest` / `iosTest`)

Tests need access to internal APIs like `Style`, `StyleNode`, `SafeStyle`, and
the native log interception already wired through the `logger` parameter. The
existing CI jobs (`connectedDebugAndroidTest`, `iosSimulatorArm64Test`) already
run these source sets.

### Assertion strategy: crashes + state correctness + log capture

- **Crashes**: test completes without exception
- **State correctness**: after operations, query the native style via the
  internal `Style` interface for expected layers (ordering, presence) and
  sources. `StyleState` only exposes `sources`; layer-ordering assertions
  (critical for #506, #516) require direct `Style` access, which is available
  since tests live in the library module.
- **Log capture**: pass a Kermit `Logger` with a custom `LogWriter` to
  `MaplibreMap` that captures warnings/errors from MapLibre Native. Assert no
  errors logged. This catches the class of bugs where the native SDK logs
  "Invalid property value" without throwing (e.g., #639, #555).

On Android, native logs route through `MLNLogger.setLoggerDefinition()` →
`KermitLoggerDefinition`. On iOS, native logs route through
`MLNLoggingConfiguration.sharedConfiguration.setHandler()`. Both are already
wired to the composable's `logger` parameter.

### Async handling: `waitUntil` from Compose test framework

The native map loads styles asynchronously. Tests use `waitUntil { condition }`
(polls a predicate with timeout) to wait for style load completion, camera
animations, etc. This works with native async events — it doesn't depend on
Compose's test clock.

### Test sharing: maximize common logic, let the harness design drive it

Follow the existing pattern: abstract base class in `commonTest` with test
logic, thin platform subclasses in `androidDeviceTest` / `iosTest` for
initialization. If platform differences require divergent test bodies, allow it
— but start with shared logic.

### Base style: `BaseStyle.Empty` for deterministic tests

Use the existing `BaseStyle.Empty` (minimal empty style JSON) to avoid network
requests and provide a deterministic starting point. Tests that need specific
base layers/sources can use inline JSON styles.

### Expression tests: regression-first, not exhaustive

Start with regression tests for each historical expression bug, not
one-per-category. Each test applies the specific problematic expression to a
real layer on a live map and verifies no crash/error. Expand coverage later.

### Common test infrastructure (in `commonTest`)

A helper that:

1. Renders `MaplibreMap` with a given `baseStyle` and `content` lambda
2. Passes a capturing `Logger` that collects warnings/errors
3. Exposes the `StyleState` for assertions
4. Provides `waitUntil`-based helpers: `waitForStyleLoad()`,
   `waitForLayerPresent(id)`, etc.
5. Provides a way to switch the base style mid-test and wait for reload

### Style lifecycle regression tests

| Test                                                        | Reproduces | What it verifies                                  |
| ----------------------------------------------------------- | ---------- | ------------------------------------------------- |
| Load empty style, add layers/sources, verify present        | General    | Basic composition → native style pipeline works   |
| Switch style at runtime, verify no crash                    | #244, #252 | `SafeStyle` handles style invalidation correctly  |
| Switch style rapidly N times, verify no crash               | #269, #336 | Race condition between style load and composition |
| Add layers, switch style, verify layers restored            | #506, #516 | Recomposition stability across style changes      |
| Verify `onMapLoadFinished` fires                            | #315       | Initialization ordering is correct                |
| Verify `getBaseSource` returns nullable after style load    | #501       | No crash on base source access                    |
| Verify `StyleState.sources` updates when user sources added | #353       | Style state reflects programmatic sources         |

### Expression regression tests

| Test                                             | Reproduces | Expression applied to real layer                                   |
| ------------------------------------------------ | ---------- | ------------------------------------------------------------------ |
| `switch` with multiple conditions in `iconImage` | #310, #327 | `switch(feature.get("type")) { ... }` with 3+ cases on SymbolLayer |
| `switch` with zero cases (only fallback)         | #639       | `switch(feature.get("x")) { fallback(const("y")) }`                |
| `textSize` with SP-unit interpolation at zoom    | #555, #573 | `interpolate(zoom, ...).sp` on SymbolLayer `textSize`              |
| `iconPadding` with correct DpPadding type        | #471       | `const(PaddingValues(4.dp))` on SymbolLayer `iconPadding`          |
| `lineProgress` with no parameters                | #679       | `lineProgress` expression on LineLayer `lineGradient`              |

### Future regression tests (out of scope for spike)

These are real historical bugs worth testing but significantly more complex to
set up. Defer until the core harness is proven.

| Test                                                         | Reproduces | Complexity                                            |
| ------------------------------------------------------------ | ---------- | ----------------------------------------------------- |
| Layer click dispatched in style order, not composition order | #69, #93   | Needs real GeoJSON data + simulated click coordinates |
| Camera state survives configuration change (Android)         | #261, #337 | Needs activity recreation in test                     |
| Programmatic vector source with null options (iOS)           | #437       | iOS-only, needs platform-specific test body           |

## Resolved Questions

1. **`BaseStyle.Empty` behavior** — Confirmed: `BaseStyle.Empty` is valid style
   JSON and fires the full style-loaded callback chain on both platforms.

## Open Questions (Spike Required)

The following questions block committing to this approach. A spike should answer
them by writing a minimal test on each platform that renders a real
`MaplibreMap` with `BaseStyle.Empty`, waits for style load, and asserts on
something trivial.

1. **iOS Compose UI testing with `UIKitView` interop** — Does `runComposeUiTest`
   on iOS actually work when the composition contains a `UIKitView` (which is
   how MapLibre renders on iOS)? The existing `IosStyleNodeTest` never renders a
   real `UIKitView`. This is the biggest unknown.

2. **Android `runAndroidComposeUiTest` with native `MapView`** — Similarly, does
   the Compose test framework handle `AndroidView` wrapping a `MapView`
   correctly? The test activity needs to host the composable and the map needs
   to actually initialize. The existing `AndroidStyleNodeTest` calls
   `MapLibre.getInstance()` but never renders a map.

3. **Test execution time** — How long does a single test that initializes a real
   map, loads a style, and tears down take? If it's 5+ seconds per test, one map
   per test class (shared state) may be preferable to one map per test
   (isolated). The spike should measure this.

</details>

<details>
<summary>Plan: Integration Test Harness for Live Map on Android</summary>

---
title: "feat: Integration test harness for live map on Android"
type: feat
status: active
date: 2026-03-01
origin: docs/brainstorms/2026-03-01-integration-testing-brainstorm.md
---

# Integration Test Harness for Live Map on Android

## Overview

Build a test harness that renders the real `MaplibreMap` composable against the
real native MapLibre Android SDK on an emulator, then use it to write regression
tests for the two historically dominant bug categories: style lifecycle crashes
and expression compilation failures.

iOS is deferred — `runComposeUiTest` on iOS uses an offscreen Skia renderer that
cannot host `UIKitView`. iOS integration testing will require a native test
harness that spins up a real Compose scene with `UIKitView` interop, which is a
separate effort.

## Problem Statement

The library has had 20+ bugs and 9+ crashes in style lifecycle and expression
compilation. The existing tests bypass the Compose composition pipeline entirely
(driving `StyleNode` imperatively with `FakeStyle`), missing the exact class of
bugs that causes the most crashes: initialization ordering, `SafeStyle`
invalidation during style switches, recomposition instability, and expression
failures that only surface on a live native map.

(See brainstorm:
`docs/brainstorms/2026-03-01-integration-testing-brainstorm.md`)

## Proposed Solution

Two phases: spike to validate the approach, then harness + regression tests.

### Phase 1: Spike — Validate `MaplibreMap` in Android instrumented tests

**Goal:** Confirm `runAndroidComposeUiTest` works with a real `MaplibreMap` and
measure timing before building the full harness.

`runAndroidComposeUiTest<ComponentActivity>` runs a real `ComponentActivity`
with full lifecycle (progresses to RESUMED). It supports `AndroidView` natively.

```kotlin
// lib/maplibre-compose/src/androidDeviceTest/
@OptIn(ExperimentalTestApi::class)
class AndroidMapSpikeTest {
  @Test
  fun mapLoadsEmptyStyle() = runAndroidComposeUiTest<ComponentActivity> {
    var loaded = false
    setContent {
      MaplibreMap(
        baseStyle = BaseStyle.Empty,
        onMapLoadFinished = { loaded = true },
      )
    }
    waitUntil(10_000) { loaded }
    assert(loaded)
  }
}
```

**What to measure:** Wall-clock time for map init + style load on the CI
emulator. This determines one-map-per-test vs one-map-per-class.

**Run the spike in CI** to establish realistic timeout baselines. Local timing
is not representative of GitHub Actions emulator performance.

#### 2a. Expose `Style` for layer assertions

`StyleState` only exposes `sources: Map<String, Source>`. Layer-ordering
assertions (critical for bugs #506, #516) need `Style.getLayers()`.

**Change:** In `StyleState.kt`, change `styleNode` from `private` to `internal`:

```kotlin
internal var styleNode: StyleNode? // was: private var styleNode: StyleNode?
```

This lets tests access `styleState.styleNode?.style?.getLayers()`. Minimal
change, no public API impact. Consistent with the existing pattern where
`StyleNode`, `SourceManager`, `LayerManager`, and `SafeStyle` are all already
`internal`.

#### 2b. Capturing log writer

A custom Kermit `LogWriter` that collects log entries for assertion.

```kotlin
// In commonTest
internal class CapturingLogWriter : LogWriter() {
  data class LogEntry(
    val severity: Severity,
    val message: String,
    val tag: String,
    val throwable: Throwable?,
  )

  private val _logs = Collections.synchronizedList(mutableListOf<LogEntry>())
  val logs: List<LogEntry> get() = _logs.toList()

  override fun log(severity: Severity, message: String, tag: String, throwable: Throwable?) {
    _logs.add(LogEntry(severity, message, tag, throwable))
  }

  fun errors(): List<LogEntry> = logs.filter { it.severity >= Severity.Error }
}
```

**Thread safety:** `KermitLoggerDefinition` on Android may be called from native
SDK background threads. The `_logs` list uses `Collections.synchronizedList` to
prevent data races between the native SDK logging thread and the test thread
reading `errors()`.

**Assertion strategy:** Assert on `errors()` only (`Severity.Error` and above).
A fresh `CapturingLogWriter` is created per test via `withMap`, so no cross-test
contamination.

**Lifecycle:** Each test's map initialization calls
`MLNLogger.setLoggerDefinition()`, overwriting the global handler. Tests run
sequentially within a class, so this is safe.

#### 2c. Test harness helper

```kotlin
// In androidDeviceTest (or commonTest if it compiles there)
internal class MapTestScope(
  val styleState: StyleState,
  val logWriter: CapturingLogWriter,
  private val baseStyleState: MutableState<BaseStyle>,
  private val composeUiTest: ComposeUiTest,
) {
  val style: Style?
    get() = styleState.styleNode?.style

  fun switchStyle(newStyle: BaseStyle) {
    composeUiTest.runOnUiThread { baseStyleState.value = newStyle }
  }

  fun waitForStyleLoad(timeoutMillis: Long = 10_000) {
    composeUiTest.waitUntil(timeoutMillis) { style != null }
  }

  fun waitForStyleReload(timeoutMillis: Long = 10_000) {
    // After a style switch, the old style is torn down (style becomes null)
    // before the new style loads (style becomes non-null again).
    composeUiTest.waitUntil(timeoutMillis / 2) { style == null }
    composeUiTest.waitUntil(timeoutMillis / 2) { style != null }
  }

  fun assertNoLogErrors() {
    val errors = logWriter.errors()
    if (errors.isNotEmpty()) {
      error(
        "MapLibre logged ${errors.size} error(s):\n" +
          errors.joinToString("\n") { "  [${it.severity}] ${it.message}" }
      )
    }
  }
}

@OptIn(ExperimentalTestApi::class)
internal fun ComposeUiTest.withMap(
  initialStyle: BaseStyle = BaseStyle.Empty,
  content: @Composable @MaplibreComposable () -> Unit = {},
  block: MapTestScope.() -> Unit,
) {
  val logWriter = CapturingLogWriter()
  val logger = Logger(loggerConfigInit(logWriter))
  val styleState = StyleState()
  val baseStyleState = mutableStateOf(initialStyle)

  setContent {
    val currentStyle by baseStyleState
    MaplibreMap(
      baseStyle = currentStyle,
      styleState = styleState,
      logger = logger,
      content = content,
    )
  }

  val scope = MapTestScope(styleState, logWriter, baseStyleState, this)
  scope.waitForStyleLoad()
  scope.block()
}
```

#### Regression tests

**Style lifecycle regression tests:**

| #   | Test method                      | Reproduces | What it does                                                                                            |
| --- | -------------------------------- | ---------- | ------------------------------------------------------------------------------------------------------- |
| 1   | `loadsEmptyStyleAndAddsLayers`   | General    | Load `BaseStyle.Empty`, add `LineLayer` via content, verify layer present in `style.getLayers()`        |
| 2   | `switchStyleWithoutCrash`        | #244, #252 | Load empty style, switch to different inline JSON style, `waitForStyleReload()`, assert no crash/errors |
| 3   | `rapidStyleSwitchWithoutCrash`   | #269, #336 | Switch base style 10 times in a loop, `waitForStyleLoad()` at end, assert no crash                      |
| 4   | `layersRestoredAfterStyleSwitch` | #506, #516 | Add layers, switch style, `waitForStyleReload()`, verify user layers present                            |
| 5   | `getBaseSourceReturnsNull`       | #501       | Call `getBaseSource("nonexistent")` inside composition, verify returns null                             |
| 6   | `styleStateSources`              | #353       | Add `GeoJsonSource` via content, verify `styleState.sources` contains it                                |

**Expression regression tests:**

| #   | Test method                     | Reproduces | Layer + Expression                                                                                                      |
| --- | ------------------------------- | ---------- | ----------------------------------------------------------------------------------------------------------------------- |
| 1   | `switchExpressionMultipleCases` | #310, #327 | `SymbolLayer` with `iconImage = switch(feature.get("type")) { ... }` with 3+ cases                                      |
| 2   | `switchExpressionZeroCases`     | #639       | `SymbolLayer` with `iconImage = switch(feature.get("type")) { fallback(const("icon-default")) }`                        |
| 3   | `textSizeInterpolation`         | #555, #573 | `SymbolLayer` with `textSize = interpolate(linear(), zoom, 10 to const(12f), 20 to const(24f))`                         |
| 4   | `iconPaddingDpType`             | #471       | `SymbolLayer` with `iconPadding = const(PaddingValues(4.dp))`                                                           |
| 5   | `lineProgressExpression`        | #679       | `LineLayer` with `lineGradient = interpolate(linear(), lineProgress, ...)` on `GeoJsonSource` with `lineMetrics = true` |

**Style switching detail (tests 2-4):** Two distinct `BaseStyle.Json` values
with different `name` fields. Test 3 alternates between them 10 times and
asserts on final state.

**Source requirements:** Tests 1-4 use `GeoJsonSource` with empty
`FeatureCollection`. Test 5 needs `GeoJsonOptions(lineMetrics = true)`.
Historical bugs were in expression parsing/compilation, not feature evaluation —
empty data is sufficient.

**Assertion:** Each test calls `assertNoLogErrors()`. Historical bugs manifested
as either crashes (caught by test framework) or native SDK error logs like
"Invalid property value" (caught by log capture).

## Acceptance Criteria

- [ ] Spike passes on Android emulator in CI
- [x] `CapturingLogWriter` captures native MapLibre SDK errors
- [x] `StyleState.styleNode` exposed as `internal` for layer assertions
- [x] `withMap` helper works with style switching, log capture, and async
      waiting
- [ ] 6 style lifecycle regression tests pass
- [ ] 5 expression regression tests pass
- [ ] Tests run in existing `connectedDebugAndroidTest` CI job
- [x] Tests use `BaseStyle.Empty` or inline JSON — no network requests

## Dependencies & Risks

**Risk: Test flakiness from timing.** Native map initialization is async and
environment-dependent. Mitigation: 10s default timeouts, measured against CI
performance in the spike.

**Risk: Thread safety in log capture.** Native log handlers fire from SDK
threads. Mitigation: `Collections.synchronizedList` in `CapturingLogWriter`.

**Risk: Global log handler state.** `MLNLogger.setLoggerDefinition()` is a
global singleton. Mitigation: each test's map initialization overwrites the
handler; tests run sequentially. Do not run integration tests in parallel.

**Risk: `waitForStyleReload` null-transition window.** During a style switch,
the old style is torn down before the new one loads. If the null window is too
brief, `waitUntil { style == null }` could miss it. Mitigation: native style
loading is async and takes measurable time. Fall back to a generation counter if
needed.

## Future Work

**iOS integration tests:** Requires a native test harness that creates a real
Compose scene with `UIKitView` interop, bypassing `SkikoComposeUiTest` (which
uses an offscreen Skia renderer incompatible with native views). This is a
separate effort.

## Sources & References

- **Origin brainstorm:**
  [`docs/brainstorms/2026-03-01-integration-testing-brainstorm.md`](docs/brainstorms/2026-03-01-integration-testing-brainstorm.md)
- Existing test pattern:
  `lib/maplibre-compose/src/commonTest/kotlin/org/maplibre/compose/style/StyleNodeTest.kt`
- Style interface:
  `lib/maplibre-compose/src/commonMain/kotlin/org/maplibre/compose/style/Style.kt`
- StyleState:
  `lib/maplibre-compose/src/commonMain/kotlin/org/maplibre/compose/style/StyleState.kt`
- MaplibreMap composable:
  `lib/maplibre-compose/src/commonMain/kotlin/org/maplibre/compose/map/MaplibreMap.kt`
- BaseStyle.Empty:
  `lib/maplibre-compose/src/commonMain/kotlin/org/maplibre/compose/style/BaseStyle.kt`
- Kermit LogWriter API: `co.touchlab:kermit:2.0.8` —
  `loggerConfigInit(vararg LogWriter)`
- Kermit log routing:
  `lib/maplibre-compose/src/androidMain/kotlin/org/maplibre/compose/util/KermitLoggerDefinition.kt`
- Compose Multiplatform iOS testing limitation: `SkikoComposeUiTest` uses
  offscreen Skia, no native view host
  ([JetBrains #4687](https://github.com/JetBrains/compose-multiplatform/issues/4687))
- Historical bugs: #244, #252, #269, #310, #315, #327, #336, #353, #471, #501,
  #506, #516, #555, #573, #639, #679

</details>


lib/maplibre-compose/src/androidDeviceTest/kotlin/org/maplibre/compose/map/ExpressionRegressionTest.kt
```diff
@@ -0,0 +1,169 @@
...
+        assertNoLogErrors()
+      }
+    }
+
+  /**
+   * Regression for #471: iconPadding with DpPadding type. The native SDK previously failed when
+   * iconPadding used PaddingValues with dp units.
+   */
+  @Test
+  fun iconPaddingDpType() =
```
sargunv: hallucination: remove iconPaddingDpType

lib/maplibre-compose/src/androidDeviceTest/kotlin/org/maplibre/compose/map/ExpressionRegressionTest.kt
```diff
@@ -0,0 +1,169 @@
...
+      }
+    }
+
+  /**
+   * Regression for #679: lineProgress expression on line gradient. Requires a GeoJsonSource with
+   * lineMetrics enabled. The native SDK previously failed when lineProgress was used in a gradient
+   * interpolation.
+   */
+  @Test
+  fun lineProgressExpression() =
```
sargunv: hallucination: remove lineProgressExpression

lib/maplibre-compose/src/androidDeviceTest/kotlin/org/maplibre/compose/map/StyleLifecycleTest.kt
```diff
@@ -0,0 +1,112 @@
...
+  /** Switch base style 10 times rapidly, assert no crash at the end. */
+  @Test
+  fun rapidStyleSwitchWithoutCrash() =
+    runAndroidComposeUiTest<ComponentActivity> {
+      withMap(initialStyle = styleA) {
+        repeat(10) { i -> switchStyle(if (i % 2 == 0) styleB else styleA) }
+        waitForStyleLoad(15_000)
+        assertNoLogErrors()
+      }
+    }
```
sargunv, line 61: maybe too rapid, pause at least a tick or two between switchStyle calls?

lib/maplibre-compose/src/androidDeviceTest/kotlin/org/maplibre/compose/map/CapturingLogWriter.kt
```diff
@@ -0,0 +1,18 @@
...
+
+  private val _logs = Collections.synchronizedList(mutableListOf<LogEntry>())
+
+  override fun log(severity: Severity, message: String, tag: String, throwable: Throwable?) {
+    _logs.add(LogEntry(severity, message))
+  }
+
+  fun errors(): List<LogEntry> =
+    synchronized(_logs) { _logs.filter { it.severity >= Severity.Error } }
+}
```
sargunv, line 18: unsure if this will catch all maplibre-native logs; we might need a different assertion

lib/maplibre-compose/src/androidDeviceTest/kotlin/org/maplibre/compose/map/MapTestScope.kt
```diff
@@ -0,0 +1,90 @@
...
+  }
+
+  fun waitForStyleReload(timeoutMillis: Long = 10_000) {
+    val before = style
+    composeUiTest.waitUntil(timeoutMillis = timeoutMillis) { style != null && style !== before }
+  }
+
+  fun waitUntil(timeoutMillis: Long = 10_000, condition: () -> Boolean) {
+    composeUiTest.waitUntil(timeoutMillis = timeoutMillis, condition = condition)
+  }
```
sargunv, line 44: should see if we can use map callbacks instead of polling loops

lib/maplibre-compose/src/androidDeviceTest/kotlin/org/maplibre/compose/map/ExpressionRegressionTest.kt
westnordost: by the way, does Claude Code process PR comments autonomously, do you have to tell it to process the review suggestions or do you have to clean it up "by hand"?

Issue #738, opened 2026-01-06
Implement synchronous updates for GeoJSON sources


michalgwo:
Recently, synchronous updates for GeoJSON sources were implemented in maplibre-native https://github.com/maplibre/maplibre-native/pull/3968. It would be nice to have it in maplibre-compose.

Blocked by #551 as it requires maplibre-native ios-v6.22.0

sargunv:
This would be great for the location puck.

Unfortunately it's been a few months since any activity from Jetbrains on https://youtrack.jetbrains.com/issue/CMP-8882/ so the iOS blocker is pretty solidly blocked. (nothing stops us from implementing on android though)

klemensz:
This looks very relevant for the current Ferrostar migration to `maplibre-compose`:

- [stadiamaps/ferrostar#853](https://github.com/stadiamaps/ferrostar/pull/853)

The current state I could reach in the migration looks very similar to the screen recordings in this issue, especially for a navigation puck rendered via GeoJSON/layers at high zoom.

https://github.com/user-attachments/assets/2e9a64a1-4717-49f7-b42a-66e17d0eb19c

So it seems very plausible that synchronous GeoJSON source updates would improve the current result significantly.

That said, I am not sure this fully replaces [#760](https://github.com/maplibre/maplibre-compose/issues/760).

My current understanding is:

- `#738` would help the existing Compose-/layer-based approach
- `#760` would still be the cleaner long-term solution for navigation, because it would allow feeding an app-provided tracked location into the native tracking / native puck pipeline

So from the Ferrostar side, `#738` looks immediately relevant, and `#760` may still remain useful as the more robust architectural solution.

If this direction makes sense and we can live with an Android-only implementation for now due to the CMP blocker, I would also be happy to help with the implementation or try to put together a PR.

sargunv:
I'm okay with an android only implementation for now

sargunv:
no longer blocked on ios

sargunv:
#834 adds this (and a lot more) on desktop

sargunv:
now wired up for ios too.

PR #761, opened 2026-04-03
Implement synchronous updates for GeoJSON sources

klemensz:
## Description

Add opt-in synchronous GeoJSON source updates via `GeoJsonOptions.synchronousUpdate` (#738).

This currently has an effect only on Android. Other platforms accept the option but ignore it for now.

The change:
- adds `synchronousUpdate` to the common `GeoJsonOptions` API
- wires the option through to the Android MapLibre SDK
- documents the intended use for small, frequently updated in-memory GeoJSON
  sources and the Android-only scope
- adds common tests and Android device tests
- adds a docs snippet and guide entry
- adds a new SynchronousGeoJsonUpdatesDemo to the demo app

This also bumps `org.maplibre.gl:android-sdk` from `12.0.1` to `12.3.1`, because the required Android SDK support is not available in `12.0.1`.

## Test plan

Ran:
- `GRADLE_USER_HOME=./.gradle ./gradlew :lib:maplibre-compose:allTests --no-daemon`
- `GRADLE_USER_HOME=./.gradle ./gradlew :lib:maplibre-compose:connectedAndroidTest -Pandroid.testInstrumentationRunnerArguments.class=org.maplibre.compose.sources.AndroidGeoJsonSourceTest --no-daemon`

The Android device test covers:
- synchronous initial render from GeoJSON source data
- synchronous update-then-render after `setData(...)`

## Checklist

No. The new behavior is opt-in via `GeoJsonOptions.synchronousUpdate`, which defaults to `false`.

- Android: Android 14 + 16

## Screen recording (Demo app)

Captured on a real device (Samsung Galaxy A52s, Android 14)

https://github.com/user-attachments/assets/a4f7fd01-48bc-44f8-9a66-88e1be855124


klemensz:
PR is ready for review: https://github.com/maplibre/maplibre-compose/pull/761

The actual change is just a couple of lines. Most of the code changes are related to tests, documentation, demo app, etc.

lib/maplibre-compose/src/commonTest/kotlin/org/maplibre/compose/sources/GeoJsonOptionsTest.kt
westnordost, line 1: This test is a bit moot, in my opinion.

The second test just tests that Kotlin data classes work correctly. I hope we can assume that they always work correctly! The first test tests something that doesn't have any logic attached (default parameter value).
IMO neither of the two need testing. But well, in the end, there are different philosophies what should be tested. In this case, it would make sense to align oneself with what Sargun would consider test-worthy (which I myself didn't actually check).

demo-app/src/androidMain/kotlin/org/maplibre/compose/demoapp/demos/SynchronousGeoJsonUpdatesDemoPreview.android.kt
westnordost, line 1: 1. Why is this in `androidMain`?
2. I don't quite see the sense in having `@Preview`s when the whole demo app is basically a preview. It looks like there are no other `@Preview`s anywhere else in the demo app, so it would be consistent to not start with it here.

lib/maplibre-compose/src/androidDeviceTest/kotlin/org/maplibre/compose/sources/AndroidGeoJsonSourceTest.kt
westnordost, line 1: I don't quite get why you think it is necessary to test it. As you wrote, the only thing we are doing in Maplibre-compose is to pass on that parameter to the native SDK. Maplibre-compose doesn't seem to be the right place to verify Maplibre-native behavior (that should be done in tests for Maplibre-native).

Plus, the test with all the setup is quite unwieldy, whilst it would already immediately be visible through the demo you added if the feature didn't render.

demo-app/src/commonMain/kotlin/org/maplibre/compose/demoapp/demos/Demo.kt
```diff
@@ -19,5 +19,7 @@ interface Demo {
 
   @MaplibreComposable @Composable fun MapContent(state: DemoState, isOpen: Boolean) {}
 
+  @UiComposable @Composable fun MapOverlayContent(state: DemoState, isOpen: Boolean) {}
+
```
westnordost, line 23: So you are using this to additionally display zoom buttons for that demos specifically.

Apparently, any demo that is currently active could add an overlay? That potentially leads to conflicts between different demos (overlapping elements). So I wonder: Do you really need this? And if you do, maybe a zoom button could just be visible for all?
Requesting input from @sargunv .

westnordost (commented):
I reviewed and commented everything except `SynchronousGeoJsonUpdatesDemo` because I don't know how @sargunv envisions how the demos should be structured. (But skimming over it, it LGTM.)

Hmm, as you say, the actual change is just a few lines of code.  But then, why did you only implement it for Android? It's just passing a parameter to the iOS SDK, isn't it? I feel that you have to document in documentation and doc comments that it only works in Android might have been more work than actually implementing it.

In general, I have the impression that through all those added unit tests, instrumented tests and quite elaborate demo (for passing a parameter to the native sdk, that is), the PR feels somewhat blown out of proportion for me. Did you write all that code by hand?

klemensz:
> But then, why did you only implement it for Android?

Because it requires a certain maplibre-native ios version that cannot be integrated currently due to some blocking issue, see https://github.com/maplibre/maplibre-compose/issues/738#issuecomment-3716900291

demo-app/src/commonMain/kotlin/org/maplibre/compose/demoapp/demos/Demo.kt
klemensz: For this specific demo, zoom buttons are valuable from my point of view because if you pinch-zoom instead then you automatically leave the camera follow mode.

I could have added them to the sheet (which I did initially) but I didn't like the UX and it didn't help testing the new feature as I needed it. 

Generally I think having zoom buttons adds value to the demo app, so like you said, it could be an option to add them generally to all demos.

MapContent is a `@MaplibreComposable` and not a `@UiComposable` so I haven't found a way to add Compose based overlays without something like this. And I didn't want to add buttons as map layers :) But I'm open for suggestions if this can be implemented in another way.

westnordost:
> Because it requires a certain maplibre-native ios version that cannot be integrated currently due to some blocking issue, see https://github.com/maplibre/maplibre-compose/issues/738#issuecomment-3716900291

Oof, I see. Maybe add a comment to the empty implementation on iOS linking to that ticket?

sargunv:
Yeah that iOS issue is more and more of a problem. It's been many months since I last investigated -- maybe today's AI models will be better able to investigate the issue than when I tried last year.

I only skimmed this PR so far, but to answer @westnordost's questions:

>  In this case, it would make sense to align oneself with what Sargun would consider test-worthy

rough rule of thumb: if it's easy to verify in a quick run through of the demo app, no need to write tests. if it may fail in some edge cases (like a bug we'd actually ship) I want it tested eventually, but it's not a requirement right now because the test harnesses aren't well thought through yet.

and no need to test self-evident stuff like do data classes work, or do parameters get passed, or does maplibre-native work.

I _do_ generally want to improve the test coverage of the project, especially around how we integrate with maplibre-native and on real regressions we've had in the past. This PR doesn't need to do that work though.

> I don't know how @sargunv envisions how the demos should be structured.

I don't know how I envision it either 😅.

I generally add or change a demo when I want to make sure to manually test that functionality for every future release. And I prefer to change/add functionality to an existing demo for a small feature rather than a whole new demo for every feature, but I have no strong rules or opinions there.

For this feature, ideally we'd include it on the existing camera follow demo? Perhaps a bit awkward because that's a multiplatform demo, but seems like the right place for it rather than a whole new one.

> Generally I think having zoom buttons adds value to the demo app, so like you said, it could be an option to add them generally to all demos.

I think adding zoom buttons generally is reasonable (I half-remember already having that though? I probably misremember, will double check).

> Maybe add a comment to the empty implementation on iOS linking to that ticket?

+1. This _should_ eventually be multiplatform; just blocked on iOS right now due to #738.

demo-app/src/androidMain/kotlin/org/maplibre/compose/demoapp/demos/SynchronousGeoJsonUpdatesDemoPreview.android.kt
klemensz: 1. Which folder do you suggest instead?
2. It's probably just out of habit, because I usually create Previews for most Composables which makes it easier to change the UI without having to (re-)start the app. But in order to be consistent with the rest of the app we can remove it as well.

westnordost: I'd suggest commonMain, but yeah, I'd suggest to remove it.

klemensz: I thought that androidx.compose.ui.tooling.preview.Preview was Android-specific, but maybe my knowledge is outdated.

westnordost: It was, there was a different import for Preview for compose multiplatform, but recently they changed the package to match that of Android preview. (But it requires a dependency nonetheless)

lib/maplibre-compose/src/commonTest/kotlin/org/maplibre/compose/sources/GeoJsonOptionsTest.kt
klemensz: I removed it.

demo-app/src/androidMain/kotlin/org/maplibre/compose/demoapp/demos/SynchronousGeoJsonUpdatesDemoPreview.android.kt
klemensz: I removed it.

lib/maplibre-compose/src/androidDeviceTest/kotlin/org/maplibre/compose/sources/AndroidGeoJsonSourceTest.kt
klemensz: I removed it.

demo-app/src/commonMain/kotlin/org/maplibre/compose/demoapp/demos/Demo.kt
klemensz: If I'm not completely mistaken then only the overlay content of the currently opened demo will be rendered:

```kt
        state.demos
          .filter { state.isDemoOpen(it) }
          .forEach { it.MapOverlayContent(state = state, isOpen = true) }
```

klemensz:
Regarding the comments, as far as I can see, the only open topic now is the demo.

I think there are two decisions to be made:

1. One decision is to whether we want the zoom buttons on all demos or not.
2. The other decision is to whether to keep the new demo or add it to another demo
   - I couldn't find a "camera follow demo" that was mentioned. 
   - So I don't think that the effect can be observed in any of the currently existing demos by just adding the parameter.

sargunv:
> One decision is to whether we want the zoom buttons on all demos or not.

I think that'd be good

> I couldn't find a "camera follow demo" that was mentioned.

Ah, you're right. Seems I removed that demo at some point (probably while adding the user location demo).

> The other decision is to whether to keep the new demo or add it to another demo

I think this new demo is fine for now. I'd prefer it be in common code, not Android code, as it's _meant_ to be common, even if only functional on Android today.

EDIT: ah I see you've already done that

demo-app/src/commonMain/kotlin/org/maplibre/compose/demoapp/demos/Demo.kt
sargunv: I believe that's correct (we use it for the location arrow composable you contributed @westnordost)

I don't feel strongly; I'm okay with the zoom buttons here. I'm also okay with them being global. Maybe slight preference for global but I'd approve either.

sargunv (approved):
lgtm

will merge after trying it out myself

klemensz:
> I think this new demo is fine for now. I'd prefer it be in common code, not Android code, as it's _meant_ to be common, even if only functional on Android today.

The demo itself is in common code, but it's visible only on Android ([Platform.android.kt](https://github.com/maplibre/maplibre-compose/pull/761/changes#diff-e07887cff69f87012baf07025fe96a2a67fa1c73e25f2679725abad01efec3af)). Could be moved to main/general demos if that makes more sense.

demo-app/src/commonMain/kotlin/org/maplibre/compose/demoapp/demos/Demo.kt
klemensz: Sure ... when I find some time i'll take a look at the demo app again to see how this can be improved, and maybe some other things.

Issue #760, opened 2026-04-01
expose native location indicator layer 


klemensz:
While migrating [Ferrostar](https://github.com/stadiamaps/ferrostar) to org.maplibre.compose here:  [stadiamaps/ferrostar#853](https://github.com/stadiamaps/ferrostar/pull/853) I ran into a gap for navigation use cases.

### Use case

In navigation, the displayed user location is often not raw GPS.

Instead, apps typically use a preferred location:

- snapped-to-route while the user is still on-route
- raw location when off-route

This avoids the puck jumping around because of GPS noise.

### How Ferrostar handles this on iOS

On iOS with [MapLibreSwiftUI](https://github.com/maplibre/swiftui-dsl), Ferrostar already does this successfully:

- `NavigationState.preferredUserLocation` returns snapped or raw location depending on route state
- NavigationMapView feeds that into a StaticLocationManager
- MapLibre native tracking (followWithCourse) uses that location source directly

So on iOS:

- native puck
- native follow camera
- snapped location

all use the same native pipeline.

### Problem on Android Compose

maplibre-compose already has:

- LocationProvider
- UserLocationState
- rememberUserLocationState
- LocationPuck
- LocationTrackingEffect

But there does not seem to be a way to let the map’s native tracking / native puck use an app-provided location source.

Because of that, libraries like Ferrostar currently have to:

- compute a snapped location themselves
- move the camera manually
- render a custom puck separately

That works to some extend, but it is not as robust as the iOS approach.

### What we would like

A way to provide a custom tracked location source to MaplibreMap, so native tracking and native puck rendering can use it.

Conceptually something like:

```kt
val preferredLocationState = rememberUserLocationState(preferredLocationProvider)

MaplibreMap(
    baseStyle = baseStyle,
    cameraState = cameraState,
    trackedLocationState = preferredLocationState,
    trackingMode = TrackingMode.FollowWithCourse,
    locationPuck = LocationPuck.Default,
)
```

### Why this matters

For navigation, this would allow:

1. app computes preferred location
  - snapped on-route
  - raw off-route
4. map consumes that source directly
5. native puck + native follow camera stay in sync

This is preferable to rendering a separate custom puck.

### KMP angle

Since maplibre-compose is a KMP library, this seems like a good fit for:

- a shared common API
- platform-specific implementations underneath

At least Android and iOS seem relevant here.

iOS in particular looks like a natural fit, because this model already exists conceptually there via native tracking + app-provided location source.

[...]

PR #763, opened 2026-04-08
Add app-provided native location tracking for Android and iOS

klemensz:
## Description

This adds an opt-in native location tracking pipeline to MaplibreMap so apps can provide their own tracked location source and feed it into the native MapLibre follow camera and native current-location puck

Corresponding issue: #760 

New API:

- nativeLocationTracking: NativeLocationTracking? = null
- NativeLocationTracking
- NativeLocationTrackingState
- rememberNativeLocationTrackingState(...)
- UserTrackingMode
- NativeLocationPuck

Implementation highlights:

- MaplibreMap now forwards native tracking as a single atomic update.
- Android uses `LocationComponent` with app-provided location updates.
- iOS uses a custom `MLNLocationManager` backed by app-provided locations.
- Native tracking mode changes are reflected back into shared state.
- Style reload and adapter cleanup are handled for Android and iOS.
- Web and Desktop keep the API as a no-op.
- Docs, tests, and a demo were added.

This is intended for navigation use cases where the app wants native tracking to follow a computed location source instead of raw platform GPS.

## Test plan

Ran:

```
./gradlew :lib:maplibre-compose:compileCommonMainKotlinMetadata --no-daemon
./gradlew :lib:maplibre-compose:compileAndroidMain --no-daemon
./gradlew :lib:maplibre-compose:compileKotlinIosSimulatorArm64 --no-daemon
./gradlew :lib:maplibre-compose:iosSimulatorArm64Test --no-daemon
./gradlew :demo-app:compileDebugKotlinAndroid :demo-app:compileKotlinIosSimulatorArm64 --no-daemon
./gradlew mkdocsBuild --no-daemon
```

Manually tested:

- Android: demo app `Native Tracking` flow validated during development
- iOS: demo app `Native Tracking` flow validated during development, including follow-with-course turn behavior

## Checklist

No. This is additive and opt-in. Existing `CameraState`, `LocationTrackingEffect`, and Compose `LocationPuck` usage remain unchanged when `nativeLocationTracking` is not used.

- Android: Demo app on Android emulator, Android 16
- iOS: Demo app on iPhone 17 simulator, iOS 26.2


docs/docs/location.md
westnordost, line 1: Hmm, from this text, I don't quite understand the need for a "Native tracking pipeline", i.e. a different API.

Why couldn't a navigation app that has a custom snap-on-logic to roads implement a custom `LocationProvider` and just use the "Compose Pipeline"?

If you want this snap-on-logic to reside in maplibre-compose rather than everyone implementing their own thing, then maplibre-compose could offer a `NavigationLocationTrackingLocationProvider` or something with configurable parameters `userTrackingMode = UserTrackingMode.FollowWithCourse` etc.

sargunv: Thanks for taking a look at this; and yeah I agree

additionally, this `org.maplibre.android.location` package is (mostly) part of the MapLibre Android SDK, not the MapLibre Native core, and so if/when we wrap the native core directly, we'd likely end up dropping this functionality.

that's not a _hard_ blocker (the ornaments we support today are similar) but it's code I'm unlikely to want to maintain.

> If you want this snap-on-logic to reside in maplibre-compose rather than everyone implementing their own thing, then maplibre-compose could offer a NavigationLocationTrackingLocationProvider or something with configurable parameters userTrackingMode = UserTrackingMode.FollowWithCourse etc.

+1 to that; I'd probably support having a somewhat more flexible / general location API on maplibre-compose itself. Though we should draw a line _somewhere_. MapLibre Compose shouldn't be a general purpose location library for every use case. But common use cases used in many maps apps? sure.

sargunv:
Gonna close this for now as it sounds like #761 will cover the same use case without the dependency on bits of the SDK we intend to drop later.

If it turns out that's not the case, feel free to reopen the discussion in the GH issue or here.

PR #766, opened 2026-04-18
fix Linux segfault from stale JAWT drawable on OpenGL backend

### LINKED ISSUE #564: Segfault on desktop
### What happened?

Since #544, we've used MapLibre Native on desktop. This works well on macOS and Windows, but it's segfaulting on Linux. The segfault happens somewhere in the `render()` call.

I've had trouble figuring out how to attach a debugger to the app and investigate deeper.

### Affected Platforms

- [ ] Android
- [ ] iOS
- [x] Desktop (JVM)
- [ ] Browser (JS)
- [ ] Browser (Wasm)

### Platform Version

Fedora 42 with KDE

### Library Version

https://github.com/maplibre/maplibre-compose/commit/e6dbc8e3b7f91b9d641f23cd05c1bdda1ca3ca4c

### Sample Code

demo app

jamesarich:
Fixes #564

The segfault inside `render()` was caused by incorrect JAWT usage: `Display*` and `Drawable` were captured once in `JawtContext`'s constructor and cached forever. The JAWT spec requires re-fetching surface info after every `Lock()` call — the X11 `Drawable` (window ID) changes when the AWT peer is recreated on first show, on resize, or when `JAWT_LOCK_SURFACE_CHANGED` is returned. Passing a stale `Drawable` to `glXMakeCurrent`/`glXSwapBuffers` is what causes the crash.

Raised as one of the items in @sargunv's comment at https://github.com/sargunv/maplibre-native/pull/2#issuecomment-4262789003.

### Changes

- `jawt_context.hpp` — `lock()` now calls `GetDrawingSurfaceInfo()` after each `Lock()` and refreshes platform surface info (display, drawable, etc.); `unlock()` calls `FreeDrawingSurfaceInfo()` before `Unlock()`, following the required JAWT lifecycle. Constructor simplified to a single lock/unlock to initialise the `Display*` (which is the JVM's persistent X11 connection and never changes).
- `canvas_opengl_backend.cpp` — destructor: replace `check()` (throws from a destructor → UB during stack unwinding) with an early-return guard; use `glXMakeCurrent(display, None, nullptr)` to release the context without needing a valid window; `activate()` checks the `glXMakeCurrent` return value and throws a descriptive error on failure rather than silently proceeding with no current context; `deactivate()` also uses `None` as the drawable when releasing, for consistency.
- `canvas_frontend.cpp` — remove the TODO comment that flagged the segfault.

### Checklist

**To your knowledge, are you making any breaking changes?** No.

- macOS: builds and runs cleanly; cannot reproduce the Linux segfault on macOS but the fix is deterministic from the JAWT spec.

### DIFF
--- a/lib/maplibre-native-bindings-jni/src/main/cpp/canvas_frontend.cpp
+++ b/lib/maplibre-native-bindings-jni/src/main/cpp/canvas_frontend.cpp
@@ -80,7 +80,6 @@ CanvasRenderer_class::render(JNIEnv* env, jCanvasRenderer renderer) {
     auto* frontend = reinterpret_cast<maplibre_jni::CanvasRenderer*>(
       java_classes::get<CanvasRenderer_class>().getNativePointer(env, renderer)
     );
-    // TODO: something in here is segfaulting on Linux
     frontend->render();
   } catch (const std::exception& e) {
     smjni::java_exception::translate(env, e);
--- a/lib/maplibre-native-bindings-jni/src/main/cpp/canvas_opengl_backend.cpp
+++ b/lib/maplibre-native-bindings-jni/src/main/cpp/canvas_opengl_backend.cpp
@@ -31,26 +31,34 @@ class OpenGLRenderableResource final : public mbgl::gl::RenderableResource {
       : backend(backend_), jawtContext(env, canvas_) {}
 
   ~OpenGLRenderableResource() {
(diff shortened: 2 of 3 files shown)


sargunv:
I've also prompted AI to help investigate this, and it loves to flag this, but the segfault persists on Linux. This bug requires some deeper exploration (and actual testing on a Linux machine)

PR #774, opened 2026-04-25
Update Compose to resolve Linux segfault

### LINKED ISSUE #564: Segfault on desktop

sargunv:
## Summary

Resolves #564. Compose v1.10 includes a version of skiko that uses the system's libstdc++ instead of statically linking it. This somehow resolves the segfault when combined with other libraries using libstdc++ like MapLibre.

Also removes apple x64 targets as they're being removed in Compose v1.11

### DIFF
--- a/demo-app/build.gradle.kts
+++ b/demo-app/build.gradle.kts
@@ -1,4 +1,3 @@
-import org.jetbrains.compose.ExperimentalComposeLibrary
 import org.jetbrains.compose.desktop.application.dsl.TargetFormat
 import org.jetbrains.kotlin.gradle.ExperimentalKotlinGradlePluginApi
 import org.jetbrains.kotlin.gradle.plugin.KotlinSourceSetTree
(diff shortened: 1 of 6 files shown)


westnordost:
> Also removes apple x64 targets as they're being removed in Compose v1.11

Ah.

Sources:

https://github.com/JetBrains/compose-multiplatform/releases/tag/v1.11.0-alpha02
https://android-review.googlesource.com/c/platform/frameworks/support/+/3889609

PR #775, opened 2026-04-26
Add desktop Vulkan canvas backend

### LINKED ISSUE #566: Vulkan support on desktop
We should add support for the Vulkan renderer backend on desktop. A sample (though AI slop) implementation for Linux, Windows, and macOS is available at https://github.com/sargunv/maplibre-native-jvm-demo/blob/main/maplibre-jni/src/main/cpp/awt_vulkan_backend.cpp

sargunv:
Resolves #566. Based on my PoC linked in that issue.

Validated on Linux x64. Will validate Windows sometime later when I get the project building on windows again.

Not supporting Vulkan on macOS for now (though it's possible with MoltenVK) since we have a Metal implementation.

### DIFF
--- a/lib/maplibre-native-bindings-jni/src/main/cpp/canvas_renderer.hpp
+++ b/lib/maplibre-native-bindings-jni/src/main/cpp/canvas_renderer.hpp
@@ -74,6 +74,8 @@ class CanvasBackend : public mbgl::vulkan::RendererBackend,
   explicit CanvasBackend(JNIEnv* env, jCanvas canvas);
   auto getDefaultRenderable() -> mbgl::gfx::Renderable& override;
   void setSize(mbgl::Size);
+  bool lockSurfaceForRender();
+  void unlockSurfaceAfterRender();
 
  protected:
   auto getInstanceExtensions() -> std::vector<const char*> override;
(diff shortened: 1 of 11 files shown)


lib/maplibre-native-bindings-jni/src/main/cpp/canvas_renderer.hpp
```diff
@@ -21,8 +21,15 @@
...
 #include <mbgl/gl/renderer_backend.hpp>
+#elif defined(USE_VULKAN_BACKEND)
+#include <mbgl/vulkan/renderable_resource.hpp>
+#include <mbgl/vulkan/renderer_backend.hpp>
 #endif
 
+// clang-format off
+#include "fix_x11_pollution.h"
+// clang-format on
+
```
westnordost, line 32: Where does this come from? Is this used at all?

sargunv: Yeah, https://github.com/maplibre/maplibre-compose/blob/8121eb01a8f121085ba1598f22d59df25ce31095/lib/maplibre-native-bindings-jni/src/main/cpp/fix_x11_pollution.h

These macros get defined by X11 and they conflict with some other identifiers in MapLibre Native's headers. Basically if you import x11's header you can't name anything `Always` or `None`

We don't use them so I include this hack to undef them

PR #777, opened 2026-04-26
update kotlin, agp, gradle, etc

sargunv:
updating some dependencies

this one was a roller coaster as I upgraded KSP to latest, encountered https://github.com/gershnik/SimpleJNI, worked around it, then realized I could downgrade KSP slightly to avoid it.

I took care of the easiest deprecations in this PR, but a later PR will have to separate the demo-app Android project from the demo-app shared KMP codebase.


mise.toml
```diff
@@ -16,6 +16,7 @@ actionlint = "1.7.8"
 bazel = "7.7.0"
 xcodes = { version = "1.6.2", os = ["macos"] }
 swiftformat = "0.61.0"
+python = "3.14.4"
```
westnordost, line 19: Why Python? I don't see any use of Python

sargunv: I thought so too, so I removed it a couple days ago, then realized it was actually used in the documentation generation gradle task

PR #780, opened 2026-04-26
Update GitHub Actions versions

sargunv:

Added a mise task to upgrade GH actions and pin their SHAs, and updated all our current actions.

### DIFF
--- a/.github/actions/run-with-avd/action.yml
+++ b/.github/actions/run-with-avd/action.yml
@@ -17,7 +17,7 @@ runs:
         sudo udevadm control --reload-rules
         sudo udevadm trigger --name-match=kvm
     - name: Cache AVD
-      uses: actions/cache@v4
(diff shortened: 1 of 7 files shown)


.github/actions/run-with-avd/action.yml
```diff
@@ -17,7 +17,7 @@ runs:
         sudo udevadm control --reload-rules
         sudo udevadm trigger --name-match=kvm
     - name: Cache AVD
-      uses: actions/cache@v4
+      uses: actions/cache@27d5ce7f107fe9357f9df03efb73ab90386fccae # v5.0.5
```
westnordost, line 20: That looks like a commit hash. Is this normal?

sargunv: Yes, to pin the exact version (similar to like a lock file with many package managers) 

It's there to guard against supply chain attacks through GitHub actions.

PR #781, opened 2026-04-27
Document iOS HarfBuzz link-order workaround and upgrade MapLibre iOS

### LINKED ISSUE #551: Compose text rendering broken with MapLibre iOS >= 6.18.0
When I implement maplibre in my project my text starts to look like this (screenshot).
In the penultimate case, the symbols look good, this was achieved by installing custom FontFamaly (without the standard ones and the box)
 
Image

P.S. I'm not sure if it's only because of maplibre, but removing it from the project fixes the problem...
 

### Affected Platforms

- [ ] Android
- [x] iOS
- [ ] Desktop (JVM)
- [ ] Browser (JS)
- [ ] Browser (Wasm)

### Platform Version

iOS 18.5

### Library Version

0.10.4

### Kotlin Version

2.2.0

### Compose Version

1.8.2

### Sample Code

```
fun Fix(){
     Text(
        text = "Some text",
        fontFamily = FontFamily.MyAnyFont, // fix problem
    )
}
```

sargunv:

Resolves #551 by documenting a known workaround. It's not a real fix, and I'm concerned there may be other related bugs lurking, but it's good enough to unblock upgrading MapLibre iOS I think

### DIFF
--- a/docs/docs/getting-started.md
+++ b/docs/docs/getting-started.md
@@ -64,6 +64,21 @@ The easiest way is to select one of these two Gradle plugins:
 - JetBrains's [CocoaPods plugin][gradle-cocoapods]
 - Third party [Swift Package Manager plugin][gradle-spm4kmp]
 
+!!! warning
+
+    In Xcode, ensure your Kotlin/Compose framework is linked before
+    `MapLibre.framework`. Select your iOS app target, open **Build Settings**,
+    search for **Other Linker Flags**, and order the flags like this:
+
+    ```text
+    -framework ComposeApp
+    -framework MapLibre
+    ```
+
+    Replace `ComposeApp` with your Kotlin framework name. The opposite order can
+    cause broken Compose text rendering on iOS because both Compose and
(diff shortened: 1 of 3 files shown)


docs/docs/getting-started.md
```diff
@@ -64,6 +64,21 @@ The easiest way is to select one of these two Gradle plugins:
...
+
+    ```text
+    -framework ComposeApp
+    -framework MapLibre
+    ```
+
+    Replace `ComposeApp` with your Kotlin framework name. The opposite order can
+    cause broken Compose text rendering on iOS because both Compose and
+    MapLibre include HarfBuzz symbols.
+
```
westnordost, line 81:
```suggestion

See [CMP-8882](https://youtrack.jetbrains.com/issue/CMP-8882/)
```

sargunv: done in https://github.com/maplibre/maplibre-compose/pull/791

PR #790, opened 2026-04-27
Xcode: improve Java lookup and extract KMP framework embedding logic to a separate shell script and 

klemensz:
## Description

On my development machine the previous script couldn't find Java (I have it installed with [SDKMAN](https://sdkman.io/)) so I improved Java lookup and extracted the script into it's own file to make it more readable.

## Test plan

Built and ran the demo app on the iOS simulator from within Xcode.

## Checklist

no

- iOS: macOS Tahoe 26.3.1, Xcode 26.4.1

_Created using Codex with GPT-5_

### DIFF
--- a/iosApp/scripts/embed-kmp-framework.sh
+++ b/iosApp/scripts/embed-kmp-framework.sh
@@ -1,0 +1,31 @@
+#!/bin/sh
+
+set -eu
+
+if [ "YES" = "${OVERRIDE_KOTLIN_BUILD_IDE_SUPPORTED:-}" ]; then
+  echo "Skipping Gradle build task invocation due to OVERRIDE_KOTLIN_BUILD_IDE_SUPPORTED environment variable set to \"YES\""
+  exit 0
+fi
+
+if ! java -version >/dev/null 2>&1 && [ -z "${JAVA_HOME:-}" ]; then
+  if [ -x "$HOME/.sdkman/candidates/java/current/bin/java" ]; then
+    export JAVA_HOME="$HOME/.sdkman/candidates/java/current"
+  elif [ -x "/usr/libexec/java_home" ]; then
+    JAVA_HOME_CANDIDATE="$(/usr/libexec/java_home 2>/dev/null || true)"
+    if [ -n "$JAVA_HOME_CANDIDATE" ]; then
+      export JAVA_HOME="$JAVA_HOME_CANDIDATE"
+    fi
+  fi
+fi
+
+if [ -n "${JAVA_HOME:-}" ]; then
+  export PATH="$JAVA_HOME/bin:$PATH"
+fi
+
+if ! java -version >/dev/null 2>&1; then
+  echo "Java runtime not found. Set JAVA_HOME or install a JDK so Gradle can run from Xcode."
+  exit 1
+fi
+
+cd "$SRCROOT/.."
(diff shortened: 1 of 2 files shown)


iosApp/scripts/embed-kmp-framework.sh
```diff
@@ -0,0 +1,31 @@
...
+set -eu
+
+if [ "YES" = "${OVERRIDE_KOTLIN_BUILD_IDE_SUPPORTED:-}" ]; then
+  echo "Skipping Gradle build task invocation due to OVERRIDE_KOTLIN_BUILD_IDE_SUPPORTED environment variable set to \"YES\""
+  exit 0
+fi
+
+if ! java -version >/dev/null 2>&1 && [ -z "${JAVA_HOME:-}" ]; then
+  if [ -x "$HOME/.sdkman/candidates/java/current/bin/java" ]; then
+    export JAVA_HOME="$HOME/.sdkman/candidates/java/current"
```
sargunv: I think rather than supporting sdkman directly, I'd prefer to rely on JAVA_HOME here. And assume however the user installs java sets either JAVA_HOME or configures /usr/libexec/java_home.

(mise sets JAVA_HOME, I assume sdkman should too?)

klemensz: Yes, generally I have JAVA_HOME available, but it looks like Xcode doesn't pick it up. It looks like it doesn't source `~/.zshrc`. 

I found this article https://www.marcogomiero.com/posts/2021/kmp-no-java-runtime-error-xcode/ but it's already 5 years old, I need to investigate more.

There is a way to set a custom location path under Xcode -> Settings -> Locations, I need to try that as well.

klemensz:
@sargunv Thanks for reacting so quickly, but with your latest commit it doesn't work (again) in my environment - see my other comment. I'll investigate more why JAVA_HOME is not available in Xcode. At least now it's easier to modify the script because it's not all in a single line any more ;)

iosApp/scripts/embed-kmp-framework.sh
sargunv: Try setting it in zprofile or zshenv, not zshrc

zshrc is for interactive terminals, session vars go in profile or env

sargunv: (I put such things in .profile and have both .bash_profile and .zprofile source _that_)

klemensz:
After doing some more research, it seems that Xcode doesn't pick up JAVA_HOME at all, neither from .bash_profile nor from .zprofile. It uses `/bin/sh` which doesn't read the shell configs. Xcode takes java from `/usr/libexec/java_home`. 

It turns out the two easiest solutions for me are:
- Install the JDK manually for example with homebrew (`brew install openjdk@21`), in addition to the SDKMAN installed version(s) 
- Use this symlink + Info.plist solution: https://stackoverflow.com/a/76523554

I opted for the second one, and it seems to work :) 

I hope that people running into the same issue (I think it's relevant for KMP/CMP projects in general if using SDKMAN) will find this and save themselves some time.

sargunv:
Interesting. I don't have any global java install, nor anything from /usr/libexec/java_home

I only install Java through mise, and I assume XCode is finding Java through the mise shim on PATH

PR #818, opened 2026-06-08
Fix IosLocationProvider locationManager and delegate garbage collection

DestBro:
## Description

This PR stabilizes the iOS location provider garbage collection.
- **iOS Location Stability:** Refactored `IosLocationProvider` to maintain a stable `CLLocationManager` and its delegate, preventing premature garbage collection and ensuring consistent updates.

grafik
If you delete any of these 3 red lines then the bug is there. 

## Test plan

Verified that location updates are correctly received and displayed in the demo app on iOS and Android.

## Checklist

No

- Android: Tested on API 34 emulator.
- iOS: Tested on iPhone 15 simulator.

## AI assistance

- **Tools:** Android Studio Bot
- **Models:** Gemini 2.0 Flash
- **Context:** Fixes for location manager garbage collection and bearing inversion.

## Summary by CodeRabbit

* **Performance**
  * Reduced iOS location provider resource use and improved sharing of location updates for more efficient background handling.

* **Refactor**
  * Simplified internal location/update handling by reusing manager and delegate instances; public APIs remain unchanged.

* **Bug Fixes**
  * Improved cleanup on stop to halt updates and clear internal delegates, reducing potential resource leaks.

### DIFF
--- a/lib/maplibre-compose/src/iosMain/kotlin/org/maplibre/compose/location/IosLocationProvider.kt
+++ b/lib/maplibre-compose/src/iosMain/kotlin/org/maplibre/compose/location/IosLocationProvider.kt
@@ -11,12 +11,14 @@ import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.FlowPreview
 import kotlinx.coroutines.channels.SendChannel
 import kotlinx.coroutines.channels.awaitClose
+import kotlinx.coroutines.flow.SharedFlow
 import kotlinx.coroutines.flow.SharingStarted
 import kotlinx.coroutines.flow.StateFlow
 import kotlinx.coroutines.flow.callbackFlow
 import kotlinx.coroutines.flow.flowOn
 import kotlinx.coroutines.flow.runningFold
 import kotlinx.coroutines.flow.sample
+import kotlinx.coroutines.flow.shareIn
 import kotlinx.coroutines.flow.stateIn
 import org.maplibre.spatialk.units.Bearing
 import org.maplibre.spatialk.units.Length
@@ -60,21 +62,25 @@ public class IosLocationProvider(
   sharingStarted: SharingStarted,
 ) : LocationProvider, OrientationProvider {
 
+  private val locationManager by lazy { CLLocationManager() }
+  private var delegate: Delegate? = null
+
   init {
+    val authorizationStatus = CLLocationManager.authorizationStatus()
     if (
       enableLocation &&
-        CLLocationManager.authorizationStatus() != kCLAuthorizationStatusAuthorizedAlways &&
-        CLLocationManager.authorizationStatus() != kCLAuthorizationStatusAuthorizedWhenInUse
+        authorizationStatus != kCLAuthorizationStatusAuthorizedAlways &&
+        authorizationStatus != kCLAuthorizationStatusAuthorizedWhenInUse
     ) {
       throw PermissionException()
     }
   }
 
-  private val updates: StateFlow<ProviderUpdate?> =
+  private val updates: SharedFlow<ProviderUpdate> =
     callbackFlow {
-        val locationManager = CLLocationManager()
-        val delegate = Delegate(channel)
-        locationManager.delegate = delegate
+        val d = Delegate(channel)
+        delegate = d
+        locationManager.delegate = d
 
         if (enableLocation) {
           locationManager.desiredAccuracy =
@@ -107,10 +113,11 @@ public class IosLocationProvider(
             locationManager.stopUpdatingHeading()
           }
           locationManager.delegate = null
+          delegate = null
         }
       }
       .flowOn(Dispatchers.Main)
-      .stateIn(coroutineScope, sharingStarted, null)
+      .shareIn(coroutineScope, sharingStarted, replay = 1)
 
   override val location: StateFlow<Location?> =
     updates
@@ -133,11 +140,10 @@ public class IosLocationProvider(
       .sample(orientationUpdateInterval)
       .stateIn(coroutineScope, sharingStarted, null)
 
-  private inner class Delegate(private val channel: SendChannel<ProviderUpdate>) :
+  private class Delegate(private val channel: SendChannel<ProviderUpdate>) :
     NSObject(), CLLocationManagerDelegateProtocol {
     override fun locationManager(manager: CLLocationManager, didUpdateLocations: List<*>) {
       @Suppress("UNCHECKED_CAST") val locations = didUpdateLocations as? List<CLLocation>
-
       locations?.forEach { channel.trySend(ProviderUpdate.LocationUpdate(it.asMapLibreLocation())) }
     }
 


jgillich:
Can you explain the bearing change? What "common code expectations"? Bearing was flipped on Android, which I just fixed in c48191f315b95c8a21c81837cc09a3d135bbccd3. But I don't remember how that came to be

Edit: Seems like you just dropped it, nevermind then :smile:

DestBro:
> Can you explain the bearing change? What "common code expectations"? Bearing was flipped on Android, which I just fixed in [c48191f](https://github.com/maplibre/maplibre-compose/commit/c48191f315b95c8a21c81837cc09a3d135bbccd3). But I don't remember how that came to be
> 
> Edit: Seems like you just dropped it, nevermind then 😄

I dropped it to make the PR & commit clean because I saw the bug with bearing could potentially add more edits if checked more properly

DestBro:
> Can you explain the bearing change? What "common code expectations"? Bearing was flipped on Android, which I just fixed in [c48191f](https://github.com/maplibre/maplibre-compose/commit/c48191f315b95c8a21c81837cc09a3d135bbccd3). But I don't remember how that came to be
> 
> Edit: Seems like you just dropped it, nevermind then 😄

On iOS right now (main branch) the Left and Right (West and East) dirrections are mirrored, I did not analyse in details but I would like a fix for it as well

DestBro:
I tried to understand it with my agent:

## Root Cause Analysis

The bug is caused by the common code misinterpreting the absolute bearing provided by the platform as a relative rotation **towards** North, rather than an angle **from** North.

### Mathematical Explanation
Imagine you are facing **East (90°)**.
- The platform correctly reports: "The bearing is 90°".
- The current code uses `clockwiseRotationTo(Bearing.North)`. This asks: "How much do I need to turn clockwise to reach North (0°)?"
- To get from 90° to 0° turning clockwise, you must turn **270°**.
- The app then applies a **270°** rotation to the UI, which points **West**.

This result is a mirrored coordinate system where every direction is inverted across the North-South axis.

DestBro:
> Can you explain the bearing change? What "common code expectations"? Bearing was flipped on Android, which I just fixed in [c48191f](https://github.com/maplibre/maplibre-compose/commit/c48191f315b95c8a21c81837cc09a3d135bbccd3). But I don't remember how that came to be
> 
> Edit: Seems like you just dropped it, nevermind then 😄

## Proposed Changes

We will update the common code to use the absolute bearing value (the angle from North) instead of the rotation required to reach North.

### [Component Name] Common Location Logic

#### [MODIFY] [LocationPuck.kt](file:///Users/dest/Documents/maplibre-compose/lib/maplibre-compose/src/commonMain/kotlin/org/maplibre/compose/location/LocationPuck.kt)
Update the GeoJson property mapping to use the absolute bearing.
```diff
-put("bearing", bearing?.value?.smallestRotationTo(Bearing.North)?.inDegrees)
+put("bearing", bearing?.value?.let { (it - Bearing.North).inDegrees })
```

#### [MODIFY] [LocationTrackingEffect.kt](file:///Users/dest/Documents/maplibre-compose/lib/maplibre-compose/src/commonMain/kotlin/org/maplibre/compose/location/LocationTrackingEffect.kt)
Update the camera tracking logic to use absolute bearings for course, orientation, and automatic tracking modes.
```diff
-bearing = currentLocation.location?.course?.value?.clockwiseRotationTo(Bearing.North)?.inDegrees
+bearing = currentLocation.location?.course?.value?.let { (it - Bearing.North).inDegrees }
...
-bearing = currentLocation.orientation?.orientation?.value?.clockwiseRotationTo(Bearing.North)?.inDegrees
+bearing = currentLocation.orientation?.orientation?.value?.let { (it - Bearing.North).inDegrees }
...
-bearing = currentLocation.mostAccurateBearing()?.value?.clockwiseRotationTo(Bearing.North)?.inDegrees
+bearing = currentLocation.mostAccurateBearing()?.value?.let { (it - Bearing.North).inDegrees }
```

DestBro:
@sargunv can it be merged now ?

sargunv:
Sounds like the intended fix is retaining a reference to the delegate, as CLLocationManager holds a weak reference. 

Why the switch to SharedFlow vs StateFlow?

Why the removal of `inner`?

> can it be merged now ?

clean up the indent issues failing CI

DestBro:
> Sounds like the intended fix is retaining a reference to the delegate, as CLLocationManager holds a weak reference. 
> 
> 
> 
> Why the switch to SharedFlow vs StateFlow?
> 
> 
> 
> Why the removal of `inner`?
> 
> 
> 
> > can it be merged now ?
> 
> 
> 
> clean up the indent issues failing CI

Both SharedFlow and StateFlow are hot flows, but the first suits better when you have two subscribers and when the items are different in the flow. SharedFlow will keep all values so far, whereas StateFlow will keep just the last value potentially dropping Orientation updates for example. You should use StateFlow just to represent a value for example in UI, and use SharedFlow when two or more subscribers are connected to potentially "hard" computations.

DestBro:
> Sounds like the intended fix is retaining a reference to the delegate, as CLLocationManager holds a weak reference. 
> 
> 
> 
> Why the switch to SharedFlow vs StateFlow?
> 
> 
> 
> Why the removal of `inner`?
> 
> 
> 
> > can it be merged now ?
> 
> 
> 
> clean up the indent issues failing CI

The removal of inner modifier was suggested by IDE because it will be ignored anyway.

DestBro:
> Sounds like the intended fix is retaining a reference to the delegate, as CLLocationManager holds a weak reference.
> 
> Why the switch to SharedFlow vs StateFlow?
> 
> Why the removal of `inner`?
> 
> > can it be merged now ?
> 
> clean up the indent issues failing CI

I used the automatic formating from the IDE, would be good to have a code-style.xml file in the project

Edit: I found you use mise, hk, dprint, so you do not need code-style.xml

DestBro:
@sargunv the pipeline is green, now it can be merged?

DestBro:
This one is interesting though:

"Sounds like the intended fix is retaining a reference to the delegate, as CLLocationManager holds a weak reference."

On the JVM I never worried about references and mistake garbage collection, it's first time I met this kind of bug, though it's the native target so unexpected things may happen.

sargunv:
> On the JVM I never worried about references and mistake garbage collection, it's first time I met this kind of bug, though it's the native target so unexpected things may happen.

It's because `delegate` is declared as a `weak var`, you can see the documentation here: https://developer.apple.com/documentation/corelocation/cllocationmanager/delegate

The JVM has the same concept: https://docs.oracle.com/en/java/javase/21/docs/api/java.base/java/lang/ref/WeakReference.html but it's not as commonly used as in the ObjC/Swift world because the latter relies on ARC rather than garbage collection, so strong reference cycles are problematic.

DestBro:
> > On the JVM I never worried about references and mistake garbage collection, it's first time I met this kind of bug, though it's the native target so unexpected things may happen.
> 
> It's because `delegate` is declared as a `weak var`, you can see the documentation here: https://developer.apple.com/documentation/corelocation/cllocationmanager/delegate
> 
> The JVM has the same concept: https://docs.oracle.com/en/java/javase/21/docs/api/java.base/java/lang/ref/WeakReference.html but it's not as commonly used as in the ObjC/Swift world because the latter relies on ARC rather than garbage collection, so strong reference cycles are problematic. 

Crazy, from beginning I was thinking it's an issue in the usage of the flows framework of kotlin, that there is a bottleneck, by refactoring things out I in the end made the delegator a class member without knowing that it was the thing that made the functionality work.

On JVM one does not think everyday about garbage collection issues, pointers, references, weak references, link counters etc. So it was hard to spot this bug.
With Kotlin Native you need to shift the attitude model, and be prepared that the native code can have special behaviours.

PR #822, opened 2026-06-18
Fix orientation by adding Bearing.inDegreesClockwise

DestBro:
### Summary
- Fixed mirrored West-East orientation in the location puck by ensuring clockwise bearing calculations.
- Improved code ergonomics by adding a `Bearing.inDegreesClockwise` extension property.

### Changes
- Added `inDegreesClockwise` extension to `Bearing` to provide a standardized clockwise angle from North.
- Refactored `LocationPuck`, `LocationTrackingEffect`, and demo applications to use the new extension, ensuring consistent orientation behavior.
- Replaced incorrect `smallestRotationTo(Bearing.North)` usage which caused the mirroring effect.

### Verification
- Verified the fix on iOS as requested by the user.

### DIFF
--- a/demo-app/src/androidMain/kotlin/org/maplibre/compose/demoapp/demos/GmsLocationDemo.kt
+++ b/demo-app/src/androidMain/kotlin/org/maplibre/compose/demoapp/demos/GmsLocationDemo.kt
@@ -29,7 +29,6 @@ import org.maplibre.compose.location.UserLocationState
 import org.maplibre.compose.location.mostAccurateBearing
 import org.maplibre.compose.location.rememberUserLocationState
 import org.maplibre.compose.material3.LocationPuckDefaults
-import org.maplibre.spatialk.units.Bearing
 import org.maplibre.spatialk.units.extensions.inDegrees
 import org.maplibre.spatialk.units.extensions.inMeters
 
@@ -94,14 +93,10 @@ object GmsLocationDemo : Demo {
             "Position: ${locationState?.location?.position?.value} +- ${locationState?.location?.position?.accuracy?.inMeters?.roundToInt()}m"
           )
           Text(
-            "Course: ${locationState?.location?.course?.value?.smallestRotationTo(Bearing.North)?.inDegrees?.roundToInt()} +- ${locationState?.location?.course?.accuracy?.inDegrees?.roundToInt()}"
+            "Course: ${locationState?.location?.course?.value} +- ${locationState?.location?.course?.accuracy?.inDegrees?.roundToInt()}"
           )
           Text(
-            "Orientation: ${
-              locationState?.orientation?.orientation?.value?.smallestRotationTo(
-                Bearing.North
-              )?.inDegrees?.roundToInt()
-            } +- ${locationState?.orientation?.orientation?.accuracy?.inDegrees?.roundToInt()}"
+            "Orientation: ${locationState?.orientation?.orientation?.value} +- ${locationState?.orientation?.orientation?.accuracy?.inDegrees?.roundToInt()}"
           )
         }
       }
--- a/demo-app/src/commonMain/kotlin/org/maplibre/compose/demoapp/demos/UserLocationDemo.kt
+++ b/demo-app/src/commonMain/kotlin/org/maplibre/compose/demoapp/demos/UserLocationDemo.kt
@@ -30,7 +30,6 @@ import org.maplibre.compose.location.LocationPuck
 import org.maplibre.compose.location.LocationTrackingEffect
 import org.maplibre.compose.map.GestureOptions
 import org.maplibre.compose.material3.LocationPuckDefaults
-import org.maplibre.spatialk.units.Bearing
 import org.maplibre.spatialk.units.extensions.degrees
 import org.maplibre.spatialk.units.extensions.inDegrees
 import org.maplibre.spatialk.units.extensions.inMeters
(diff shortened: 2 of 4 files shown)


lib/maplibre-compose/src/commonMain/kotlin/org/maplibre/compose/location/BearingExtensions.kt
```diff
@@ -0,0 +1,8 @@
+package org.maplibre.compose.location
+
+import org.maplibre.spatialk.units.Bearing
+import org.maplibre.spatialk.units.extensions.inDegrees
+
+/** Returns the bearing in degrees clockwise from North in the range [0, 360). */
+public val Bearing.inDegreesClockwise: Double
+  get() = Bearing.North.clockwiseRotationTo(this).inDegrees
```
sargunv: why add a public extension rather than inlining this small logic (or an `internal` helper at most)?

sargunv:
Was the bug present on both Android + iOS or just iOS?

If just iOS, what's different about the two Bearings such that it works on Android?

DestBro:
> Was the bug present on both Android + iOS or just iOS?
> 
> If just iOS, what's different about the two Bearings such that it works on Android?

Yes it's on Android as well:
grafik

DestBro:
Some changes were made on the orientation common code not long ago and it broke both platforms.

jgillich:
I forgot to update the demos in c48191f315b95c8a21c81837cc09a3d135bbccd3. Lgtm

lib/maplibre-compose/src/commonMain/kotlin/org/maplibre/compose/location/BearingExtensions.kt
DestBro: > "Why adding a public extension" ?

It makes the readability of the code easier than inlining the code. So far it has same visibility as other methods of the Bearing class. I could add the method to the spatialk library but It would take more time to do work in multiple code-bases.


> (or an internal helper at most)

Not sure what problem are you trying to solve, this property/getter method is scoped already to the Bearing class, everywhere where you can access Bearing, you should be able to call inDegreesClockwise property as well.
Maybe I do not understand your point here, a code example and a comment about how it would make things better is much appreciated

DestBro:
> I forgot to update the demos in [c48191f](https://github.com/maplibre/maplibre-compose/commit/c48191f315b95c8a21c81837cc09a3d135bbccd3). Lgtm

There were changes about how you calculate the bearing itself:
`bearing?.value?.smallestRotationTo(Bearing.North)?.inDegrees`
changed to:
`Bearing.North.clockwiseRotationTo(course?.value).inDegrees`

`course?.value?.clockwiseRotationTo(Bearing.North)?.inDegrees`
changed to:
`Bearing.North.clockwiseRotationTo(bearing.value).inDegrees`

lib/maplibre-compose/src/commonMain/kotlin/org/maplibre/compose/location/BearingExtensions.kt
sargunv: > Not sure what problem are you trying to solve

1. MapLibre Compose isn't the right place to be adding public APIs for Spatial-K, unless that helper is very specifically related to interop with MapLibre

2. The bearing value class is designed to be agnostic to what your anchor (North in this case) is. So adding a special helper for North doesn't really fit its design

sargunv: I think all you really need at the call site is `(bearing - North).inDegrees`

sargunv: And in the cases where you're displaying text, no need to convert it to a Double at all; Bearing can handle its own toString (renders a decimal quadrant bearing by default, or you can `toDmsString` for degrees/minutes/seconds)

DestBro: @sargunv fixed

demo-app/src/androidMain/kotlin/org/maplibre/compose/demoapp/demos/GmsLocationDemo.kt
```diff
@@ -94,14 +93,10 @@ object GmsLocationDemo : Demo {
...
-            "Course: ${locationState?.location?.course?.value?.smallestRotationTo(Bearing.North)?.inDegrees?.roundToInt()} +- ${locationState?.location?.course?.accuracy?.inDegrees?.roundToInt()}"
+            "Course: ${locationState?.location?.course?.value} +- ${locationState?.location?.course?.accuracy?.inDegrees?.roundToInt()}"
           )
           Text(
-            "Orientation: ${
-              locationState?.orientation?.orientation?.value?.smallestRotationTo(
-                Bearing.North
-              )?.inDegrees?.roundToInt()
-            } +- ${locationState?.orientation?.orientation?.accuracy?.inDegrees?.roundToInt()}"
+            "Orientation: ${locationState?.orientation?.orientation?.value} +- ${locationState?.orientation?.orientation?.accuracy?.inDegrees?.roundToInt()}"
```
sargunv, line 99: you can leave bearing and angle as the value class and let the included toString handle it

demo-app/src/commonMain/kotlin/org/maplibre/compose/demoapp/demos/UserLocationDemo.kt
```diff
@@ -196,14 +195,10 @@ object UserLocationDemo : Demo {
...
-              "Course: ${state.location?.course?.value?.smallestRotationTo(Bearing.North)?.inDegrees?.roundToInt()} +- ${state.location?.course?.accuracy?.inDegrees?.roundToInt()}"
+              "Course: ${state.location?.course?.value} +- ${state.location?.course?.accuracy?.inDegrees?.roundToInt()}"
             )
             Text(
-              "Orientation: ${
-                state.orientation?.orientation?.value?.smallestRotationTo(
-                  Bearing.North
-                )?.inDegrees?.roundToInt()
-              } +- ${state.orientation?.orientation?.accuracy?.inDegrees?.roundToInt()}"
+              "Orientation: ${state.orientation?.orientation?.value} +- ${state.orientation?.orientation?.accuracy?.inDegrees?.roundToInt()}"
```
sargunv, line 201: ditto

demo-app/src/androidMain/kotlin/org/maplibre/compose/demoapp/demos/GmsLocationDemo.kt
DestBro: the code from above (unchanged by my MR) does it like you said?

I'm afraid it will not be consistent with the whole codebase, and anyway I closed my PC, I will check tomorrow.

sargunv: fair enough, can clean it up later

DestBro: Lol, I was just approaching my PC, now waiting a release, I don't like the Snapshots from github

Issue #592, opened 2025-09-06
Reduce API indirection


sargunv:
Just about every API we expose in maplibre-compose goes through multiple layers of indirection to the actual implementation:

* The KMP API, typically a pure Kotlin class or an expect/actual class
* Calls the platform API (MapLibre Android, MapLibre iOS, maplibre-native-bindings, etc)
* Calls the C++ native core API

Except for JS, where the KMP API calls the JS API.

Previously, the KMP API was the only one we controlled, and it's the one we used to unify the varying platform APIs. But with the native core integration for desktop, we now control one of the platform APIs, so we're now writing _two_ layers ourselves instead of one. 

Probably after #572, we should look at collapsing two of the layers into one, eliminating indirection and simplifying the API by matching the original one more closely. 

The remaining complication will be in how to unify the native core and JS APIs into one.

Issue #486, opened 2025-06-27
Demo app bugs


sargunv:
Since #476:
- [x] the hide/show button on certain layers causes a crash (segfault)
  - very likely the same cause as #236
- [ ] bottom sheet doesn't work great with certain screens / landscape mode
- [x] there's no suitable UI for non-blending platforms (desktop and web)
- [ ] the marker dialog on the markers demo doesn't show unless the demo is selected
- [ ] some demos have blank UI
- [ ] network calls crash the app if they fail (due to no connection, timeout, etc)
- [ ] style switcher doesn't work on desktop and js
- [ ] spatial-k v0.4.0 crashes on numeric feature ids (amtrak stations)

mtpdog:
> * network calls crash the app if they fail (due to no connection, timeout, etc)

In my opinion, this is the most significant bug at the moment.

mtpdog:
@sargunv  Is there any chance that network crashes will be fixed? This bug appears on slow connections as well and makes the Maplibre Composable unusable in production.

sargunv:
This issue refers to bugs in the demo app, not in the library. Could you file a bug report about the crashes you've observed?

sargunv:
These are network requests specific to the demo app, not in the MapLibre Compose library. 

> and makes the Maplibre Composable unusable in production

Did you experience similar crashes with network requests initiated by the library?

PR #834, opened 2026-07-25
Rewrite desktop on maplibre-native-ffi

### LINKED ISSUE #184: The map on desktop is always on top of Compose UI
There's an upstream issue to support SwingPanel interop blending by default: https://youtrack.jetbrains.com/issue/CMP-4941/Make-SwingPanel-interop-blending-by-default

I tried enabling it manually with:

```kt
System.setProperty("compose.interop.blending", "true")
```

but the app crashes on launch with that enabled. Unsure if it's an issue with the property in general or specific to the webview implementation we're using.

### LINKED ISSUE #568: Building maplibre-native from source takes a long time. We should use the prebuilt core.
Since #544, desktop CI is very slow because it's building MapLibre Native from source.

We should investigate caching, especially of the MapLibre Native C++ builds. Or, use the prebuilt native core static library, though I've run into issues with that before.

### LINKED ISSUE #570: Desktop target feature parity
Since #544 we're using the maplibre native c++ core for the desktop target. Functionality is minimal right now.

TODOs to get to feature parity with web:

- [ ] Load a map with Compose resource URIs	
- [ ] Configure gestures (pan, zoom, rotate, pitch)	
- [ ] Respond to a map click or long/right click	
- [ ] Query visible map features	
- [ ] Get, set, and animate the camera position
- [ ] Convert between screen and geographic coordinates
- [ ] Get the currently visible region and bounding box	

TODOs to then get to parity with mobile:

- [ ] Insert, remove, and replace layers
- [ ] Configure layers with expressions
- [ ] Add data sources by URI or GeoJSON	
- [ ] Add images to the style	
- [ ] Add Material 3 controls	
- [ ] Download offline regions

Once we're at parity with Android/iOS, we can look at using the native core directly for those platforms too.

### LINKED ISSUE #595: desktop: isCameraMoving state is not true during gestures
### What happened?

Other camera state fields work, but `isCameraMoving` remains `false` while a gesture is moving the map.

It's because each movement during a gesture is an `IMMEDIATE` mode movement that calls `onCameraWillChange` and `onCameraDidChange` each frame.

We should add separate callbacks for gesture start/end, and use those instead of MapObserver callbacks in immediate mode.

### Affected Platforms

- [ ] Android
- [ ] iOS
- [x] Desktop (JVM)
- [ ] Browser (JS)
- [ ] Browser (Wasm)

### Platform Version

asd

### Library Version

asd

### Sample Code

asd

### LINKED ISSUE #611: EXCEPTION_ACCESS_VIOLATION while loading dll on windows
Originally posted by @haffadan in https://github.com/maplibre/maplibre-compose/discussions/598. I commented there with a request for more info.

```
Host: Intel(R) Core(TM) Ultra 7 155H, 22 cores, 31G,  Windows 11 , 64 bit Build 26100 (10.0.26100.5074)
Time: Mon Sep  8 20:34:55 2025 FLE, sommartid elapsed time: 23.090449 seconds (0d 0h 0m 23s)

---------------  T H R E A D  ---------------

Current thread (0x000002281b0e9100):  JavaThread "AWT-EventQueue-0"        [_thread_in_native, id=12788, stack(0x000000dc20200000,0x000000dc20300000) (1024K)]

Stack: [0x000000dc20200000,0x000000dc20300000],  sp=0x000000dc202fd600,  free space=1013k
Native frames: (J=compiled Java code, j=interpreted, Vv=VM code, C=native code)
C  [msvcp140.dll+0x12f58]  (no source info available)
C  [maplibre-jni.dll+0x3501f7]  (no source info available)
C  [maplibre-jni.dll+0x350420]  (no source info available)
C  [maplibre-jni.dll+0x33a3b]  (no source info available)
C  [java.dll+0x40e5]  (no source info available)
C  0x00000227bf6acfc7  (no source info available)

The last pc belongs to native method entry point (kind = native) (printed below).
Java frames: (J=compiled Java code, j=interpreted, Vv=VM code)
j  jdk.internal.loader.NativeLibraries.load(Ljdk/internal/loader/NativeLibraries$NativeLibraryImpl;Ljava/lang/String;ZZ)Z+0 java.base@22.0.2
j  jdk.internal.loader.NativeLibraries$NativeLibraryImpl.open()Z+57 java.base@22.0.2
j  jdk.internal.loader.NativeLibraries.loadLibrary(Ljava/lang/Class;Ljava/lang/String;Z)Ljdk/internal/loader/NativeLibrary;+254 java.base@22.0.2
j  jdk.internal.loader.NativeLibraries.loadLibrary(Ljava/lang/Class;Ljava/io/File;)Ljdk/internal/loader/NativeLibrary;+51 java.base@22.0.2
j  java.lang.ClassLoader.loadLibrary(Ljava/lang/Class;Ljava/io/File;)Ljdk/internal/loader/NativeLibrary;+31 java.base@22.0.2
j  java.lang.Runtime.load0(Ljava/lang/Class;Ljava/lang/String;)V+61 java.base@22.0.2
j  java.lang.System.load(Ljava/lang/String;)V+7 java.base@22.0.2
j  org.maplibre.kmp.native.util.SharedLibraryLoader.extractAndLoadLibrary(Ljava/lang/String;)V+221
```

### LINKED ISSUE #627: update docs to clarify desktop build (include one capability at a time, not all)
### What happened?

I think u need to embed this lib possible as this issue gone plague on Linux system 
```
Exception in thread "AWT-EventQueue-0" java.lang.UnsatisfiedLinkError: /tmp/maplibre-native-10291293915907124624/libmaplibre-jni.so: libicuuc.so.74: cannot open shared object file: No such file or directory
	at java.base/jdk.internal.loader.NativeLibraries.load(Native Method)
	at java.base/jdk.internal.loader.NativeLibraries$NativeLibraryImpl.open(Unknown Source)
	at java.base/jdk.internal.loader.NativeLibraries.loadLibrary(Unknown Source)
	at java.base/jdk.internal.loader.NativeLibraries.loadLibrary(Unknown Source)
	at java.base/java.lang.ClassLoader.loadLibrary(Unknown Source)
	at java.base/java.lang.Runtime.load0(Unknown Source)
	at java.base/java.lang.System.load(Unknown Source)
	at org.maplibre.kmp.native.util.SharedLibraryLoader.extractAndLoadLibrary(SharedLibraryLoader.kt:71)
	at org.maplibre.kmp.native.util.SharedLibraryLoader.load(SharedLibraryLoader.kt:81)
	at org.maplibre.kmp.native.util.AutoCleanPointer.<clinit>(AutoCleanPointer.kt:19)
	at org.maplibre.kmp.native.map.CanvasRenderer.<init>(CanvasRenderer.kt:16)
	at org.maplibre.kmp.native.map.MapCanvas.addNotify(MapCanvas.kt:44)
	at java.desktop/java.awt.Container.addNotify(Unknown Source)
	at java.desktop/javax.swing.JComponent.addNotify(Unknown Source)
	at java.desktop/java.awt.Container.addImpl(Unknown Source)
	at java.desktop/javax.swing.JLayeredPane.addImpl(Unknown Source)
	at java.desktop/java.awt.Container.add(Unknown Source)
	at androidx.compose.ui.viewinterop.SwingInteropViewHolder.insertInteropView(SwingInteropViewHolder.desktop.kt:125)
	at androidx.compose.ui.viewinterop.SwingInteropContainer.place$lambda$3(SwingInteropContainer.desktop.kt:208)
	at androidx.compose.ui.viewinterop.ScheduledUpdatesSwapchain.execute(SwingInteropContainer.desktop.kt:96)
	at androidx.compose.ui.viewinterop.SwingInteropContainer.executeScheduledUpdates(SwingInteropContainer.desktop.kt:230)
	at androidx.compose.ui.viewinterop.SwingInteropContainer.postponingExecutingScheduledUpdates(SwingInteropContainer.desktop.kt:254)
	at androidx.compose.ui.scene.ComposeSceneMediator.onRender(ComposeSceneMediator.desktop.kt:598)
	at org.jetbrains.skiko.SkiaLayer.update$skiko(SkiaLayer.awt.kt:565)
	at org.jetbrains.skiko.redrawer.AWTRedrawer.update(AWTRedrawer.kt:54)
	at org.jetbrains.skiko.redrawer.LinuxOpenGLRedrawer$Companion$frameDispatcher$1.invokeSuspend(LinuxOpenGLRedrawer.kt:122)
	at kotlin.coroutines.jvm.internal.BaseContinuationImpl.resumeWith(ContinuationImpl.kt:33)
	at kotlinx.coroutines.DispatchedTask.run(DispatchedTask.kt:100)
	at java.desktop/java.awt.event.InvocationEvent.dispatch(Unknown Source)
	at java.desktop/java.awt.EventQueue.dispatchEventImpl(Unknown Source)
	at java.desktop/java.awt.EventQueue$4.run(Unknown Source)
	at java.desktop/java.awt.EventQueue$4.run(Unknown Source)
	at java.base/java.security.AccessController.doPrivileged(Unknown Source)
	at java.base/java.security.ProtectionDomain$JavaSecurityAccessImpl.doIntersectionPrivilege(Unknown Source)
	at java.desktop/java.awt.EventQueue.dispatchEvent(Unknown Source)
	at java.desktop/java.awt.EventDispatchThread.pumpOneEventForFilters(Unknown Source)
	at java.desktop/java.awt.EventDispatchThread.pumpEventsForFilter(Unknown Source)
	at java.desktop/java.awt.EventDispatchThread.pumpEventsForHierarchy(Unknown Source)
	at java.desktop/java.awt.EventDispatchThread.pumpEvents(Unknown Source)
	at java.desktop/java.awt.EventDispatchThread.pumpEvents(Unknown Source)
	at java.desktop/java.awt.EventDispatchThread.run(Unknown Source)
	Suppressed: kotlinx.coroutines.internal.DiagnosticCoroutineContextException: [StandaloneCoroutine{Cancelling}@601813ad, SwingDispatcher@60c9426e]
Exception in thread "AWT-EventQueue-0" java.lang.IllegalStateException: Unknown glyph format: 0
	at java.desktop/sun.font.XRGlyphCacheEntry.getType(Unknown Source)
	at java.desktop/sun.font.XRGlyphCache.separateGlyphTypes(Unknown Source)
	at java.desktop/sun.font.XRGlyphCache.uploadGlyphs(Unknown Source)
	at java.desktop/sun.font.XRGlyphCache.cacheGlyphs(Unknown Source)
	at java.desktop/sun.font.XRTextRenderer.drawGlyphList(Unknown Source)
	at java.desktop/sun.java2d.pipe.GlyphListPipe.drawString(Unknown Source)
	at java.desktop/sun.java2d.pipe.ValidatePipe.drawString(Unknown Source)
	at java.desktop/sun.java2d.SunGraphics2D.drawString(Unknown Source)
	at java.desktop/sun.swing.SwingUtilities2.drawString(Unknown Source)
```

### Affected Platforms

- [ ] Android
- [ ] iOS
- [x] Desktop (JVM)
- [ ] Browser (JS)
- [ ] Browser (Wasm)

### Platform Version

Arch Linux 

### Library Version

0.11.1

### Sample Code

 MaplibreMap(modifier = Modifier.size(300.dp))

### LINKED ISSUE #751: test: verify onMapLoadFinished callback fires (#315)

sargunv:
## Description

Replaces the desktop JNI integration with the `org.maplibre.nativeffi` Kotlin Multiplatform bindings, bringing desktop to near-parity with mobile and laying the foundation for Android and iOS to share the same maplibre-native integration layer in the future.

The line count may seem unexpectedly positive, since we're getting rid of our C++ integration. But of the +11k, about 8k is new test coverage we never had on the old stack, and the code we're getting rid of was missing a _lot_ of desktop functionality that's now implemented.

Bigger reductions will happen when the android and ios (and web?) sides drop their bespoke implementations and build on the same `mlnFfiShared` source set, and then when the owned facade types are replaced with exposing mln-ffi types directly. 

## Test plan

200 new desktop tests, including headless tests with a fake map host and real rendering tests.

Also a new demo app figure that boots the desktop app under compose-glfw for better native fidelity, and proves the custom host (no reflection) path

## How it works

MapLibre renders into a texture that Skia draws as part of the Compose scene, so the map composites like any other composable—overlays, alpha, and transforms apply to it. This requires a bridge between the two graphics APIs (MapLibre supports OpenGL, Vulkan, Metal, while Compose uses OpenGL, D3D, and Metal) and reflection into Compose internals to grab the gpu context we need. An alternative `glfw-fixture/` is a second demo-app on a non-AWT Compose host, which shows how this can be done without reflection as long as a custom compose host exposes its graphics context.

## Caveats

- **Uses Compose internals**, because Compose does not yet expose a way to access the graphics context, and that's the only way to composite the map into the compose scene on desktop. This means even minor compose upgrades may be breaking changes for us.
- **Requires Java 25**, and consumers must pass `--enable-native-access=ALL-UNNAMED`. The binding is Java 24 bytecode and uses FFM.
- **One runtime per platform**, packaged for its OS and architecture. Backend pairs are Vulkan→OpenGL on Linux, Vulkan→Direct3D 12 on Windows, Metal→Metal on macOS. Remaining runtime options will follow.
- **Device location and orientation is out of scope** — the FFI does not provide it, so `DesktopOrientationProvider` remains a stub.

## Relevant issues

- Resolves #184
- Resolves #568
- Resolves #595
- Resolves #627
- Resolves #751
- Resolves #570
- Resolves #611

Partially addresses:

- #29
- #486
- #230
- #738

Not yet addressed, but a prerequisite for:

- #592
- #572
- #550

## Checklist

Yes. The `org.maplibre.compose:maplibre-native-bindings-jni` artifact and its OS/architecture/renderer capabilities are no longer published, the desktop target requires Java 25, and consumers must pass `--enable-native-access=ALL-UNNAMED`.

## AI assistance

- **Tools:** Claude Code w/ Opus 5 for api design and implementation, Codex with GPT 5.6 Sol as advisor/reviewer 
- **Context:** The planning doc the agents worked from is visible in the early commits. I reviewed line by line once I was happy with the overall architecture (well, I still didn't pay _too_ close attention to the tests)


demo-app/src/desktopMain/kotlin/org/maplibre/compose/demoapp/util/Platform.desktop.kt
```diff
@@ -3,17 +3,32 @@ package org.maplibre.compose.demoapp.util
...
 actual object Platform {
   actual val name = System.getProperty("os.name")!!
 
   actual val version = System.getProperty("os.version")!!
 
-  actual val supportedFeatures = emptySet<PlatformFeature>()
+  /**
+   * `InteropBlending` works because MapLibre renders into a texture Skia draws as part of the
+   * Compose scene, so overlays, alpha, and transforms apply to the map like any other composable.
+   */
```
sargunv: superfluous comment

demo-app/src/desktopMain/kotlin/org/maplibre/compose/demoapp/util/Platform.desktop.kt
```diff
@@ -3,17 +3,32 @@ package org.maplibre.compose.demoapp.util
...
+  actual val supportedFeatures = PlatformFeature.Everything
 
-  actual val extraDemos: List<Demo> = listOf(GestureOptionsDemo)
+  actual val extraDemos: List<Demo> =
+    listOf(
+      GestureOptionsDemo,
+      RenderOptionsDemo,
+      // Honored on desktop: mbgl reads synchronousUpdate from the GeoJSON source's own JSON.
```
sargunv: ditto

glfw-fixture/src/desktopMain/kotlin/org/maplibre/compose/glfw/GlfwDesktopMapHostFactory.kt
```diff
@@ -0,0 +1,82 @@
...
+import org.maplibre.compose.desktop.MapRenderBackend
+
+/**
+ * A [DesktopMapHostFactory] backed by a compose-glfw window's own graphics context, reaching into
+ * no AWT or Skiko internals.
+ *
+ * Only the Metal path is implemented; the others are named rather than omitted so the diagnostic
+ * says the bridge is unwritten instead of "no backend".
+ */
+public class GlfwDesktopMapHostFactory(private val renderContext: RenderContext) :
```
sargunv: our api for custom hosts seems overly obtuse. we should only need the gpu context from the host; the rest should be handled internally, and ideally with the same code path on awt vs custom

glfw-fixture/build.gradle.kts
```diff
@@ -0,0 +1,73 @@
...
+  jvmToolchain(properties["jvmToolchain"]!!.toString().toInt())
+
+  jvm("desktop") { compilerOptions { jvmTarget = project.getDesktopJvmTarget() } }
+
+  sourceSets {
+    val desktopMain by getting
+
+    desktopMain.dependencies {
+      // Everything below the SPI line — compose-glfw, LWJGL, the Metal bridge — belongs to this
+      // fixture; neither `:lib:maplibre-compose` nor `:demo-app` may depend on it.
```
sargunv: superfluous

glfw-fixture/build.gradle.kts
```diff
@@ -0,0 +1,73 @@
...
+      implementation(project(":lib:maplibre-compose"))
+      implementation(project(":demo-app"))
+      implementation(libs.composeGlfw)
+      implementation(libs.jetbrains.compose.foundation)
+      implementation(libs.jetbrains.compose.runtime)
+      implementation(libs.jetbrains.compose.ui)
+      // Objective-C messaging for the Metal bridge, the same way the default host does it.
```
sargunv: superfluous

glfw-fixture/build.gradle.kts
```diff
@@ -0,0 +1,73 @@
...
+/**
+ * Keeps AWT's coroutine main dispatcher (pulled in by `:demo-app`) off the fixture's classpath.
+ * `androidx.lifecycle` treats whichever thread `Dispatchers.Main` lands on as the main thread, so
+ * with it present every navigation transition throws `addObserver must be called on the main
+ * thread` from the GLFW thread. With no main dispatcher at all the check permits any thread.
+ */
+configurations.named("desktopRuntimeClasspath") {
+  exclude(group = "org.jetbrains.kotlinx", module = "kotlinx-coroutines-swing")
+}
+
```
sargunv: stale? library should not import coroutines swing anymore I think

gradle/libs.versions.toml
```diff
@@ -4,6 +4,9 @@ accompanistPermissions = "0.37.3"
 alchemist = "0.2.0"
 androidx-activity = "1.11.0"
 bytesize = "2.2.1"
+# The alternative Compose host the desktop SPI fixture runs against. Built against Compose 1.11.1
+# and skiko 0.144.6, so it pins `gradle-compose` below rather than following it.
```
sargunv: cut

gradle/libs.versions.toml
```diff
@@ -13,6 +16,9 @@ lifecycleRuntimeCompose = "2.10.0"
 maplibre-android-sdk = "13.0.2"
 maplibre-android-plugins = "3.0.2"
 maplibre-js = "5.12.0"
+maplibre-nativeFfi = "0.1.0-SNAPSHOT"
+# Keep in sync with the LWJGL version maplibre-native-ffi's Compose example uses.
+lwjgl = "3.4.1"
```
sargunv: example? or binding? we don't need to sync with any example

lib/maplibre-compose/src/desktopMain/kotlin/org/maplibre/compose/desktop/bridge/VulkanOpenGlMapHost.kt
```diff
@@ -0,0 +1,696 @@
+package org.maplibre.compose.desktop.skiko
+
+import androidx.compose.ui.graphics.drawscope.DrawScope
+import org.lwjgl.opengl.EXTMemoryObject.GL_DEDICATED_MEMORY_OBJECT_EXT
```
sargunv, line 4: how does a user get lwjgl natives? transient dependency through our runtime?

lib/maplibre-compose/src/desktopMain/kotlin/org/maplibre/compose/desktop/skiko/LinuxVulkanOpenGlHost.kt
```diff
@@ -0,0 +1,696 @@
...
+
+  override val backends: DesktopBackendPair =
+    DesktopBackendPair(MapRenderBackend.VULKAN, ComposeRenderBackend.OPENGL)
+
+  override fun resize(extent: DesktopMapExtent) {
+    // Importing into GL needs Compose's context current, which only holds inside the draw
+    // callback, so the reallocation happens lazily in acquireFrame. Reallocating on the renderer
+    // thread instead would deadlock: it would wait on the AWT event thread, which is already
+    // waiting on the renderer.
+  }
```
sargunv: we designed this interface in this very pr; the necessity of such a blank method + comment is a smell the interface is the wrong shape

lib/maplibre-compose/src/desktopMain/kotlin/org/maplibre/compose/desktop/DesktopMapHost.kt
```diff
@@ -0,0 +1,128 @@
...
+/**
+ * Bridges MapLibre Native's rendering to a Compose drawing surface.
+ *
+ * A host owns the GPU objects on both sides of the handoff: it allocates the target MapLibre
+ * renders into, synchronizes the producer and consumer, and draws the finished target into Compose.
+ * One host serves one map.
+ *
+ * Implement [DesktopMapHostFactory] to supply your own; see [LocalDesktopMapHostFactory].
+ */
+public interface DesktopMapHost : AutoCloseable {
```
sargunv: might make sense for the interface to not be desktop-specific? assuming a similar contract will need to be implemented on android/ios later

ditto for everything Desktop prefixed. is it really Desktop or is it mln-ffi

maybe an mlnffiShared source set holding most of our code to make it explicit, desktop only consumer for _now_

lib/maplibre-compose/src/mlnFfiShared/kotlin/org/maplibre/compose/mlnffi/MlnFfiMapRenderer.kt
```diff
@@ -0,0 +1,52 @@
...
+}
+
+/**
+ * Receives surface lifecycle and frames from a [DesktopMapHost].
+ *
+ * MapLibre Compose implements this to drive a map. Applications implement [DesktopMapHostFactory],
+ * not this.
+ *
+ * Every method is called with renderer access held, on the host's renderer thread.
+ */
```
sargunv, line 19: why an interface, especially a public interface, if applications don't implement this

lib/maplibre-compose/src/desktopMain/kotlin/org/maplibre/compose/desktop/DesktopRuntimeOptions.kt
```diff
@@ -0,0 +1,55 @@
...
+private fun defaultCachePath(): Path {
+  val os = System.getProperty("os.name")?.lowercase().orEmpty()
+  val home = System.getProperty("user.home") ?: "."
+  val base =
+    when {
+      os.contains("mac") -> Paths.get(home, "Library", "Caches")
+      os.contains("windows") ->
+        System.getenv("LOCALAPPDATA")?.let(Paths::get) ?: Paths.get(home, "AppData", "Local")
+      // Linux and anything else: the XDG base directory spec, with its documented fallback.
+      else -> System.getenv("XDG_CACHE_HOME")?.let(Paths::get) ?: Paths.get(home, ".cache")
```
sargunv: why not respect xdg paths on other platforms?

lib/maplibre-compose/src/desktopMain/kotlin/org/maplibre/compose/layers/FeatureLayer.kt
```diff
@@ -1,11 +1,36 @@
...
+internal actual sealed class FeatureLayer(id: String, actual val source: Source) : Layer(id) {
+
+  /**
+   * Exists only to match the common `expect` constructor; throws, because a layer with no id cannot
+   * be added to a style. Subclasses use the id-carrying constructor above.
+   */
+  protected actual constructor(source: Source) : this("", source) {
+    error("A desktop feature layer must be constructed with its layer id")
+  }
+
```
sargunv: where is it called? should we delete if it's dead?

lib/maplibre-compose/src/mlnFfiShared/kotlin/org/maplibre/compose/layers/Layer.kt
```diff
@@ -1,28 +1,241 @@
...
+  protected open val sourceId: String? = null
+
+  /**
+   * The source descriptor this layer draws from, when it has one.
+   *
+   * The C API rejects a layer naming a source that does not exist yet, and Compose adds a layer
+   * before the effect that adds its source runs, so [attach] attaches this source first.
+   */
+  internal open val sourceDescriptor: org.maplibre.compose.sources.Source?
+    get() = null
```
sargunv, line 43: where is it called or overridden? delete if dead? or at least TODO it if android/ios still depend on it

lib/maplibre-compose/src/desktopMain/kotlin/org/maplibre/compose/map/DesktopMapInput.kt
```diff
@@ -0,0 +1,248 @@
...
+import androidx.compose.ui.input.pointer.isSecondaryPressed
+import androidx.compose.ui.input.pointer.isShiftPressed
+import androidx.compose.ui.input.pointer.pointerInput
+import androidx.compose.ui.unit.Density
+import androidx.compose.ui.unit.DpOffset
+import androidx.compose.ui.unit.dp
+import kotlin.time.Duration.Companion.milliseconds
+
+/** How far a press may move, in logical pixels, and still count as a click rather than a drag. */
+private const val CLICK_SLOP_PX = 4f
```
sargunv: does compose have platform-specific helpers for this? I'd expect compose to handle click vs drag for us with platform-specific slop

lib/maplibre-compose/src/desktopMain/kotlin/org/maplibre/compose/map/DesktopMapInput.kt
```diff
@@ -0,0 +1,248 @@
...
+
+/** How far the camera moves per keyboard pan, in logical pixels. */
+private const val KEYBOARD_PAN_STEP = 100.0
+
+/** Zoom applied per keyboard step and per double click, as a scale multiplier. */
+private const val KEYBOARD_ZOOM_STEP = 2.0
```
sargunv: maybe into gesture options? or a TODO if android/ios share those today

lib/maplibre-compose/src/desktopMain/kotlin/org/maplibre/compose/map/DesktopMapInput.kt
```diff
@@ -0,0 +1,248 @@
...
+ * GL JS. A repeat supersedes the transition still in flight, so a held arrow key pans continuously.
+ */
+private val INPUT_ANIMATION_DURATION = 300.milliseconds
+
+/**
+ * Zoom exponent per unit of scroll, applied as `2^(-scroll * factor)`. Matches the
+ * maplibre-native-ffi Compose example.
+ */
+private const val SCROLL_ZOOM_FACTOR = 0.25
+
```
sargunv: should figure out typical standards for this (maplibre gl js on web is maybe a good reference). the other defaults too. rather than the compose-map example in mln-ffi repo.

lib/maplibre-compose/src/desktopMain/kotlin/org/maplibre/compose/map/DesktopMapInput.kt
```diff
@@ -0,0 +1,248 @@
...
+      }
+    }
+    .focusRequester(focusRequester)
+    .focusable()
+    .pointerInput(session, options, density) {
+      // MapLibre keeps the gesture flag until it is cleared, so a drag ended by coroutine
+      // cancellation rather than a pointer-up must clear it in the finally below.
+      var dragging = false
+      try {
+        awaitPointerEventScope {
```
sargunv: gesture code is extremely hard to follow

that sort of comes with the territory but should make a readability pass at it anyway

lib/maplibre-compose/src/mlnFfiShared/kotlin/org/maplibre/compose/map/MlnFfiMapRuntimeLoop.kt
```diff
@@ -0,0 +1,298 @@
...
+
+/**
+ * How long a park lasts before the loop pumps regardless of any wake.
+ *
+ * A backstop, not the cadence: run-loop timers and file descriptors set the wake flag only when
+ * they queue owner-thread work, so anything armed directly on the owner run loop needs this bound
+ * to be serviced at all.
+ */
+private const val PUMP_PARK_MILLIS = 100L
+
```
sargunv, line 26: "anything armed directly on the owner run loop" like what?

a 100ms backstop is a smell; why can't we pump on wake only?

lib/maplibre-compose/src/mlnFfiShared/kotlin/org/maplibre/compose/map/MlnFfiMapRuntimeLoop.kt
```diff
@@ -0,0 +1,298 @@
...
+/** Bound on waiting for the render session to be closed before the map is destroyed. */
+private const val SHUTDOWN_WAIT_MILLIS = 5_000L
```
sargunv, line 28: in what scenario would it stall?

lib/maplibre-compose/src/mlnFfiShared/kotlin/org/maplibre/compose/map/MlnFfiMapRuntimeLoop.kt
```diff
@@ -0,0 +1,298 @@
...
+ * The map is deliberately not on the presenting thread: the loop parks inside the native pump, so
+ * style parsing, tile loads, and resource responses advance whether or not anything is drawing.
+ * Camera transitions are the exception — mbgl steps them from `onDidFinishRenderingFrame`, so
+ * moving the camera still takes a rendered frame.
+ *
+ * The render session is not this loop's: since maplibre-native-ffi #399 a session's owner is
+ * whichever thread attached it, and native refuses to destroy a map that still has one attached, so
+ * teardown waits for the presenting thread to close it.
+ *
+ * A runtime belongs to the thread that created it and there may be only one per thread, so this is
```
sargunv, line 43: trim historical comments, describe today's state only. though you can reference https://github.com/maplibre/maplibre-native-ffi/pull/433 if it would make this model obsolete.

lib/maplibre-compose/src/desktopMain/kotlin/org/maplibre/compose/map/DesktopMapSession.kt
```diff
@@ -0,0 +1,1213 @@
...
+    }
+  }
+
+  private var maximumFps: Int? = null
+  private var lastRenderTime = TimeSource.Monotonic.markNow()
+
+  private val frameTimer = TimeSource.Monotonic
+  private var lastFrameTime = frameTimer.markNow()
+
+  // ───────────────────────────── host surface lifecycle ─────────────────────────────
```
sargunv: use IDE compatible region comments rather than these.

lib/maplibre-compose/src/desktopMain/kotlin/org/maplibre/compose/map/DesktopMapSession.kt
```diff
@@ -0,0 +1,1213 @@
...
+      // has never seen".
+      RuntimeEventType.MAP_IDLE,
+      RuntimeEventType.MAP_LOADING_STARTED,
+      RuntimeEventType.MAP_RENDER_FRAME_STARTED,
+      RuntimeEventType.MAP_RENDER_MAP_STARTED,
+      RuntimeEventType.MAP_RENDER_MAP_FINISHED,
+      RuntimeEventType.MAP_STILL_IMAGE_FINISHED,
+      RuntimeEventType.MAP_STILL_IMAGE_FAILED,
+      RuntimeEventType.MAP_TILE_ACTION -> Unit
+
```
sargunv: not acted on why?

lib/maplibre-compose/src/desktopMain/kotlin/org/maplibre/compose/map/DesktopMapSession.kt
```diff
@@ -0,0 +1,1213 @@
...
+    animate(duration) { map, animation ->
+      map.flyTo(cameraForBounds(map, boundingBox, bearing, tilt, padding), animation)
+    }
+  }
+
+  /**
+   * Starts a camera transition and suspends until MapLibre reports that it released the camera.
+   * Resumes normally however the transition ended, as Android's `CancelableCallback.onCancel` does.
+   */
+  private suspend fun animate(
```
sargunv: name it something that indicates more clearly "suspends until MapLibre reports that it released the camera"

lib/maplibre-compose/src/desktopMain/kotlin/org/maplibre/compose/offline/DesktopOfflineManager.kt
```diff
@@ -0,0 +1,383 @@
...
+    }
+    pack.progressState.value = progress
+  }
+
+  private fun publishPacks() {
+    // Snapshotted so Compose never sees the mutable map behind it.
+    packsState.value = packsById.values.toSet()
+  }
+
+  // ───────────────────────────── operation plumbing ─────────────────────────────
```
sargunv: ditto, use ide compatible region blocks for stuff like this

lib/maplibre-compose/src/mlnFfiShared/kotlin/org/maplibre/compose/offline/MlnFfiOfflineRuntime.kt
```diff
@@ -0,0 +1,313 @@
...
+import org.maplibre.nativeffi.runtime.RuntimeHandle
+import org.maplibre.nativeffi.runtime.WakeSource
+
+/**
+ * How long a park lasts before the loop pumps regardless of any wake. A backstop rather than the
+ * cadence: timers and ready sockets set the wake flag only when they queue owner-thread work, so a
+ * download waiting out a retry has nothing to signal with.
+ */
+private const val PUMP_PARK_MILLIS = 100L
+
```
sargunv, line 21: ditto other runtime park comment

lib/maplibre-compose/src/mlnFfiShared/kotlin/org/maplibre/compose/offline/util.kt
```diff
@@ -0,0 +1,151 @@
...
+import org.maplibre.spatialk.geojson.Polygon
+import org.maplibre.spatialk.geojson.Position
+
+/**
+ * Whether downloaded packs include CJK glyphs.
+ *
+ * True on desktop, unlike Android and iOS, because those platforms render ideographs from a local
+ * system font and MapLibre Native's desktop map has no local-font option.
+ */
+private const val INCLUDE_IDEOGRAPHS = true
```
sargunv, line 31: "on desktop" or "on mln ffi"?

lib/maplibre-compose/src/mlnFfiShared/kotlin/org/maplibre/compose/util/conversions.kt
```diff
@@ -1,63 +1,70 @@
...
 
-internal fun CameraPosition.toMlnCameraOptions(layoutDirection: LayoutDirection) =
-  MLNCameraOptions.centered(
-    center = target.toMlnLatLng(),
-    bearing = bearing,
-    pitch = tilt,
-    zoom = zoom,
-    padding = padding.toMlnEdgeInsets(layoutDirection),
-  )
+internal fun LatLng.toPosition(): Position = Position(longitude = longitude, latitude = latitude)
```
sargunv, line 24: a TODO, eventually our commonMain public types should just re-export the mln-ffi types (when mln-ffi adds js and web)

lib/maplibre-compose/build.gradle.kts
```diff
@@ -85,8 +84,15 @@ kotlin {
...
-        implementation(project(":lib:maplibre-native-bindings"))
+        // Backend-independent binding only; the application selects the native runtime.
+        implementation(libs.maplibre.nativeFfi)
+
+        // The default Skiko host needs direct Vulkan/OpenGL access; the natives are the app's
+        // concern.
+        implementation(libs.lwjgl.core)
+        implementation(libs.lwjgl.egl)
+        implementation(libs.lwjgl.opengl)
+        implementation(libs.lwjgl.vulkan)
```
sargunv: between these natives and the mln-ffi runtime, we should consider publishing our own runtime meta-package, or maybe a bom

sargunv:
@greptile-apps review

demo-app/build.gradle.kts
```diff
@@ -34,9 +34,13 @@ android {
   testOptions { animationsDisabled = true }
 }
 
+val desktopHostPlatform = DesktopHostPlatform.current()
```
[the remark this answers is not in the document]
sargunv: intel mac is out of scope.

mise.toml
```diff
@@ -3,27 +3,40 @@ experimental = true
...
+"aqua:koalaman/shellcheck" = "0.11.0"
 xcodes = { version = "1.6.2", os = ["macos"] }
 swiftformat = "0.61.0"
 python = "3.14.4"
 
 [hooks]
 postinstall = [
-  "hk install --mise",
-  "mise run _xcodes-select",
+  { run = "hk install --mise" },
```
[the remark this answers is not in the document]
sargunv: that version is way old, and the latest version doesn't work with that old form anymore.

DestBro:
## Validated Bug Report: maplibre-compose Desktop FFI Rewrite

Five critical bugs were confirmed in the `desktop-ffi-rewrite` branch by cross-referencing the initial analysis with the actual source code. All five affect thread safety, memory management, or application stability.

---

### Bug 1: Thread-Safety Violation in `Direct3D12Presenter`

`Direct3D12Presenter` manages a collection of `TexturePresenter` objects wrapping Skia and Direct3D resources.

The `presenters` map is declared as a plain `mutableMapOf<Long, TexturePresenter>()`. It is accessed on the **UI thread** in `draw()` via `presenters.getOrPut(target.texture.address)`, and concurrently accessed and modified on the **GPU thread** in `forget()` and `close()` via `gpuHost.runOnGpuThread`.

**Impact:** Concurrent modification of a non-thread-safe map causes `ConcurrentModificationException` or silent memory corruption.

**Location:** `Direct3D12Presenter.kt`

---

### Bug 2: Memory Visibility Race in `VulkanDirect3D12MapHost`

`currentExtent` tracks the size of the render target. It is **read on the UI thread** inside `acquireFrame()` to decide whether a resize is needed, and **written on the Renderer thread** inside `resizeOnRendererThread()`. The property is not marked `@Volatile` and there is no synchronization between the two threads.

**Impact:** The UI thread may read a stale value, causing missed resizes, redundant allocations, or swapchain "out of date" errors.

**Location:** `VulkanDirect3D12MapHost.kt` — `acquireFrame()` (read) and `resizeOnRendererThread()` (write)

---

### Bug 3: Native Handle Use-After-Free in `MlnFfiResourceProvider`

The shutdown sequence in `MlnFfiRuntimeOwner.close()` first closes `MlnFfiResourceProvider`, then closes the `RuntimeHandle`. `MlnFfiResourceProvider.close()` shuts down its reader thread and waits up to `QUIESCE_TIMEOUT_SECONDS` (5 s) for pending reads to finish. If a read exceeds that timeout, `close()` proceeds and the `RuntimeHandle` is subsequently destroyed. The background reader thread may still be executing and eventually call `handle.complete(response)` on a `ResourceRequestHandle` belonging to the now-closed runtime.

**Impact:** Use-after-free in native code, causing segmentation faults during application shutdown.

**Location:** `MlnFfiResourceProvider.kt` — background reader thread calling `handle.complete()`; `MlnFfiRuntimeOwner.kt` — shutdown ordering

---

### Bug 4: Resource and Thread Leak in `MlnFfiOfflineManager`

`MlnFfiOfflineManager` caches instances globally, keyed by `MlnFfiRuntimeOptions`. If an application creates multiple maps with varying options (e.g. different `cachePath` or `maximumCacheSizeBytes`), a new manager is created for each combination and **never removed**. Each manager spins up its own `MlnFfiOfflineRuntime` thread and a native MapLibre runtime.

**Impact:** Severe memory and thread leak. Every leaked manager holds native memory and a dedicated background thread for the lifetime of the process.

**Location:** `MlnFfiOfflineManager.kt` — `instances` map with no eviction

---

### Bug 5: Potential Deadlock in `MlnFfiMapSession.stopLoop()`

When a map is disposed, `MlnFfiMapSession.close()` calls `stopLoop()`. `stopLoop()` calls `closeRenderSession()`, which calls `host.withRendererAccess { handle.close() }`, blocking the calling thread until the Renderer thread executes the closure. If the Renderer thread is simultaneously blocked waiting for the Owner thread (e.g. during a synchronous hop), and the Owner thread is blocked waiting for the Renderer thread during teardown, a circular wait forms.

**Impact:** UI freeze or application hang during map disposal.

**Location:** `MlnFfiMapSession.kt` — `stopLoop()` / `closeRenderSession()`

DestBro:
## Validated Bugs

---

### Bug 1: Data Race on `renderSession` in `MlnFfiMapSession.closeRenderSession()`

**Severity:** Critical — potential double-close or use-after-close of a native render session handle

`renderSession` is documented as **renderer-thread state**, but `closeRenderSession()` reads and nullifies it on the **Compose UI thread (EDT)** — called from `onDispose` via `renderer.close()` — while `ensureAttached()` concurrently writes it on the **Renderer thread**. The field is not `@Volatile` and there is no lock protecting it.

Under the Java Memory Model this is a confirmed data race with two failure modes:

- The EDT sees a stale `null` → the live session handle is never closed → native session leak
- The EDT sees a stale non-null handle → it is closed on the Renderer thread while the Renderer thread is still using it → use-after-close / crash

| Location | Thread | Access |
|---|---|---|
| `MlnFfiMapSession.kt:139` | — | field declaration (`var`, not `@Volatile`) |
| `MlnFfiMapSession.kt:451, 458` | Renderer thread | **write** (`ensureAttached`) |
| `MlnFfiMapSession.kt:350–352` | EDT (`onDispose`) | **read + nullify** (`closeRenderSession`) |

**Suggested fix:** Read and nullify `renderSession` inside `host.withRendererAccess { }` rather than on the calling thread, so all accesses to the field are confined to the Renderer thread.

---

### Bug 2: `MapRendererThread.close()` Does Not Drain In-Flight Tasks Before Vulkan Teardown

**Severity:** High — native use-after-free in the Vulkan driver during window close

`MapRendererThread.close()` calls `executor.shutdown()`, which initiates shutdown but **does not wait** for already-submitted tasks to finish (`awaitTermination` is never called). In `VulkanDirect3D12MapHost.close()`, the Vulkan device is destroyed (`closingVulkan?.close()`) **before** `rendererThread.close()` is called:

```kotlin
// VulkanDirect3D12MapHost.close()
closingVulkan?.close()      // Vulkan device destroyed here
rendererThread.close()      // executor.shutdown() — non-blocking, tasks may still run
```

If a `Canvas` draw cycle submits a task to the renderer thread (e.g. via `withProducerAccess`) in the narrow window after `renderer.close()` returns but before `host.close()` completes, that task executes against an already-destroyed Vulkan device, causing a native use-after-free.

| Location | Issue |
|---|---|
| `MapRendererThread.kt:39` | `executor.shutdown()` is non-blocking — does not drain the queue |
| `VulkanDirect3D12MapHost.kt:183–188` | `vulkan.close()` called before `rendererThread.close()` |

**Suggested fix:** Replace `executor.shutdown()` with `executor.shutdown(); executor.awaitTermination(...)` in `MapRendererThread.close()`, or invert the teardown order so `rendererThread.close()` (and drain) precedes `vulkan.close()`.

demo-app/src/commonMain/kotlin/org/maplibre/compose/demoapp/design/FrameRateListItem.kt
```diff
@@ -0,0 +1,32 @@
...
+import androidx.compose.material3.Text
+import androidx.compose.runtime.Composable
+import androidx.compose.runtime.LaunchedEffect
+import androidx.compose.ui.Modifier
+import kotlin.math.roundToInt
+import kotlin.math.roundToLong
+import org.maplibre.compose.demoapp.FrameRateState
+
+/**
+ * Shows how fast the map is actually drawing. The number to expect is zero: frames are requested
```
sargunv: Not true when animated layers are enabled

.github/actions/setup-desktop-vulkan/action.yml
```diff
@@ -0,0 +1,95 @@
...
+          mesa-utils-extra \
+          mesa-vulkan-drivers
+        {
+          echo "EGL_PLATFORM=surfaceless"
+          echo "LIBGL_ALWAYS_SOFTWARE=true"
+        } >> "$GITHUB_ENV"
+
+    # macOS ships no Vulkan loader at all, so both halves come from Homebrew: the loader, and
+    # MoltenVK as the ICD behind it. This is what `mise run bootstrap` installs for contributors.
+    - name: Install the Vulkan loader and MoltenVK (macOS)
```
sargunv: Verify if this is actually necessary, lwjgl provides some of it at least

.github/actions/setup-desktop-vulkan/action.yml
```diff
@@ -0,0 +1,95 @@
+name: Setup desktop Vulkan
+description: >-
+  Give the runner a Vulkan implementation the desktop GPU tests can drive.
+
+# The desktop test suite renders through a real Vulkan device (HeadlessVulkanMapHost), on every
```
sargunv: Trim narration comments

sargunv:
on MlnFfiOfflineManager retention, it's a minor hazard for now; it's not a leak on every map as the ai claimed, but on every _unique_ set of options passed to `rememberOfflineManager`. A typical app will never run into this, or at perhaps exactly one instance. Disposing of it with DisposableEffect would cancel downloads unexpectedly.

Should be replaced with an offline manager held outside the composition entirely; part of a future api redesign once we're off the maplibre android and ios sdks, just like the map runtime / render separation.

lib/maplibre-compose/src/mlnFfiShared/kotlin/org/maplibre/compose/resource/MlnFfiResourceProvider.kt
```diff
@@ -0,0 +1,234 @@
...
+        Thread.currentThread().interrupt()
+        logger?.w(interruption) { "Interrupted while quiescing resource reads" }
+        false
+      }
+    if (!quiesced) {
+      logger?.e {
+        "Resource reads did not finish within ${QUIESCE_TIMEOUT_SECONDS}s; " +
+          "closing the runtime with ${reader.queue.size + reader.activeCount} still outstanding"
+      }
+    }
```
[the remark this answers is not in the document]
sargunv: invalid, but updated comments to clarify

gradle/libs.versions.toml
```diff
@@ -4,15 +4,18 @@ accompanistPermissions = "0.37.3"
...
 kermit = "2.1.0"
 kotlin-wrappers = "2026.4.13"
 kotlinx-coroutines = "1.10.2"
 ktor = "3.4.3"
 lifecycleRuntimeCompose = "2.10.0"
+lwjgl = "3.4.1"
 maplibre-android-sdk = "13.0.2"
 maplibre-android-plugins = "3.0.2"
 maplibre-js = "5.12.0"
+maplibre-nativeFfi = "0.1.0-SNAPSHOT"
```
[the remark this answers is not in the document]
sargunv: invalid; mln-ffi will be published before merge.

lib/maplibre-compose/build.gradle.kts
```diff
@@ -100,6 +114,31 @@ kotlin {
...
+    create("mlnFfiSharedTest") {
+      dependsOn(commonTest.get())
+      getByName("desktopTest").dependsOn(this)
+    }
+
+    // Runtime dependencies belong to platform/backend adapters. One native runtime is loaded per
+    // test process; a CI matrix adds processes for additional applicable backends.
+    val desktopTest by getting
+    desktopTest.dependencies {
+      val platform = DesktopHostPlatform.current()
```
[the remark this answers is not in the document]
sargunv: no

lib/maplibre-compose/src/mlnFfiShared/kotlin/org/maplibre/compose/map/MlnFfiMapSession.kt
```diff
@@ -0,0 +1,1282 @@
...
+    val host = hostSession ?: return null
+    return host.withRendererAccess(action)
+  }
+
+  // endregion
+
+  // region MapAdapter
+
+  override fun setBaseStyle(style: BaseStyle) {
+    if (style == requestedStyle) return
```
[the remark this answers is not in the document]
sargunv: This deduplication is intentional and matches the Android, iOS, and JS adapters. `MlnFfiMapView` calls `setBaseStyle` from a `SideEffect` after every recomposition, so treating every repeated equal value as a retry can turn a load-failure callback that updates Compose state into an unbounded retry loop. A deliberate retry needs an explicit cross-platform API rather than desktop-only equality semantics; I’ll correct the misleading internal comment separately if we touch this area.

lib/maplibre-compose/src/mlnFfiShared/kotlin/org/maplibre/compose/mlnffi/MlnFfiMapSurface.kt
```diff
@@ -0,0 +1,331 @@
...
+    try {
+      host.resize(extent)
+      renderer.onSurfaceChanged(extent)
+      session?.requestFrame()
+    } catch (error: Throwable) {
+      state =
+        MlnFfiMapSurfaceState.Failed(
+          "Map host failed to resize to ${extent.width}x${extent.height}",
+          error,
+        )
```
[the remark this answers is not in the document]
sargunv: Fixed in 21973ec1. A resize or `onSurfaceChanged` failure now closes the renderer immediately through the idempotent draw-state cleanup path, and the regression test verifies that a later resize failure closes it exactly once.

lib/maplibre-compose/src/mlnFfiShared/kotlin/org/maplibre/compose/mlnffi/MlnFfiRuntimeOptions.kt
```diff
@@ -0,0 +1,61 @@
...
+  val home = System.getProperty("user.home") ?: "."
+  val base =
+    System.getenv("XDG_CACHE_HOME")?.takeIf { it.isNotBlank() }?.let(Paths::get)
+      ?: when {
+        os.contains("mac") -> Paths.get(home, "Library", "Caches")
+        os.contains("windows") ->
+          System.getenv("LOCALAPPDATA")?.let(Paths::get) ?: Paths.get(home, "AppData", "Local")
+        else -> Paths.get(home, ".cache")
+      }
+  return base.resolve("maplibre-compose").resolve("maplibre-cache.db")
```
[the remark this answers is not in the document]
sargunv: Fixed in 667736a8. Desktop no longer has a process-independent default: `ProvideMapHost` requires `DesktopRuntimeOptions`, `cachePath` is required, and `desktopCachePath(applicationId)` provides an OS-aware application-scoped convenience path. The shared FFI options are now internal, and the demo plus getting-started guide supply explicit application IDs.

sargunv:
@greptile-apps review

PR #886, opened 2026-08-11
Remove platform SDK ornaments; move Compose controls into the core

### LINKED ISSUE #870: Replace ornaments with dependency-free compose-based maplibre logo and attribution
We have material3 components for map controls: compass, scale bar, location button, pointer pin, and attribution. Historically, we relied on MapLibre Native's ornaments to provide a default attribution and logo with no Material 3 dependency.

Now, desktop and web are fleshed out feature wise, and without any default attribution/logo. Most maps will require attribution under OSM guidelines, so we should include a simple implementation by default available in the base library with no additional dependencies.

This is more important as #572 lands. We should not ship a release wherre #572 is implemented for Android or iOS and a default attribution ui isn't.

sargunv:
## Description

Removes `OrnamentOptions` and gives `MaplibreMap` an `overlay` slot that draws dependency-free map controls by default, plus a `contentWindowInsets` that keeps them clear of system bars and sheets.

The attribution, compass, and scale bar components move down from `maplibre-compose-material3` into `maplibre-compose`, rebuilt on Compose Foundation; the Material 3 versions keep their names and become wrappers that supply colors, typography, and widgets.

Android and iOS hard-disable their native ornament views, which drops the native compass and scale bar. They'll be removed entirely by #572.

Also filled out the missing mise tasks for building/running on all platforms.

Closes #870.

## Test plan

Ran the demo on desktop. Added a small test that the logo xml is successfully parsed.


lib/maplibre-compose/src/commonMain/kotlin/org/maplibre/compose/overlay/Attribution.kt
```diff
@@ -0,0 +1,344 @@
...
+  contentAlignment: Alignment = Alignment.BottomEnd,
+  toggleButton: @Composable (onClick: () -> Unit) -> Unit = AttributionDefaults.button,
+  expandedContent: @Composable (attributions: List<String>, textStyle: TextStyle) -> Unit =
+    AttributionDefaults.content,
+  expandedStyle: AttributionStyle = AttributionDefaults.expandedStyle(),
+  collapsedStyle: AttributionStyle = AttributionDefaults.collapsedStyle(),
+  expand: (Alignment) -> EnterTransition = AttributionDefaults.expand,
+  collapse: (Alignment) -> ExitTransition = AttributionDefaults.collapse,
+) {
+  val attributions by remember { derivedStateOf { styleState.attributions() } }
```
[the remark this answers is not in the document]
sargunv: a62091ed

lib/maplibre-compose/src/commonMain/kotlin/org/maplibre/compose/overlay/CompassButton.kt
```diff
@@ -0,0 +1,220 @@
...
+  enterTransition: EnterTransition = fadeIn(),
+  exitTransition: ExitTransition = fadeOut(),
+  getHomePosition: (CameraPosition) -> CameraPosition = { it.copy(bearing = 0.0, tilt = 0.0) },
+  slop: Double = 0.5,
+) {
+  val visible = remember { MutableTransitionState(false) }
+
+  val homePosition by remember { derivedStateOf { getHomePosition(cameraState.position) } }
+
+  val shouldBeVisible by remember {
```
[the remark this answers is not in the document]
sargunv: a62091ed

lib/maplibre-compose/src/commonMain/kotlin/org/maplibre/compose/overlay/CompassButton.kt
```diff
@@ -0,0 +1,220 @@
...
+
+  AnimatedVisibility(
+    visibleState = visible,
+    modifier = modifier,
+    enter = enterTransition,
+    exit = exitTransition,
+  ) {
+    CompassButton(
+      cameraState = cameraState,
+      modifier = modifier,
```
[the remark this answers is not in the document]
sargunv: a62091ed

lib/maplibre-compose/src/commonMain/composeResources/values-zh-rTW/strings.xml
westnordost, line 1: Note that starting with [Compose Multiplatform 1.12.0-beta01](https://github.com/JetBrains/compose-multiplatform/releases/tag/v1.12.0-beta01) (1.12.0-rc01 came out just now), it is finally possible to have strings.xml files in `values-b+zh+Hans` and `values-b+zh+Hant`, so `values-zh-rTW`, `values-zh-rHK` will not be necessary anymore.

Also, a `values-b+sr-Cyrl/strings.xml` and `values-b+sr-Latn/strings.xml` ca be added. (Serbian is the only? language which is commonly written in both Latin and Cyrillic scripts)

lib/maplibre-compose/src/commonMain/composeResources/drawable/maplibre_logo.xml
westnordost, line 1: maybe add a comment that points to the source of this vector drawable? (probably an SVG somewhere in the maplibre-native repository?)

lib/maplibre-compose/src/nextCommonTest/kotlin/org/maplibre/compose/overlay/MaplibreLogoTest.kt
westnordost, line 1: Why would it not? You seem to be testing core Compose Multiplatform functionality.

lib/maplibre-compose/src/commonMain/kotlin/org/maplibre/compose/overlay/MaplibreLogo.kt
```diff
@@ -0,0 +1,49 @@
...
+      modifier.then(
+        onClick?.let { Modifier.clickable(onClick = it, role = Role.Image) } ?: Modifier
+      ),
+  )
+}
+
+@Composable
+internal fun rememberOpenWebsite(): () -> Unit {
+  val uriHandler = LocalUriHandler.current
+  return remember(uriHandler) { { uriHandler.openUri(MaplibreWebsiteUrl) } }
```
westnordost: This crashes if no browser is installed. Is this intended? In my app, I basically replaced all occurences of `openUri` for links out of the app with 

```kt
/**
 * Try to open given URL in browser. Returns `false` if given [uri] is invalid and/or can't be
 * handled by the system (e.g. if no browser is installed).
 */
fun UriHandler.tryOpenUri(uri: String): Boolean =
    try {
        openUri(uri)
        return true
    } catch (e: IllegalArgumentException) {
        return false
    }
```
because  when the user has no browser installed, he will not be surprised that links don't open. While clicking on (buttons and) links may sometimes happen inadvertently (while interacting with the map).

lib/maplibre-compose/src/commonMain/kotlin/org/maplibre/compose/overlay/MapOverlay.kt
westnordost, line 1: I don't quite understand why this is necessary at all. (I.e. why the MapLibreMap has an `overlay` parameter at all.) The overlay content is not "in" the map, it's just on top. So, the overlay content could just be specified as siblings of the `MapLibreMap`, no?

lib/maplibre-compose-material3/src/commonMain/kotlin/org/maplibre/compose/material3/util/measures.kt
```diff
@@ -78,14 +83,3 @@ internal val regionsUsingYardsAndMiles =
...
-/**
- * default scale bar measures to use, depending on the user's locale (or system preferences, if
- * available)
- */
-@Composable
-public fun defaultScaleBarMeasures(): ScaleBarMeasures {
-  val region = Locale.current.region
-  val primary = systemDefaultPrimaryMeasure() ?: fallbackDefaultPrimaryMeasure(region)
-  return ScaleBarMeasures(primary = primary, secondary = defaultSecondaryMeasure(primary, region))
-}
```
westnordost, line 91: Was this deleted without replacement or did I miss it?

westnordost (commented):
Github doesn't show the diff in a way that one can see the difference between the erstwhile material3 components and the base-variants. I'll just assume that nothing other than default parameters have been changed.

lib/maplibre-compose/src/commonMain/kotlin/org/maplibre/compose/overlay/MapOverlay.kt
sargunv: It's only necessary if we want attribution and logo visible by default, like the other MapLibre SDKs. I'm not strongly committed to that, but I do think most people will use OSM data and won't really think about the OSM attribution if we don't show it by default.

lib/maplibre-compose/src/nextCommonTest/kotlin/org/maplibre/compose/overlay/MaplibreLogoTest.kt
sargunv: Compose Multiplatform has its own parser for android vector drawables, and in some cases it rejects input that android would accept. I copied this XML from MapLibre Android, so added a small test proving it's also accepted by non-android compose.

lib/maplibre-compose/src/commonMain/composeResources/values-zh-rTW/strings.xml
sargunv: filed https://github.com/maplibre/maplibre-compose/issues/887

lib/maplibre-compose/src/commonMain/kotlin/org/maplibre/compose/overlay/MapOverlay.kt
westnordost: Yeah, the attribution is important.

My take on the logo stuff is that it is superfluous. The only reason why it is included at all in the SDKs I'd say is because it was included in the MapBox SDKs, not for any license reasons.

As to why MapBox does it: I think it is because they intended that MapBox maps also use tiles from MapBox (just like Google Maps uses data from Google) and for visibility, of course. But I don't see any "Jetpack Compose" logo in the corner of every app that uses that UI framework either, in my opinion it doesn't belong there by default.
But I can understand that MapLibre, the organization, (also) wants visibility, and to show that by default certainly helps.

sargunv:
> Github doesn't show the diff in a way that one can see the difference between the erstwhile material3 components and the base-variants. I'll just assume that nothing other than default parameters have been changed.

That plus the components were reimplemented on Foundation primitives rather than M3 ones.  Some style defaults changed bc no dynamic theme, so the non-M3 compass, attribution are translucent, and the scale bar has a halo by default.

The Material3 versions retain their existing Material style. 

Some simulated diff snippets:

### ScaleBar

```patch
- measures = defaultScaleBarMeasures()
- color = LocalContentColor.current
- haloColor = backgroundColorFor(color)
- haloWidth = 0.dp
- textStyle = MaterialTheme.typography.labelSmall
+ measures = ScaleBarDefaults.measures()
+ color = Color.Black.copy(alpha = 0.75f)
+ haloColor = Color.White.copy(alpha = 0.75f)
+ haloWidth = 1.dp
+ textStyle = TextStyle(fontSize = 11.sp, color = ContentColor)
```

Plus the alchemist to spatial-k units port.

### CompassButton

Reimplement M3 ElevatedButton on Foundation:

```patch
- ElevatedButton(colors, shape, contentPadding) { Image(needle) }
+ Box(
+   .shadow(hovered ? hoveredShadowElevation : shadowElevation, shape)
+   .background(containerColor, shape)
+   .clip(shape)
+   .clickable(...) { animate home }
+ ) { Image(needle) }
```

plus the api change, accepting a `style: CompassButtonStyle` instead of M3 `ButtonColors`.

### DisappearingCompassButton

the above, plus bug fixes:

- rememberUpdatedState for getHomePosition / 
- derivedStateOf keyed on cameraState
- no longer passes modifier twice into the inner button

### Attribution

reimplemented Surface using Box:

```patch
- Surface(color, contentColor, tonalElevation, shadowElevation, shape, border)
+ Box(
+   .shadow(shadowElevation, shape)
+   .border?(border, shape)
+   .background(containerColor, shape)
+   .clip(shape)
+   .pointerInput(Unit) {}   // blocks map pan through the control
+   .semantics { isTraversalGroup = true }
+ )
```

and style:

```patch
- AttributionButtonStyle(+ tonalElevation)
+ AttributionStyle(textStyle instead of tonalElevation; tonal resolved by M3 wrapper)

- IconButton + Icon(vector)
+ Box + Image(painter, tint)

- Text(...)
+ BasicText(..., style = textStyle)

- linkStyles: primary + underline
+ base default: underline only (M3 wrapper still uses primary)

- expandedContent: (List<String>) -> Unit
+ expandedContent: (List<String>, TextStyle) -> Unit
```

lib/maplibre-compose-material3/src/commonMain/kotlin/org/maplibre/compose/material3/util/measures.kt
sargunv: moved to ScaleBarDefaults.measures()

sargunv:
@cursoragent add the try/catch recommended [here](https://github.com/maplibre/maplibre-compose/pull/886#discussion_r3758375996) and the xml comment recommended [here](https://github.com/maplibre/maplibre-compose/pull/886#discussion_r3757586982) (it comes from `https://github.com/maplibre/maplibre-native/blob/main/platform/android/MapLibreAndroid/src/main/res/drawable/maplibre_logo_icon.xml`)

lib/maplibre-compose/src/commonMain/composeResources/drawable/maplibre_logo.xml
sargunv: added in [27ea3d4](https://github.com/maplibre/maplibre-compose/pull/886/commits/27ea3d4333d0cfb1c01f6de77916f0af66a245d5)

lib/maplibre-compose/src/commonMain/kotlin/org/maplibre/compose/overlay/MaplibreLogo.kt
sargunv: added in [27ea3d4](https://github.com/maplibre/maplibre-compose/pull/886/commits/27ea3d4333d0cfb1c01f6de77916f0af66a245d5)

PR #1022, opened 2026-08-21
Carve the location engine into an independent location module

### LINKED ISSUE #1000: separate location module

sargunv:
## Description

Resolves #1000. Move the location and orientation providers, `LocationState`, and `LocationTrackingEffect` into a new `org.maplibre.compose:location` artifact that works without a map, cut the desktop backends over from `ComposeMapHost` to an `XdgPortalWindow` seam, and rename the backend modules to `location-runtime-linux|macos|windows|gms`.

## Validation

`test:desktop`, `test:android` (host), `test:js`, and `test:ios` pass on macOS.

## AI assistance

Written by Claude Code (Fable 5)


lib/location/src/androidMain/kotlin/org/maplibre/compose/location/AndroidOrientationProvider.kt
westnordost, line 1: https://github.com/maplibre/maplibre-compose/blob/501a8ec5953d196b1fa1842c4dd41764454de881/lib/location/src/androidMain/kotlin/org/maplibre/compose/location/AndroidOrientationProvider.kt#L73-L74

According to the [Android documentation on the `values` array in the sensor changed event for TYPE_ROTATION_VECTOR](https://developer.android.com/reference/android/hardware/SensorEvent#sensor.type_rotation_vector:), it **is** available:

> - values[4]: estimated heading Accuracy (in radians) (-1 if unavailable)
> 
> values[3], originally optional, will always be present from SDK Level 18 onwards. values[4] is a new value that has been added in SDK Level 18.

lib/location/build.gradle.kts
```diff
@@ -0,0 +1,69 @@
...
+  id(libs.plugins.kotlin.composeCompiler.get().pluginId)
+  id(libs.plugins.android.library.get().pluginId)
+  id(libs.plugins.compose.get().pluginId)
+  id(libs.plugins.mavenPublish.get().pluginId)
+}
+
+mavenPublishing {
+  pom {
+    name = "MapLibre Compose Location"
+    description = "Multiplatform location and orientation providers for Compose apps."
```
westnordost, line 14: Ah, right, it's dependent on compose because of the `remember*` functions, but otherwise there is no dependency on compose, I think.

So, if `maplibre-compose` depends on maplibre-location, then, couldn't the `remember*` functions remain in `maplibre-compose` instead? (or alternatively move to a `maplibre-location-compose` if one really wants it that fine-grained...)

It seems a little bit cleaner to me because location updates/permissions/orientation don't cry "UI framework" to me (but primarily: "KMP")

sargunv: Can be done, but you lose the common constructor in the separate module as Android needs a Context passed in, the other platforms don't need anything

Still, probably better that way, will do

westnordost:
There is one more thing that I think might be included, too, because it is a common use case.

On Android the developer is supposed to do a little [dance](https://developer.android.com/training/permissions/requesting), as alluded to already by the kdoc in the Android-specific part:

1. First request for permission as normal
2. Second request for permission when it was denied requires showing some rationale (`ActivityCompat.shouldShowRequestPermissionRationale` = `true`) (in a dialog etc) before actually making the request
3. Further requests may fail silently. Here, `ActivityCompat.shouldShowRequestPermissionRationale` will return false again (according to chat with gemini, didn't test!)

On iOS, I read that the first denied permission might be final. Further requests will be silently(?) ignored. So, afterwards, the app could send the user to the system settings for that app where he could enable the permission from there instead after a confirmation dialog, which might be a common pattern in iOS.

Now, maybe that part of the flow could be integrated. Of course, in both cases, it would require to open a dialog with the (user-provided) rationale, which would already be dependent on a compose Material2/Material3 dependency. Hm. Do you see a way to provide more for library users to make that dance simpler to implement without adding a dependency?

---

Thinking about it, I think the following things could be done:

1. Given the above logic, it might be possible on Android to reliably detect when to set `LocationPermission.NotGranted(canRequest = false)`. Currently, it isn't.:
  When 1. the permission location has been requested and denied at least once before and 2. `ActivityCompat.shouldShowRequestPermissionRationale` returns `false` now. This information would have to be stored persistently and cleared when the location permission is granted.

2. henceforth, the class could be extended to `LocationPermission.NotGranted(requiresShowRationale: Boolean, canRequest: Boolean)`. On iOS `requiresShowRationale` would just always be false.)

3. Then, when `canRequest=false`, the library user might want to send the user to the _system settings → apps → {this app} → permissions_, if possible. This module could offer an `expect/actual` implementation for that

westnordost:
Ah, another thing, looking at my [`LocationRequestFragment`](https://github.com/streetcomplete/StreetComplete/blob/e365234ad8705a3888488751da631af36b57679b/app/src/androidMain/kotlin/de/westnordost/streetcomplete/util/location/LocationRequestFragment.kt), this is also about a common use case which is probably present in the demo app as well:

I have a location button on the map with the following states

[Screencast_20260821_151924.webm](https://github.com/user-attachments/assets/8154706e-8e84-4560-a17e-8e5f0ab5f8a4)

From left to right: 
- location permission denied
- **location permission granted but location services are off**
- location services are on (app didn't request updates yet)
- waiting for first fix
- receiving location updates

Now, what is done when the user taps on the button when permission have been denied? Doing the request-permission dance, of course.

But what is done when the user taps on the button when permission has been granted **but location services are off?**

Either, a toast or similar could just inform the user that the location services are off. What my app does, though, and what others might want to do, too, is to send the user to the system location settings (after confirmation) so that the user can turn the location services on.

**To cut to the chase, with everything else in place, provide an `actual/expect` implementation to send the user to the system location settings.**

On Android, this would be the intent `Intent(Settings.ACTION_LOCATION_SOURCE_SETTINGS)`.

lib/location/build.gradle.kts
westnordost: constructor of which class?

sargunv: the rememberDefaultLocationProvider 

There's no non-compose commonMain equivalent because android needs to provide the Context. That's fine I think, you just do each construction in platform source sets and use the common interface everywhere else

sargunv: Might do this in a follow up though as I want to consider how it interacts with #1023

westnordost: Right, yeah, that's to be expected. In a non-compose-only project, these would then be provided by dependency injection instead.

sargunv:
> This information would have to be stored persistently and cleared when the location permission is granted.

This part makes me slightly wary, feels like a bit too much magic for a library, though I agree that'd be a valuable functionality.

I did have something like that in a draft of a previous PR, and removed it (after your comment) because the value would be wrong if the caller had already requested via some other mechanism (not this library). But that was in compose code, so it was possible to bypass the persistence even within this library, and if we move it to the compose-independent class I think it's okay to assume the caller didn't bypass the library entirely. 

So, I think yes but will think on it and probably defer to another follow up PR to review independently

westnordost:
Regarding storing something persistently, I may be mistaken, but I think this is exactly what `ActivityCompat.shouldShowRequestPermissionRationale` is doing.

> because the value would be wrong if the caller had already requested via some other mechanism (not this library)

Hmm. maybe this logic would work:
`canRequest = currentPermission_has_been_denied && ActivityCompat.shouldShowRequestPermissionRationale == false`.

In any case, storing a value persistently seems to be not necessary then. (If it is possible to query whether the last permission request has been denied)

sargunv:
follow ups from review comments:
- #1029 
- #1030 
- #1031 
- #1032 

each in draft mode until I self-review the ai output

sargunv:
all follow ups out of draft now

PR #1023, opened 2026-08-21
Discover the gms location as default android backend when available

sargunv:
## Description

- Added an `AndroidLocationBackend` SPI mirroring the desktop backend SPI with the framework providers as the fallback
- Register the fused providers from `location-runtime-gms` as such a backend so adding that dependency upgrades default providers on devices where Google Play services is available when the dependency is declared

## Validation

`test:android` (host) passes on macOS, including new resolver and ServiceLoader-registration tests.

## AI assistance

Written by Claude Code (Fable 5)

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -139,9 +139,9 @@ rendering interactive maps across Android, iOS, Desktop, and Web platforms.
   - `maplibre-compose`: Main map composables and core functionality
   - `maplibre-compose-material3`: Material 3 themed UI components
   - `location`: Location and orientation providers, usable without a map
-  - `location-runtime-gms`: Google Play Services location providers
-  - `location-runtime-linux|macos|windows`: Desktop location backends, loaded
-    through `ServiceLoader`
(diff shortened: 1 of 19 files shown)


westnordost:
> …on devices where Google Play services is available.

…and the dependency is declared, I presume, right?

sargunv:
Yeah, clarified the PR description

sargunv:
There's one footgun though, it'll currently return Misconfigured if multiple are registered and available (except the platform fallback)

So, on an app that bundles multiple providers for Android, if the device supports multiple backends, rather than picking a reasonable default we return Misconfigured 

This probably affects Huawei devices before the ban, if they had both GMs and HMS, and current Huawei devices if the user uses MicroG

I think we can add some priority property and tie breaking rule to avoid Misconfigured always.

westnordost:
I agree. The fused location provider should also not return `Misconfigured` when location is available from both the GNSS sensor and location by wifi, but just choose the better one. ;-)

PR #1029, opened 2026-08-21
Report the permission request path when location is not granted

sargunv:
## Description

See https://github.com/maplibre/maplibre-compose/pull/1022#issuecomment-5370183204

`LocationPermission.NotGranted` gains `shouldShowRationale`, and Android sets `canRequest` by persisting a denial record that clears whenever permission is granted or the rationale check returns true.

## Validation

Android host tests cover the denial tracker's truth table, and the Android, iOS, JS, desktop, and macOS-runtime targets build and pass their existing tests.

## AI assistance

Claude Fable 5 via Claude Code.

### DIFF
--- a/lib/location/src/androidMain/kotlin/org/maplibre/compose/location/AndroidLocationPermissionRequester.kt
+++ b/lib/location/src/androidMain/kotlin/org/maplibre/compose/location/AndroidLocationPermissionRequester.kt
@@ -29,19 +29,10 @@ import kotlinx.coroutines.flow.StateFlow
  */
 public class AndroidLocationPermissionRequester(private val context: Context) {
 
-  private val preferences =
-    context.applicationContext.getSharedPreferences(
-      "org.maplibre.compose.location",
-      Context.MODE_PRIVATE,
-    )
-
-  private var permanentlyDenied: Boolean
-    get() = preferences.getBoolean(PERMANENTLY_DENIED_KEY, false)
-    set(value) {
-      if (value != permanentlyDenied) {
-        preferences.edit().putBoolean(PERMANENTLY_DENIED_KEY, value).apply()
-      }
-    }
+  // In-memory only: Android can silently make the permission requestable again (auto-reset,
+  // revocation in settings), and no API distinguishes that from a permanent denial, so a
+  // persisted record could go stale forever.
+  private var permanentlyDenied = false
 
   private val mutableStatus = MutableStateFlow(readStatus())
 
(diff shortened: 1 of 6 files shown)


lib/location/src/androidMain/kotlin/org/maplibre/compose/location/AndroidLocationPermissionRequester.kt
```diff
@@ -67,8 +76,11 @@ public class AndroidLocationPermissionRequester(private val context: Context) {
       activity.activityResultRegistry.register(
         "org.maplibre.compose.location.permission.${nextKey.getAndIncrement()}",
         ActivityResultContracts.RequestMultiplePermissions(),
-      ) {
+      ) { results ->
         requestPending = false
+        // An empty result map is a cancelled request, not a user denial.
+        val denied = results.isNotEmpty() && results.values.none { it }
```
westnordost: `results.values.none { it }` ?

I don't follow. Shouldn't it be something like

```
val denied = results.values.any { it == PackageManager.PERMISSION_DENIED }
``` 	

?

westnordost (commented):
The AndroidPermissionDenialTracker with its PErmantentDenialStore  and SharedPreferencesDenial store look a bit needlessly unwiedly/abstract, IMO. It could be just one class, or not a class and all and be done in AndroidLocationPermissionREquester itself.

Also, when permission is finally granted, the store needs to be reset. Did I miss this somewhere?

sargunv:
while it's in draft mode, it's still ai slop (I usually self-review after pushing pr, before kicking out of draft)

westnordost:
ok, sorry!

sargunv:
no worries! you're welcome to review even in draft; just want to set the expectation of what you'll see at that point

sargunv:
> Also, when permission is finally granted, the store needs to be reset. Did I miss this somewhere?

it's there but buried as a side effect in the unnecessary abstraction

cleaning it up

lib/location/src/androidMain/kotlin/org/maplibre/compose/location/AndroidLocationPermissionRequester.kt
sargunv: result is  `Map<String, Boolean>`, permission name -> granted

so it was essentially "denied if there are results and none were granted", now more explicit

sargunv:
I think due to [this finding](https://github.com/maplibre/maplibre-compose/pull/1029#discussion_r3834306206) it's probably best to keep the flag in memory only rather than persistent

PR #1068, opened 2026-08-24
Add Windows WGL-to-D3D12 map bridge

sargunv:
## Description

Adds an opt-in Windows WGL-to-D3D12 map bridge, publishes x64 and arm64 OpenGL runtimes, and adds Mesa-backed CI coverage while retaining Vulkan as the default.

## Validation

Validated the Windows ARM64 OpenGL and Vulkan desktop suites, pinned-Mesa interop tests, runtime POM generation, documentation site build and link checks, and all locally runnable static checks.

## AI assistance

OpenAI Codex desktop agent using GPT-5 implemented and validated the change under human direction.


sargunv:
turns out performance is even worse than the vulkan map, not gonna merge unless a use case appears

angle might be interesting as it's backed by d3d

otherwise, should dig into why the vulkan map performs so poorly on windows.

PR #1098, opened 2026-08-26
add script-specific translations

westnordost:
Left in place `sr` and `zh` as fallbacks when script is unspecified


sargunv:
@cursoragent check the CI failures. my read is it's the same naming issue called out by the review bots. if you agree, fix it.

PR #1145, opened 2026-08-29
Redesign location measurement API

### LINKED ISSUE #1136: org.maplibre.compose.location.Location not actually @Serializable
### What happened?

`org.maplibre.compose.location.Location` is marked with @kotlinx.serialization.Serializable but actually, it isn't serializable because the field `timestamp: kotlin.time.TimeMark` has no serializer.

Error thrown when trying to serialize a `Location`:

> kotlinx.serialization.SerializationException: Serializer for subclass 'ValueTimeMark' is not found in the polymorphic scope of 'TimeMark'.
> Check if class with serial name 'ValueTimeMark' exists and serializer is registered in a corresponding SerializersModule.
> To be registered automatically, class 'ValueTimeMark' has to be '@ Serializable', and the base class 'TimeMark' has to be sealed and '@ Serializable'.

### Library Version

0.15.0

### Sample Code

```kt
Json.encodeToString(
  Location(
    position = PositionWithAccuracy(
      value = Position(longitude = -75.0, latitude = 45.0),
      accuracy = null
    ),
    timestamp = TimeSource.Monotonic.markNow()
  )
)
``̀`

### PR DESCRIPTION
**Pull request by sargunv**
## Description

- Replaced the nested Location/PositionWithAccuracy/BearingWithAccuracy model with a flat, serializable `LocationReading`
- Renamed "fix" to "reading" throughout
- Renamed "orientation" to "heading" throughout (orientation = 3d, including attitude. heading = 2d)
- moved the TimeMark to LocationEvent; left a serializable Instant in LocationReading
- changed OrientationProvider’s StateFlow into a cold Flow
- Redesigned `LocationPuck` with state-based and value-based entrypoints
- Added some tests, including serialization tests (resolve #1136)

### CODE REVIEW

File: lib/location/src/commonMain/kotlin/org/maplibre/compose/location/LocationProvider.kt [hunk 1]
```diff
@@ -144,8 +147,16 @@ public enum class LocationAccuracy {
...
-  public data class Fix(val location: Location) : LocationEvent
+  /**
+   * A location reading delivered by the provider.
+   *
+   * @property reading The delivered location reading.
+   * @property measurementMark Process-local monotonic mark for when [reading] was measured.
+   */
+  public data class Update(
+    val reading: LocationReading,
+    val measurementMark: TimeMark = reading.measurementMark(),
```
westnordost, line 158: I find it a bit odd that `TimeMark` is constructed from 
```
TimeSource.Monotonic.markNow() - (Clock.System.now() - measuredAt).coerceAtLeast(Duration.ZERO)
```

so, why not put `Clock.System.now()` there directly?

Anyway, it would probably be helpful to document what would be the difference between `LocationEvent.Update.measurementMark` and `LocationReading.measuredAt` exactly (if there should be any). If there is no (use case for a) difference, might leave out this wrapper.

lib/location/src/commonMain/kotlin/org/maplibre/compose/location/LocationProvider.kt
```diff
@@ -144,8 +147,16 @@ public enum class LocationAccuracy {
...
-  /** A measured location. */
-  public data class Fix(val location: Location) : LocationEvent
+  /**
+   * A location reading delivered by the provider.
+   *
+   * @property reading The delivered location reading.
+   * @property measurementMark Process-local monotonic mark for when [reading] was measured.
+   */
+  public data class Update(
+    val reading: LocationReading,
```
westnordost, line 157: `LocationReading` sounds a bit odd in my ears. (Association with a "reading", e.g. of a book.) But you're the native speaker.

lib/location/src/commonMain/kotlin/org/maplibre/compose/location/Heading.kt
```diff
@@ -0,0 +1,20 @@
...
+
+/**
+ * One measured horizontal direction that a device faces.
+ *
+ * @property bearing Absolute direction from north.
+ * @property accuracy Estimated bearing error, or `null` when unknown.
+ * @property measuredAt Wall-clock instant when the heading was measured.
+ */
+@Serializable
+public data class Heading(
```
westnordost, line 16: Slight inconsistency with `LocationReading`: This is also a "reading", just like `LocationReading`, no?

lib/location/src/commonMain/kotlin/org/maplibre/compose/location/LocationProvider.kt
sargunv: was LocationFix better? how about LocationMeasurement?

sargunv: to me, "location fix" reads like gps/gnss jargon, but maybe that's fine here. I changed it to try to be friendler to non-native speakers but "reading" is just as bad I guess? maybe `LocationMeasurement` is the way to go

lib/location/src/commonMain/kotlin/org/maplibre/compose/location/LocationProvider.kt
sargunv: the TimeMark should be used for elapsed time calculations, because it's unaffected by system clock corrections

the Instant is there purely to provide a timestamp that can be serialized, and then this helper provides a default `measurementMark` when constructing an event from an existing reading, because some location providers (browser, windows, linux) don't actually provide a monotonic mark

but I think it's clearer to just take the delivery time mark and don't try to reconstruct from a non-monotonic measuredAt, and remove the default helper.

lib/location/src/commonMain/kotlin/org/maplibre/compose/location/LocationProvider.kt
westnordost: An LLM chat (Qwen 3.7-Plus) also thinks that "Reading" would be best (in a general context of reading data off a sensor or instrument), and I trust that LLMs are pretty good at the task of "finding the most common/fitting word for something"

> **Most Common/General:**
> - **Reading** - The most common and natural term (e.g., "The thermometer reading was 72°F")
> - **Measurement** - Emphasizes the act of measuring
> - **Value** - Simple and widely used in technical contexts
>
> […]
> 
> **My recommendation:** **"Reading"** is probably the best general-purpose word. It's clear, commonly understood, and specifically implies something measured or detected by an instrument.

lib/location/src/commonMain/kotlin/org/maplibre/compose/location/LocationProvider.kt
westnordost: (`LocationEvent` is not serializable, though, just in case you assumed that right now. Didn't follow you completely.)

lib/location/src/commonMain/kotlin/org/maplibre/compose/location/LocationProvider.kt
sargunv: I had the same chat, that's how I picked "Reading" 😅 

when I asked the LLM again but to optimize for minimal confusion for non-native speakers, it went for "Measurement"

lib/location/src/commonMain/kotlin/org/maplibre/compose/location/LocationProvider.kt
sargunv: - LocationReading (now, `LocationMeasurement`) is serializable, and has the `measuredAt: Instant`
- `LocationEvent` is not serializable and has the `measurementMark: TimeMark`

PR #1166, opened 2026-08-30
continue refining location api

sargunv:
## Description

see comments on https://github.com/maplibre/maplibre-compose/pull/1145

- Rename the public location and heading records to `LocationMeasurement` and `HeadingMeasurement`. Considered:
  - "location fix" bit jargon-y, and doesn't work for heading
  - "location/heading reading" confused with "a reading of a book"
  - "location/heading observation" seems fine, maybe a bit scientific, next-best after "measurement"
- require every location provider to supply its monotonic measurement mark explicitly
- remove the helper that estimated a monotonic measurement from a non-monotonic Instant

## Validation

Validated with `mise run check`, `mise run test:desktop`, `mise run test:android`, `caffeinate -dimsu mise run test:js`, and `mise run test:ios`.

## AI assistance

OpenAI Codex implemented and validated the change under user direction.


westnordost:
> "location/heading reading" confused with "reading a book"

Actually, I meant "reading" as in **the** reading of a book. I.e. the event, where you go to, and the author of a book reads some sections of his book, and then you get to ask him questions and can buy the book, get an autograph. That's called a reading in English, right?

sargunv:
> That's called a reading in English, right?

yup, that's a reading too. A reading can also be a measurement from something, or what something displays, or a noun for something assigned to be read (school homework, "assigned reading"), or an interpretation ("my reading of this rule"), it's very overloaded lol

https://www.merriam-webster.com/dictionary/reading

- "data indicated by an instrument" was the intended usage here
- "a particular performance of something" is the one you're thinking of

PR #1168, opened 2026-08-30
Separate location provider setup from tracking failures

sargunv:
## Description

`LocationState.availability` now reports the provider's stable setup (`Available`, `Unsupported`, or `Misconfigured`), `permission` continues to report user authorization, and `status` reports only the active request because failures such as disabled system location services are discovered when tracking starts and appear as `Unavailable(ServicesDisabled)`.

## Validation

Validated on macOS with `mise run fix`, `mise run check`, desktop and Android tests, and the full documentation build.

## AI assistance

Implemented with OpenAI Codex using GPT-5 in the Codex desktop app.


westnordost:
I understand you've looked at whether it is possible to include something that'd amount to something similar to  `locationServicesAvailability: StateFlow<Boolean>` in `LocationProvider` but found that it is technically not possible because on some platforms, this information is actually not available before one has started requesting for a location.

I've looked at this PR but cannot really comment on it, because I haven't closely looked at or used the `LocationState` interface yet.

sargunv:
> found that it is technically not possible because on some platforms, this information is actually not available before one has started requesting for a location

not quite "technically not possible" but rather "technically not straightforward". It's probably possible with some consideration

PR #1291, opened 2026-09-05
Update native FFI to 0.202609.0

### LINKED ISSUE #956: Report visible-region corners that survive a repeated world
Read from `main` at `0238fe57`.

`queryVisibleRegion` returns four corners, and mbgl wraps unprojected longitudes into ±180°. That
describes a pane narrower than the world exactly. Once the pane is wider — at zoom 1 the world is
1024 dp across, so any window wider than that — both edges land on the same stretch of longitude,
and the four numbers describe the overlap rather than the extent. Nothing in them says which of the
two you have: the camera still falls inside the reported range, and the values are as plausible as
any others.

Anything sized from those bounds then covers a fraction of the map, and the result reads as a
rendering fault rather than a measurement one. It cost me a bug report against this library that was
wrong — MapLibre draws a whole-world overlay correctly, and the bounds I had built it from were the
problem. Once I measured the pane's own span, with longitudes running past ±180°, the content drew
correctly however many copies of the world were on screen.

#904 taught the FFI `getVisibleBoundingBox` to unwrap its corners around the pane centre, so that
box may extend past ±180° and the antimeridian case is right; #934 moved that work into the viewport
snapshot, where a comment states the rule — mbgl wraps unprojected longitudes to ±180, and like GL
JS the box may then extend past it (`MlnFfiMapSession.kt:852-868`). Two gaps remain, and both are
visible in that same function. `mirroredVisibleRegion` is built from the raw corners, so
`getVisibleRegion` still hands back wrapped ones (`:840-846`, `:1042`). And unwrapping around a
centre cannot express a span wider than 360° in any case, so a pane wider than the world still
under-reports through either call.

What would help is an answer that keeps its continuity: corners unwrapped relative to the camera, as
the bounding box now does, plus a longitude span or some other way to say "wider than the world".
Failing that, a line in the KDoc of both methods saying that the corners are wrapped and what that
means once the world repeats. Neither KDoc mentions it today — they say only that the region is a
four-sided polygon — so every caller sizing content to the viewport has to rediscover this.

### LINKED ISSUE #1289: A process that has used map runtimes can hang forever at exit: mln::Log::~Log() joins a thread during static destruction
A JVM that has created and closed map runtimes sometimes never exits. The VM thread blocks in C++
static destruction and the process has to be killed:

```
VM_Exit::doit → os::exit → exit → __cxa_finalize_ranges
  → mln::Log::~Log()                    (libmaplibre-native-c.dylib)
    → mln::ThreadedScheduler::~ThreadedScheduler()
      → std::thread::join() → __ulock_wait
```

A blocking `join()` reached through `__cxa_finalize_ranges` cannot be made safe from the outside:
the scheduler's thread may no longer be able to finish, and the wait cannot be interrupted.

**I have no deterministic reproducer, which is why this is a heads-up rather than a bug with a
recipe.** It is a race at shutdown: the same case exits cleanly most times and wedges occasionally.
Six wedges over three days, every stack byte-identical, in two unrelated test classes — one that
builds about a dozen runtimes with snapshotters and captures, one that takes nine captures through a
single runtime. Small single cases (a runtime created and closed, a snapshotter with no capture, one
capture, 3/5/8/11 runtimes with captures) all exit cleanly on repeat.

**Environment.** maplibre-compose `0.15.1-20260903.101931-9` through `0.15.1-20260905.101546-11`,
`maplibre-native-ffi 0.202608.3`, macOS on Apple silicon, Metal, JetBrains Runtime 25.0.2, headless
under Gradle. No windowing toolkit is involved: the process has no window and no event loop, and the
sample shows no frame outside `libjvm`, `libsystem_*`, `libmaplibre-native-c` and `libc++`.

**What it costs.** In an interactive app, little — the user closes it. In anything expected to
terminate (a test suite, a CI job, a snapshot or tile-rendering batch) the hang is indistinguishable
from an infinite loop, and the usual diagnosis does not work: the VM thread is the blocked one, so
the JVM stops answering the attach protocol and `jstack` reports *"target process doesn't respond"*.
The stack above came from macOS `sample`, which needs no cooperation from the VM. A hung run also
loses whatever output it had already produced.

Happy to run anything that would help pin it down.

sargunv:
## Description

Update maplibre-native-ffi from `0.202608.3` to `0.202609.0`, including the JVM logging shutdown fix.

- Move feature-state operations to the native map and remove Compose's mirror and renderer replay.
- Migrate Vulkan non-dispatchable handles to `VulkanHandle`.
- Use unwrapped unprojection so native viewport bounds and screen conversion preserve repeated worlds, including viewports wider than 360°. Returned longitudes may now exceed ±180°.
- Replace resource-cancellation polling with native callbacks
- Expose native source storage policy through `SourceHandle.isVolatile`, guarded by the existing handle lifetime.

Android ARMv7 and superseding stalled style loads will be separate stacked PRs. musl Linux is out of scope for now.

Fixes #1289
Fixes #956

## Validation

Initial upgrade validation passed:

- Static checks and style-spec parity.
- Desktop: 712 passed; four Linux-specific interop tests skipped on macOS ARM64/Metal.
- iOS simulator: 575 tests.
- Real headless Chrome: 451 tests.
- Android host: 299 tests.
- Android 16 ARM64 emulator: 645 tests across modules.

Shutdown comparison on macOS ARM64/JDK 25.0.4 swapped only the published binding and Metal runtime jars. Both old-release cases reached the exit marker and hung beyond 12 seconds; stack samples show the logger scheduler join/JVM detach deadlock. Both new-release cases exited with code 0 in about one second, with the callback installed or cleared. Initial stress verification also passed six child JVMs, exercising 18 Compose runtime lifecycles and 36 snapshots. The permanent regression now uses one child, one runtime, and one snapshot, observing the warning through Compose's real log bridge. native-ffi already covers the raw installed/cleared callback matrix.

Linux/Windows Vulkan execution, Android Vulkan/x64 Goldfish, physical devices, and ARMv7 were not validated locally. Remote CI is reported by the PR checks.

## AI assistance

Implemented and validated with Codex (GPT-6), including the upstream audit and regression tests.


lib/maplibre-compose/src/commonMain/kotlin/org/maplibre/compose/camera/Viewport.kt
```diff
@@ -23,7 +23,8 @@ internal constructor(
    * The smallest bounding box that contains the currently visible area.
    *
    * This north-aligned rectangle can include areas outside [visibleRegion] when the map is rotated
-   * or tilted.
+   * or tilted. Longitudes preserve world copies, so the box may extend past ±180° or span more than
+   * 360°.
    */
   public val visibleBoundingBox: BoundingBox,
```
sargunv, line 29: should we offer a wrapped and unwrapped version? or alternatively BoundingBox.wrapped extension?

lib/maplibre-compose/src/commonMain/kotlin/org/maplibre/compose/map/MapRuntime.kt
```diff
@@ -926,7 +926,11 @@ internal constructor(
     it.screenLocationFromPosition(position)
   }
 
-  /** Unprojects [offset] into a geographic position, or returns null without a viewport. */
+  /**
+   * Unprojects [offset] into a geographic position, or returns null without a viewport.
+   *
+   * Longitude preserves the visible world copy and may extend past ±180°.
+   */
   public fun positionFromScreenLocation(offset: DpOffset): Position? = withAttachmentRead {
```
sargunv, line 934: ditto for Position, maybe a param for wrapping, or Position.wrapped extension?

lib/maplibre-compose/src/commonMain/kotlin/org/maplibre/compose/util/VisibleRegion.kt
```diff
@@ -5,6 +5,9 @@ import org.maplibre.spatialk.geojson.Position
 /**
  * Four-sided polygon representing the visible area of the map composable. If the camera has tilt
  * (pitch), this polygon is a trapezoid instead of a rectangle.
+ *
+ * Longitudes preserve the visible world copy and may extend past ±180°. A viewport wider than one
+ * world can span more than 360° of longitude.
  */
 public data class VisibleRegion(
```
sargunv, line 12: ditto

lib/maplibre-compose/src/maplibreNativeMain/kotlin/org/maplibre/compose/sources/SourceVolatility.kt
```diff
@@ -0,0 +1,20 @@
...
+ *
+ * Like other handle operations, access fails after the source is removed or its style is replaced.
+ */
+public var SourceHandle.isVolatile: Boolean
+  get() = operation {
+    checkNotNull((style as MlnFfiStyleBinding).readMap { it.styleSourceInfo(id)?.volatileSource })
+  }
+  set(value) {
+    operation { (style as MlnFfiStyleBinding).mutateMap { it.setStyleSourceVolatile(id, value) } }
+  }
```
sargunv, line 20: why an extension rather than a member? due to expect/actual, or existing patterns?

gradle/libs.versions.toml
```diff
@@ -46,7 +46,7 @@ lifecycleRuntimeCompose = "2.11.0"
 lwjgl = "3.4.2"
 materialKolor = "5.0.1"
 maplibre-js = "6.6.0"
-maplibre-nativeFfi = "0.202608.3"
+maplibre-nativeFfi = "0.202609.0"
```
sargunv, line 49: did we re-enable the tests previously disabled pending the goldfish mitigation?

PR #1304, opened 2026-09-06
test: remove redundant coverage and strengthen behavioral contracts

sargunv:
## Description

Remove redundant smoke tests and assertions that duplicate implementation details. Retain distinct lifecycle, cancellation, rendering, and binding contracts, and consolidate overlapping serialization and engine checks.

Strengthen the remaining tests with positive query controls, pixel checks for requested repaints, independent expected values, and explicit synchronization. Map fixtures now own their coroutine work and synchronize the native style binding before returning from a load. Unsupported environments report skips instead of passing silently.

This removes 101 test cases and 2,302 lines net across 73 files. All changes are in tests and fixtures.

## Validation

Passed locally:

- `mise run check`
- `mise run test:android`
- `mise run test:android:device` (two animation tests skipped because the emulator disables system animations)
- `mise run test:desktop` (macOS; platform-specific tests skipped where unsupported)
- `mise run test:js` (real Chrome maps)
- `mise run test:ios` (iOS simulator)
- `mise run ci:test-scripts` (53 tests)

Temporary mutable-list and numeric-literal defects caused the strengthened tests to fail. Separate task-wake and shutdown-wake mutations also failed the retained offline metadata lifecycle test. Production code was restored before the passing runs.

Linux and Windows native execution awaits full CI.

## AI assistance

OpenAI Codex (GPT-6) assisted with the test audit, implementation, validation, and PR description.


sargunv:
The parked-wake suggestion is not being restored: those tests inferred parking from a sleep, and the retained metadata lifecycle test fails both when the task wake signal is removed (operation timeout) and, in a separate mutation, when the shutdown wake signal is removed (close assertion); production code has been restored.

PR #1315, opened 2026-09-07
Replace the visible bounding box with VisibleBounds

sargunv:
## Description

MapLibre repeats the world horizontally, but viewport queries returned spatial-k `BoundingBox`, a GeoJSON type whose RFC 7946 convention (antimeridian crossing encoded as east < west) can't express what the map reports. See #956.

- New `VisibleBounds` type for viewport results: continuous longitudes that preserve world copies, `east >= west` enforced, spans past 360° representable. `toBoundingBox()` converts to an RFC 7946 `BoundingBox`.
- Breaking: `Viewport.visibleBoundingBox` → `visibleBounds`; `MapState.getVisibleBoundingBox()` → `getVisibleBounds()`.
- `screenLocationFromPosition` now projects the world copy nearest the camera on every platform; GL JS previously projected a longitude as given, so a position equivalent to a visible one modulo 360° could land offscreen.

Companion: maplibre/spatial-k#387 adds `wrapped()` and `splitAtAntimeridian()` to spatial-k.

## Validation

`mise run check`, `test:desktop`, `test:js`, `test:android` (host), `test:ios` (simulator), and `build:docs` pass. New live-map tests pin the documented contract on native and Chrome. Not run: Android device, Windows, Linux.

## AI assistance

Kimi K3


sargunv:
Re the zero-width-on-antimeridian review note: fixed in 693f0680a. On representation, for the record:

- Zero width is `west == east`: RFC 7946 §5.1 defines bbox edges by interpolation (`lon = west + (east − west)·t`), so equal endpoints *are* the degenerate strip; mbgl agrees (`LatLngBounds::singleton(a)` = `{a, a}`, `include/mln/util/geo.hpp`).
- Full world is `west=-180, east=180`: RFC 7946 §5.3 (pole boxes use `[-180, lat, 180, lat]`), and mbgl `LatLngBounds::world()` = `{{-90, -180}, {90, 180}}` (same file).
- The meridian sign differs per engine: mbgl `util::wrap(lon, -180, 180)` maps +180 → −180 (`include/mln/math/wrap.hpp`), while GL JS `wrap()` maps −180 → +180 (`src/util/util.ts`). We follow spatial-k`'s `Position.wrapped()` and mbgl: a zero-width bound on the antimeridian wraps to −180/−180.

Covered by `a_zero_width_bound_stays_zero_width` in `VisibleBoundsTest`.

Issue #958, opened 2026-08-18
Anchor layers by role, such as below the first symbol layer


mipastgt:
Read from `main` at `0238fe57`.

Overlays usually belong under the base map's labels. Every anchor that is not `Top` or `Bottom`
names a layer id: `Above`, `Below` and `Replace` are all data classes over a `layerId`
(`Anchor.kt:44`, `:50`, `:56`). Those ids differ per provider and per style, so an application that
supports more than one base style carries a table of ids and updates it whenever a provider renames
theirs.

The `MaplibreMap` KDoc names exactly this use — adding "a layer just below the first symbol layer"
from the base style (`MaplibreMap.kt:108`) — and the API can only express it by id.

An anchor that names the role instead would say what the application means and survive a style
change. It looks resolvable in the wrapper from the loaded style's layer list, without anything new
from the renderer.

sargunv:
It sounds like you want an anchor with a predicate, rather than an anchor based on name, something like:

```kt
Anchor.BelowFirst({it is SymbolLayer}) {
  SomeLayers()
}
```

mipastgt:
Yes, a predicate is what I mean, and `Anchor.BelowFirst { … }` reads exactly right.

One thing to be aware of before it is written that way: a `Layer`-typed predicate cannot match on
`main` at `470681d2`. Base-style layers never arrive as typed layers — `getLayers()` wraps every one
of them in `UnknownLayer` on all four targets (`AndroidStyle.kt:88`, `IosStyle.kt:73`,
`MlnFfiStyle.kt:127` via `reconstructLayer` at `:235`, `GlJsStyle.kt:174`) — so `{ it is SymbolLayer }`
would be false for every base layer everywhere. `Layer` is `internal` as well, and its `type` is
`protected` (`nextCommonMain/…/Layer.kt:21`, `:24`), so neither the class nor the property can appear
in a public predicate signature as things stand.

The information is there, though: on the next targets `UnknownLayer` already parses `type`, `source`
and the other root keys out of the reported definition JSON, and on Android/iOS the platform layer
behind `impl` carries its class. So the predicate wants a small public descriptor rather than a
`Layer` — id, type, source, source-layer — and then

    Anchor.BelowFirst { it.type == LayerType.Symbol } { SomeLayers() }

works on every target with no renderer support. For what it is worth, a plain `Anchor.BelowSymbols`
would cover my case and, I suspect, most of the ones that motivate the general form; a predicate as
well is strictly better.

[...]

PR #1332, opened 2026-09-07
Remove Anchor.Replace

sargunv:
## Description

`Anchor.Replace` existed because base-style layers could not be written from the composition. `LayerHandle` covers that now through `mapState.style.layers`, so this removes `Replace` and the reconciler paths that removed and restored base layers for it. The composition never modifies the base style; the docs point at `LayerHandle` instead. Part of #958.

## Validation

- `mise run check`: passed.
- `mise run test:desktop`: passed. Includes the edited `StyleCompositionOrderTest`, `MapPresentationTest`, and `MlnFfiStyleSwitchTest` (via `androidJvmTest`).
- `caffeinate mise run test:js`: passed, including the rewritten `BrowserStyleConformanceTest`.
- `mise run build:docs`: passed on the initial commit; the later doc edits only remove text and link to the `LayerHandle` API reference, verified by the docs CI job.
- Not run: Android host/device and iOS suites.

## AI assistance

Claude Fable 5.1 via Claude Code (workflow-driven implementation), pending author review.

### DIFF
--- a/demo-app/common/src/commonMain/kotlin/org/maplibre/compose/docsnippets/Composition.kt
+++ b/demo-app/common/src/commonMain/kotlin/org/maplibre/compose/docsnippets/Composition.kt
@@ -3,6 +3,8 @@
 package org.maplibre.compose.docsnippets
 
 import androidx.compose.runtime.Composable
+import androidx.compose.runtime.LaunchedEffect
+import kotlinx.serialization.json.JsonPrimitive
 import org.maplibre.compose.map.DefaultMapRuntime
 import org.maplibre.compose.map.MaplibreMap
 import org.maplibre.compose.map.rememberMapState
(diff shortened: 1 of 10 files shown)


demo-app/common/src/commonMain/kotlin/org/maplibre/compose/docsnippets/Composition.kt
```diff
@@ -21,4 +23,11 @@ fun Composition() {
...
   MaplibreMap(state = state)
   // #endregion base-plus-content
+
+  // #region base-layer-handle
+  val buildings = state.style.layers["building"]
+  LaunchedEffect(buildings) {
+    buildings?.setPaintProperty("fill-color", JsonPrimitive("#2196f3"))
+  }
+  // #endregion base-layer-handle
 }
```
sargunv: does not belong in composition docs; imperative api is already separately documented

docs/src/content/docs/composition.mdx
```diff
@@ -25,8 +25,10 @@ positions in the layer stack.
 
 The library manages the sources and layers in the style block. It
 does not modify the rest of the base style. To change a base style layer,
-use its handle in `mapState.style.layers`, or replace it with
-<Api of="Anchor.Replace" /> and declare its replacement yourself.
+look up its <Api of="LayerHandle" /> in `mapState.style.layers` and write the
+property through the handle, in an effect keyed on the handle:
+
+<Code code={region(composition, "base-layer-handle")} lang="kotlin" title="App.kt" />
```
sargunv: just link to imperative api without further details here

Issue #1303, opened 2026-09-06
explore compiler plugin for expressions in normal kotlin


sargunv:
not anytime soon, but would be really neat to someday explore something like https://github.com/Kronos-orm/kotlin-expr-tree-kcp so normal kotlin code could replace our expressions dsl

sargunv:
@cursoragent orchestrate subagents to prototype this; replace our entire expression DSL with a kotlin compiler plugin where users write their expressions with normal Kotlin syntax and control flow. Open a draft PR just for visibility; it's a throwaway PR not intended to merge, but it _should_ work and express the full expression api with idiomatic Kotlin and type safety. Keep committed markdown notes about what went well, what didn't go well, major decisions being made, and include sample expressions in the PR description showing usage.

PR #1335, opened 2026-09-07
Prototype: write MapLibre expressions as ordinary Kotlin

sargunv:
## Description

**Throwaway draft.** Not intended to merge. It explores replacing the expression DSL with a Kotlin compiler plugin so users write `if` / `when`, operators, and `feature` / `zoom` as normal Kotlin. The plugin lowers that to the existing `Expression` AST.

Inspired by [kotlin-expr-tree-kcp](https://github.com/Kronos-orm/kotlin-expr-tree-kcp). Design notes: `lib/maplibre-compose-expr-compiler/PROTOTYPE.md`.

The plugin now checks const-able captures at compile time and covers the remaining style-spec surface, including the cases that fail for engine or Kotlin reasons rather than missing visitor branches.

### Sample usage

Constants:

```kotlin
CircleLayer(
  id = "quakes-constant",
  source = earthquakes,
  radius = expr { 4.dp },
  color = expr { Color.Red },
)
```

Feature data with `when` / `if` (evaluated per feature, not at composition):

```kotlin
CircleLayer(
  id = "quakes-by-magnitude",
  source = earthquakes,
  radius = expr {
    val mag = feature.number("mag")
    when {
      mag >= 6 -> 16.dp
      mag >= 4 -> 8.dp
      else -> 4.dp
    }
  },
  color = expr {
    if (feature.number("mag") > 5) Color.Red else Color.Yellow
  },
)
```

Subject `when` becomes `match`. Comparison `when` stays `case`:

```kotlin
val color: Expression<ColorValue> = expr {
  when (feature.string("kind")) {
    "park" -> Color.Green
    "water" -> Color.Blue
    else -> Color.Gray
  }
}
```

Zoom interpolation, format spans, GeoJSON, and a composition-time helper:

```kotlin
fun themeRed() = Color.Red
val radius = expr { interpolate(exponential(2), zoom, 5 to 2.dp, 10 to 8.dp) }
val label = expr {
  format(span(feature.string("name"), font = listOf("Open Sans"), textColor = Color.Red))
}
val inside = expr { feature.within(Point(Position(0.0, 0.0))) }
val painted = expr { themeRed() }
```

### Compile-time rejects

These are compile errors, not runtime throws from `ExprEmit.lit`:

```kotlin
// type is not const-able
expr { mapState }

// the map cannot run Kotlin per feature
fun paint(mag: Double) = if (mag > 5) Color.Red else Color.Yellow
expr { paint(feature.number("mag")) }

// MapLibre has no array constructor for dynamic components
expr { Offset(feature.number("x").toFloat(), feature.number("y").toFloat()) }
```

`var`, loops, and `try` are also compile errors.

### Limitations the hard cases showed

- MapLibre cannot build `Offset`, `padding`, or `listOf` from feature numbers. `asOffset()` only asserts an existing two-number array.
- Text-unit offsets (`offset(12.sp, 4.em)`) are composition-time `TextUnit` values only.
- `image(bitmap/painter)` options (size, SDF, stretch) register a bitmap at composition time. A style-image *name* can be map-evaluated.
- Reified `asEnum<T>()` cannot be an interface member. The helper takes `List<String>` of style-spec names.
- `kotlin.math.PI` freezes as a number. Scope `pi` / `e` / `ln2` emit the map operators.
- `null` is typed `Nothing?` in IR; that has to be treated as `nil`.
- Matching helpers by package prefix is wrong: tests live in `org.maplibre.compose.expressions.kotlin`.
- Kotlin/JS does not treat a primitive `boolean` as `is Boolean`. Boolean IR constants call typed `litBoolean(Boolean)`. Interning `BooleanLiteral.True`/`False` can return `undefined` during companion init.
- Kotlin/JS compiles `is Int` / `is Float` / `is Double` to `typeof === 'number'`. `lit(2.5)` used to take the Int branch; `IntCache[2.5]` is `undefined`. One `Number` path now keeps whole values as Int and the rest as Float.
- Dokka `failOnWarning` rejects an unresolved `[EnumValue]` KDoc link on `ExprScope.asEnum` (and the inherited stub). The comment now names the type in prose.

The old `const` / `switch` / `gt` DSL remains as the lowering target and escape hatch. Layer parameters are still `Expression<T>`.

## Validation

Draft CI on `68b23cdd` failed three jobs:

- **docs** and **android / 36** (`test:publishing`): Dokka `failOnWarning` on `[EnumValue]` in `ExprScope.asEnum`. The KDoc link is gone.
- **js / chromium**: `TypeError: Cannot read properties of undefined (reading 'compile_…')` on `lit_encodes_scalars` and `constants_become_literals`. After the boolean intern fix, the remaining hole was `lit(2.5)` matching `is Int` on JS and indexing `IntCache` with a fractional key.

Ran on Linux after `b12ba853`:

- `:lib:maplibre-compose-expr-compiler:test` — 3/3
- `:lib:maplibre-compose:jvmTest` for `expressions.kotlin` (32) and `ExpressionJsonTest` (10) — all passed
- `:lib:maplibre-compose:jsBrowserTest` with system Chrome — **515/515 passed**, including the two previously failing tests (ExprEmitTest 6, KotlinExprPluginTest 26)

Not run locally: `dokkaGenerate` (no Android SDK here), iOS, Android device, desktop GPU. No `ci:full`: shared Kotlin plus a JVM plugin, no ABI or native loading change.

## AI assistance

Cursor Grok 4.6 (cloud agent). I designed the Map vs Compose classification and the const-able check, then implemented the remaining DSL lowering, CI fixes, and tests.

<div><a href="https://cursor.com/agents/bc-e85d346d-0a91-4df6-bc7c-6218caf62006?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-e85d346d-0a91-4df6-bc7c-6218caf62006&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>

### DIFF
--- a/buildSrc/src/main/kotlin/applyMapLibreExprPlugin.kt
+++ b/buildSrc/src/main/kotlin/applyMapLibreExprPlugin.kt
@@ -1,0 +1,30 @@
+import org.gradle.api.Project
+import org.gradle.api.tasks.bundling.Jar
+import org.gradle.kotlin.dsl.named
+import org.gradle.kotlin.dsl.withType
+import org.jetbrains.kotlin.gradle.tasks.AbstractKotlinCompile
+import org.jetbrains.kotlin.gradle.tasks.KotlinCompilationTask
+
+/**
+ * Applies the in-repo MapLibre `expr { }` compiler plugin to Kotlin compilations in this project.
+ *
+ * JVM, JS, Android, and metadata compilations use `pluginClasspath`. Native compilations get
+ * `-Xplugin` because they use a different task type.
+ */
+fun Project.applyMapLibreExprCompilerPlugin() {
+  val compilerPath = ":lib:maplibre-compose-expr-compiler"
+  val compiler = rootProject.findProject(compilerPath) ?: return
+  evaluationDependsOn(compilerPath)
+
+  val jar = compiler.tasks.named("jar", Jar::class)
+
+  tasks.withType<AbstractKotlinCompile<*>>().configureEach { pluginClasspath.from(jar) }
+
+  tasks.withType<KotlinCompilationTask<*>>().configureEach {
+    dependsOn(jar)
+    if (this !is AbstractKotlinCompile<*>) {
+      val pluginArg = jar.flatMap { it.archiveFile }.map { "-Xplugin=${it.asFile.absolutePath}" }
+      compilerOptions.freeCompilerArgs.add(pluginArg)
+    }
+  }
+}
--- a/demo-app/common/build.gradle.kts
+++ b/demo-app/common/build.gradle.kts
@@ -7,6 +7,8 @@ plugins {
   id(libs.plugins.compose.get().pluginId)
 }
 
+applyMapLibreExprCompilerPlugin()
+
 kotlin {
   jvmToolchain(libs.versions.java.toolchain.get().toInt())
 
--- a/demo-app/common/src/commonMain/kotlin/org/maplibre/compose/docsnippets/Expressions.kt
+++ b/demo-app/common/src/commonMain/kotlin/org/maplibre/compose/docsnippets/Expressions.kt
@@ -5,16 +5,7 @@ package org.maplibre.compose.docsnippets
 import androidx.compose.runtime.Composable
 import androidx.compose.ui.graphics.Color
 import androidx.compose.ui.unit.dp
-import org.maplibre.compose.expressions.dsl.asNumber
-import org.maplibre.compose.expressions.dsl.condition
-import org.maplibre.compose.expressions.dsl.const
-import org.maplibre.compose.expressions.dsl.exponential
-import org.maplibre.compose.expressions.dsl.feature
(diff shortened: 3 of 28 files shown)


sargunv:
Concept is proven successfully. Requires some deeper thought about what sorts of syntax are allowed vs rejected, how that's surfaced at compile time. I don't plan to work on that soon, so closing now, #1303 should track any future planning around this.

PR #1337, opened 2026-09-07
Anchor layers by predicate

### LINKED ISSUE #958: Anchor layers by role, such as below the first symbol layer

sargunv:
## Description

`Anchor.Above` and `Anchor.Below` now take a predicate over the base style's `LayerHandle`s: `Below` lands under the lowest match and `Above` over the highest, and with no match `Below` falls back to the top of the stack and `Above` to the bottom, so a missing layer ID no longer fails a revision. The reconciler resolves every anchor to a placement (top, bottom, or below one base layer) before it groups and diffs layers, so a fresh lambda on each recomposition never removes, re-adds, or moves a layer, and MapLibre Native's annotation layer is neither offered to predicates nor reported as a style layer. The `layerId` forms remain as sugar, and the transit and data demos anchor their layers below `symbol` layers.

Stacked on #1332. Closes #958.

## Validation

- `mise run check`: passed
- `mise run test:desktop`: 835 tests, 0 failures, 7 skipped (pre-existing skips)
- `caffeinate mise run test:js`: 532 tests, 0 failures
- `mise run build:docs`: built; the later KDoc and doc trims only remove text, verified by `mise run check`, a JVM compile, and the docs CI job
- The new `AnchorPlacementTest` (live, both engines) fails on native without the annotation-layer filter and fails on both engines without the layer-node key change; verified by reverting each temporarily.

## AI assistance

Claude Fable 5.1 via Claude Code (workflow-driven implementation), pending author review.

### DIFF
--- a/docs/src/content/docs/composition.mdx
+++ b/docs/src/content/docs/composition.mdx
@@ -53,8 +53,10 @@ current state, and let the runtime reconcile it.
 When you assign `mapState.style.baseStyle`, the map loads the new style and evaluates
 the style block against it. Each evaluation declares your
 sources and layers against the new style. They persist across the switch
-without extra code. An `Anchor` that names a base layer applies only when that
-layer exists in the new style.
+without extra code. An `Anchor` resolves again against the new style's layers,
+so a predicate that matches by layer type lands in the equivalent place in each
+style, and one that matches nothing falls back to the top or bottom of the
+stack.
 
 To share one definition between several maps, write the style content as a
 composable function and call that function from each style block. The same
@@ -74,4 +76,4 @@ read. Content that runs on both hosts branches on it.
 Layers and sources have string IDs because the underlying MapLibre style
 identifies them by ID. Within one map, each layer ID must be unique, including
 against the base style's layers. A feature query takes layer IDs to limit its
-scope, and an `Anchor` names a base style layer by its ID.
+scope, and an `Anchor` can select a base style layer by its ID.
--- a/docs/src/content/docs/composition.mdx
+++ b/docs/src/content/docs/composition.mdx
@@ -53,10 +53,7 @@ current state, and let the runtime reconcile it.
 When you assign `mapState.style.baseStyle`, the map loads the new style and evaluates
 the style block against it. Each evaluation declares your
 sources and layers against the new style. They persist across the switch
-without extra code. An `Anchor` resolves again against the new style's layers,
-so a predicate that matches by layer type lands in the equivalent place in each
-style, and one that matches nothing falls back to the top or bottom of the
-stack.
+without extra code. An `Anchor` is resolved against the new style's layers.
 
 To share one definition between several maps, write the style content as a
 composable function and call that function from each style block. The same
--- a/docs/src/content/docs/layers.mdx
+++ b/docs/src/content/docs/layers.mdx
@@ -59,10 +59,10 @@ after the property. The default, null, uses the global transition that
 By default, your layers draw above the base style layers. <Api of="Anchor" />
 inserts a layer at another position: at the `Bottom`, at the `Top`, or `Above`
 or `Below` base layers that a predicate selects. The predicate receives a
-<Api of="LayerHandle" /> for each base layer, so it can match on the layer's
-type, ID, or properties. `Below` lands under the lowest matching layer and
-`Above` lands over the highest one. When nothing matches, `Below` places the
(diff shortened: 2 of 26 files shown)


docs/src/content/docs/composition.mdx
```diff
@@ -56,8 +56,10 @@ current state, and let the runtime reconcile it.
 When you assign `mapState.style.baseStyle`, the map loads the new style and evaluates
 the style block against it. Each evaluation declares your
 sources and layers against the new style. They persist across the switch
-without extra code. An `Anchor` that names a base layer applies only when that
-layer exists in the new style.
+without extra code. An `Anchor` resolves again against the new style's layers,
+so a predicate that matches by layer type lands in the equivalent place in each
+style, and one that matches nothing falls back to the top or bottom of the
+stack.
```
sargunv: too much metaphor and flourish

docs/src/content/docs/layers.mdx
```diff
@@ -58,8 +58,16 @@ after the property. The default, null, uses the global transition that
...
+<Api of="LayerHandle" /> for a base layer, so it can match on the layer's type,
+ID, or properties. `Below` lands under the lowest matching layer and `Above`
+lands over the highest one. When nothing matches, `Below` places the layers at
+the top and `Above` places them at the bottom, without an error.
 
 <Code code={rememberedMapStyle(region(layers, "anchors"))} lang="kotlin" title="App.kt" />
 
+Only base style layers are candidates: a predicate never sees the layers that
+you declare, so anchor your own layers relative to each other by their order in
+the style block.
```
sargunv: ditto. see the repo's docs skill

lib/maplibre-compose/src/commonMain/kotlin/org/maplibre/compose/layers/Anchor.kt
```diff
@@ -10,79 +10,155 @@ import org.maplibre.compose.util.MaplibreComposable
...
- * **Note:** This mechanism can only be used to anchor layers at `layerId`s from the *base map
- * style* loaded by [MapStyleState.baseStyle][org.maplibre.compose.map.MapStyleState.baseStyle].
- * Anchoring layers defined in the composition to other layers defined in the composition is not
- * possible.
+ * Only the layers of the loaded base style are candidates. A layer declared in the style content is
+ * never matched, so an anchor cannot refer to layers of another anchor. Layers that resolve to the
+ * same position keep their order from the style content.
  *
- * See [Anchor.Companion] for [Composable] functions to use in the layers composition.
+ * See [Anchor.Companion] for the composable functions that apply an anchor to a block of layers.
```
sargunv, line 17: don't document every subtype on the supertype, just link to them

lib/maplibre-compose/src/commonMain/kotlin/org/maplibre/compose/layers/Anchor.kt
```diff
@@ -10,79 +10,155 @@ import org.maplibre.compose.util.MaplibreComposable
...
-   * base map style. See [Anchor.Companion.Above] to use this in the layers composition.
+   * Layers are placed directly over the highest base-style layer that [predicate] accepts, or at
+   * the bottom of the stack when it accepts none. See [Anchor.Companion.Above] to use this in the
+   * style content.
+   *
+   * Every time the style content is applied to the loaded style, the predicate is called with a
+   * [LayerHandle] for one base-style layer at a time, from the top of the stack down, until it
+   * accepts one. The handles are read-only: a property setter throws
+   * [StyleHandleException][org.maplibre.compose.style.StyleHandleException].
    */
```
sargunv, line 39: document only semantics, not implementation narration

Issue #821, opened 2026-06-18
`SymbolLayer` `textOffset` scales with OS text size changes (potential em unit issue)


michalgwo:
### What happened?

I noticed that when the system-level text size increases (e.g., via OS accessibility settings), the `textOffset` in `SymbolLayer` scales up proportionally as well.

This behavior likely occurs because `textOffset` is calculated using `em` units relative to the font size. While this might be intentional for maintaining relative spacing at different design scales, it creates a suboptimal user experience when the font size changes due to user accessibility preferences. The visual gap between the symbol and the text becomes aggressively large in larger text settings.

For example, this is how marker looks like with text size 110%:
<img width="232" height="303" alt="Image" src="https://github.com/user-attachments/assets/d5a76a99-6ace-4f78-89fc-4be75d8b3a28" />

The same marker with text size 200%:
<img width="268" height="316" alt="Image" src="https://github.com/user-attachments/assets/70ebea20-5b35-42c5-b278-e9a52edceb91" />
As you can see, a big gap appears between the marker and text, which doesn't look good.

I noticed this on Android, and I didn't check if it happens on other platforms.

### Affected Platforms

- [x] Android
- [ ] iOS
- [ ] Desktop (JVM)
- [ ] Browser (JS)
- [ ] Browser (Wasm)

### Platform Version

Android 15

### Library Version

0.13.0

### Sample Code

``` kotlin
SymbolLayer(
        id = layerId,
        source = source,
        iconImage = image(iconImage),
        iconAnchor = const(SymbolAnchor.Bottom),
        textAnchor = const(SymbolAnchor.Bottom),
        textOffset = offset(0.em, (-2.8).em),
        textSize = const(12.sp),
        textField = textField,
        textFont = const(listOf("Noto Sans Regular")),
    )
```

sargunv:
Respecting the accessibility settings (SP) is the intended default behavior for composition-driven layers, but you can (from memory) divide SP by something like `LocalDensity.current.fontScale` to set in DP

sargunv:
oh wait, you mean textOffset

hmm, will need to take another look to determine if this should be at dp or sp scale

sargunv:
the style spec unit [is in ems](https://maplibre.org/maplibre-style-spec/layers/#text-offset), so I think SP scale is correct.

Would be nice to at least provide some example though on how to configure offset to make marker labels look good at different accessibility scales, so will tag with api-ergonomics

westnordost:
[...]

PR #1350, opened 2026-09-08
Replace drag and drop demo with editable markers

sargunv:
## Description

Replace the Drag & drop demo with an editable collection of places. Pins support live label and color editing, selection and hover animations, map-level dragging, and deletion through a trash drop target or the Material list. Tap, drag, and outside-tap behavior keep editing separate from creating and moving pins.

The text-size slider demonstrates labels with a constant visual gap at accessibility sizes. Demo map content receives the active style's font stack, preserving the chosen basemap; the raster OSM style gains a glyph endpoint for labels.

Related to #821.

## Validation

https://github.com/user-attachments/assets/dddadf32-63c0-4f56-afc1-342e4cefda17

## AI assistance

Kimi K3 and GPT-6 Astra


westnordost:
Looking at the video, there is one tiny bug: 

When creating a new pin, the speech bubble is visible immediately without animation instead of animating in the same way it does when editing a pin.

PR #1365, opened 2026-09-09
Register TTF and OTF files for symbol-layer text-font

### LINKED ISSUE #1361: support loading a compose font into the map
start here: https://kotlinlang.org/docs/multiplatform/compose-multiplatform-resources-usage.html#fonts and https://maplibre.org/maplibre-style-spec/font-faces/

potentially something analogous to our image api

sargunv:
Resolve #1361

## Description

Adds `font(name, bytes, fallbacks)` and `rememberFont(name, FontResource, fallbacks)`, which register a TTF or OTF file under a `text-font` stack name the way `image()` registers a bitmap. Files are served to the engine from a content-addressed store through the existing resource routing. A registered name replaces a base-style `font-faces` entry with the same name.

On the browser the font is set at runtime through `map.setFontFaces()`. MapLibre Native has no runtime setter (maplibre/maplibre-native-ffi#714), so a registered font is merged into a `BaseStyle.Json` document at its next load, the snapshotter loads its document again with the fonts the composition declared, and a `BaseStyle.Uri` cannot use registered fonts yet.

## Validation

- `mise run check`, `mise run test:desktop`, `mise run test:js`, `mise run test:android`, iOS simulator test compile.
- New live test draws text from a generated font while the glyph server returns 404. It passes on desktop, browser, iOS, and the arm64 Android emulator; it is skipped on x86_64 Android, where the engine receives the file but draws nothing (maplibre/maplibre-native-ffi#713).

## AI assistance

Claude Code.


sargunv:
Dropping this for now: MapLibre Native has no runtime font-faces setter and adding one to the core is not trivial, so on native the font would only apply at the next base style load. Revisit once maplibre/maplibre-native-ffi#714 lands.

